PR #1, opened 2022-06-15
Add maven verification workflow

laeubi:


akurtakov:
Please don't merge things until both repos are merged and switch is done to prevent any delays.

laeubi:
> Please don't merge things until both repos are merged and switch is done to prevent any delays.

This should be independent of that or do you think there are any problems?

laeubi:
@mickaelistria the build complains:

> Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:3.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.equinox.compendium.sdk: Only qualifier changed for (org.eclipse.equinox.compendium.sdk.feature.jar/3.22.300.v20220615-0641). Expected to have bigger x.y.z than what is available in baseline (3.22.300.v20220503-1634)

As this does not touch any code: could it be related to the recent move of the repository?

mickaelistria:
> could it be related to the recent move of the repository?

Yes, most likely the merge commit has affected that module a way or another so the versions should be bumped. If you don't want to bump the version, you'll need to see what has changed in history compared to 3.22.300.v20220503-1634 and maybe exclude this file from the "watched" ones that are used to derive the qualifier. But just bumping is much easier...

laeubi:
Okay as @akurtakov has already mentioned its maybe good to wait until the merge of the runtime is done (i'll try to find some time if no one else step up here), so we catch all bumps at once, I just wanted to make sure that its not a useless bump.

vogella:
Currently there was no merge just a direct push to a new repo. I only added
a new remote in my local clone and pushed to the new repo url

akurtakov:
I think this can be finished/merged now.

laeubi:
I found it useful in the past two have both but of course this might change over time.

PR #28, opened 2022-06-16
[GH-build] Consider target-files in deps-caching and set build timeout

HannesWell:
The dependencies stored in the .m2 folder not only depend on the content of pom.xml files but also on the content of *.target-files. Therefore the latter should be considered when computing cache-hash.

An enhancement of the `setup-java` action to simplify the setup was already requested https://github.com/actions/setup-java/issues/345

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


HannesWell:
When thinking about this again, we don't gain anything with this change since we use the platform-aggregator target-file here that is not stored within the repo...

Issue #15, opened 2022-05-27
Deadlock between "SCR Component Actor" and "main" thread


HannesWell:
After updating one of my Eclipse installations (Eclipse Modlling-tools provided by Oomph plus some more plugins) on Linux, it seemed not to start anymore (no UI showed up, not even the icon becomes visible in the task-bar). Attempting to start it a second time did not work because the workspace is blocked. Nevertheless the System Monitor showed that the eclipse process is running and VisualVM detected the following deadlock:

```
2022-05-27 10:21:37
Full thread dump OpenJDK 64-Bit Server VM (11.0.15+10-post-Debian-1deb10u1 mixed mode, sharing):

Threads class SMR info:
_java_thread_list=0x00007fad1c1ddc20, length=25, elements={
0x00007fad78019800, 0x00007fad78216800, 0x00007fad78218800, 0x00007fad7821e000,
0x00007fad78220000, 0x00007fad78222800, 0x00007fad78224800, 0x00007fad78226800,
0x00007fad78244800, 0x00007fad787a0800, 0x00007fad787f7000, 0x00007fad78802800,
0x00007faca0031800, 0x00007faca022c000, 0x00007faca049e000, 0x00007faca050c800,
0x00007faca06ac800, 0x00007fad788b7000, 0x00007fad78f7f000, 0x00007fad28001000,
0x00007fac782ac000, 0x00007facc4020800, 0x00007faccc260000, 0x00007faccc067000,
0x00007facc4021800
}

"main" eclipse-equinox/equinox.framework#1 prio=6 os_prio=0 cpu=1071,22ms elapsed=62,52s tid=0x00007fad78019800 nid=0x97c waiting for monitor entry  [0x00007fad7d001000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at org.eclipse.osgi.internal.serviceregistry.ServiceRegistrationImpl.getService(ServiceRegistrationImpl.java:541)
        - waiting to lock <0x0000000092005488> (a org.eclipse.osgi.internal.serviceregistry.ServiceFactoryUse)
        at org.eclipse.osgi.internal.serviceregistry.ServiceRegistry.getService(ServiceRegistry.java:534)
        at org.eclipse.osgi.internal.framework.BundleContextImpl.getService(BundleContextImpl.java:660)
        at org.apache.felix.scr.impl.manager.SingleRefPair.getServiceObject(SingleRefPair.java:88)
[...]

PR #31, opened 2022-06-16
Add Test-case for lazy DS-Component activation deadlock (issue #15)

HannesWell:
_This PR was originally created as https://github.com/eclipse-equinox/equinox.framework/pull/61 and is now moved to this repository due to the merge of the equinox.framework and equinox.bundles repository._

---
Test case to reproduce: https://github.com/eclipse-equinox/equinox/issues/15

The case is currently designed to be very close to scenario that happens to me where m2e is in involved.
Before this is merged it will likely has to be generalized and simplified.

Instead of a `CountDownLatch` in `MavenModelManager.HeavyComponent` that is count down in `MavenProjectManager.HeavyComponent` only a `Thread.sleep()` could be used or, when being debugged one could place a break-point there.
Using a CountDownLatch just reliably ensures that the test passes falsely due to race-conditions and the test does not take unnecessarily long. But it might influence a solution of this problem badly.


HannesWell:
I think the new tests should be moved to `org.eclipse.equinox.ds.tests`, which seems to be a more suitable location.

HannesWell:
Integrated into https://github.com/eclipse-equinox/equinox/pull/68

PR #33, opened 2022-06-16
Simplify String-pooling in EclipsePreferences and instantly share preference constants

HannesWell:
The JVM probably can pool Strings better than using a HashMap.
But even if the JVM does the same it will have the same performance and we have saved some code.


merks:
That the JVM does it better is hopefully true today but was definitely not true in the past.

vogella:
LGTM, maybe can we also replace the ImmutableMap with a standard implementation at some point in time.

merks:
I'm not sure we should make assumptions that the JVM does it better:

https://www.baeldung.com/java-string-performance
https://itecnote.com/tecnote/java-performance-penalty-of-string-intern/

If that assumption is wrong, how and when will we notice?

I do recall that there have been major improvements in how it's implemented in the JVM but in the past it was really horrible...

merks:
@tjwatson you probably know better what we should expect from using String.intern and why that wasn't done in the past.

vogella:
Also adding @[jukzi](https://github.com/jukzi) as performance expert.

HannesWell:
> LGTM, maybe can we also replace the ImmutableMap with a standard implementation at some point in time.

Already started that locally. I just wanted to keep the changes smaller :)

> I'm not sure we should make assumptions that the JVM does it better:
> 
> https://www.baeldung.com/java-string-performance
> https://itecnote.com/tecnote/java-performance-penalty-of-string-intern/
> 
> If that assumption is wrong, how and when will we notice?
> 
> I do recall that there have been major improvements in how it's implemented in the JVM but in the past it was really horrible...

Thanks for the links. Following the second one I would agree that String intern should not be used and a ConcurrentHashMap does a way better job. But I'm missing the Java version this benchmark was done with.
On the other hand from skimming http://java-performance.info/string-intern-in-java-6-7-8/ I read the opposite that a manual String pool does not work well. But that benchmark is using a WeakHashMap.

Given that String.intern() is also slow for Java-11 and later I just wonder why the JVM fails to do it better than a simple ConcurrentHashMap? Aren't we always told that the JVM is so smart? Has the JVM more general requirements a ConcurrentHashMap does not serve? And if not, why does the JVM not simply use a ConcurrentHashMap internally?
But that are questions we would have to ask the JDK folks.

But we should first find out what the real performance is for us.

laeubi:
> Given that String.intern() is also slow for Java-11 and later I just wonder why the JVM fails to do it better than a simple ConcurrentHashMap?

String.intern() is always global and at least at linux a native method, so it *might* require some OS level synchronization, while a local benchmark might benefit from some speed improvements but require mores memory (you can often trade space for time)! And if I take a look at the benchmarks at all we are talking about *millions* of strings taking just a few milliseconds so what can we really gain here even if JVM method is "slower"?

tjwatson:
I have basically the same opinion and recollection as others.  I know in the past this String.intern would cause serious issues with performance.  I also know I've heard that has improved from release to release.  But I still worry about putting large amounts of String data into the intern pool.  In the framework we only selectively intern string for particular cases where we want to guarantee == operations will work for string equality for performance reasons.

laeubi:
> But I still worry about putting large amounts of String data into the intern pool.

But that's the point, how "large" could this grow here for eclipse preference keys? a few hundres maybe? So probably @HannesWell first want to add a counter of strings, or use a debugger to see how many a typical eclipse Workspace actually contains...

bjhargrave:
> Not using `String.intern` means you can dispose of the string cache if you need to. Using `String.intern` means it lasts forever. I think this is the primary distinction.

For prefs, I would choose not using `String.intern` since prefs is a bundle that can be installed, started and stopped. So the lifecycle of the string cache should be tied to the lifecycle of the bundle.

laeubi:
Interned String do not "live forever": http://java-performance.info/string-intern-in-java-6-7-8/

> all strings in the JVM string pool are eligible for garbage collection if there are no references to them from your program roots.

HannesWell:
> > But I still worry about putting large amounts of String data into the intern pool.
> 
> But that's the point, how "large" could this grow here for eclipse preference keys? a few hundres maybe? So probably @HannesWell first want to add a counter of strings, or use a debugger to see how many a typical eclipse Workspace actually contains...

Probably the amount of strings simply depends on the number of projects one has in the workspace and how many workspace and project specific preferences are specified. With the latter one you can probably increase the number arbitrarily. Since each key and each value of each *.prefs file is pooled.

But with the given resource I cannot tell if it is clearly better (more performant, less memory, ...) in the end to use `String.intern()` or a manually crafted pool. 

However I updated to PR now to pool all key/values immediately and not only after 5min and to use a static final ConcurrentHashMap as a pool.
I would say that preferences usually don' change that fast and pooling them immediately should have the advantage that the duplicate is faster garbage-collected since it cannot make it in older GC generations. Having a global cache avoids rebuilding the map on each share operation and also ensures that each string only exists once in the preferences.
I assume that most preference constants don't vanish over time and therefore this does not create a memory leak.
Additionally this pool gets eligible for garbage collection if the bundle's classloader is GCed.

HannesWell:
Do you have any more remarks here? Do you think the current change is good, should be changed or abandoned?

HannesWell:
Alternatively we could use [`-XX:+UseStringDeduplication`](https://dzone.com/articles/usestringdeduplication) in Eclipse and let the Garbage-collector do the duplication. In recent Java versions more GCs support this feature, not only `G1`.
But I have not yet full thought through if the exact behavior of that feature fits good for this use case or not.

vogella:
> Do you have any more remarks here? Do you think the current change is good, should be changed or abandoned?

I suggest to merge this, IMHO is it nice to use standard functionality instead of a custom pool. I assume that the JVM stuff will receive better attention compare to a custom pool hidden in some Equinox prorject.

HannesWell:
> > Do you have any more remarks here? Do you think the current change is good, should be changed or abandoned?
> 
> I suggest to merge this, IMHO is it nice to use standard functionality instead of a custom pool. I assume that the JVM stuff will receive better attention compare to a custom pool hidden in some Equinox prorject.

In the meantime I have reworked this a bit to still use a own bundle pool, but more directly. See https://github.com/eclipse-equinox/equinox/pull/33#issuecomment-1167845947. But it would be trivial to use String.intern instead.

> Alternatively we could use [`-XX:+UseStringDeduplication`](https://dzone.com/articles/usestringdeduplication) in Eclipse and let the Garbage-collector do the duplication. In recent Java versions more GCs support this feature, not only `G1`. But I have not yet full thought through if the exact behavior of that feature fits good for this use case or not.

That would also be an option, that was already suggested elsewhere too.

vogella:
@HannesWell thanks for the update. Looks to me that you are still creating our own String pool instead of "simply" relying on the String.intern.

HannesWell:
> @HannesWell thanks for the update. Looks to me that you are still creating our own String pool instead of "simply" relying on the String.intern.

Exactly, but I just did it because in the previous discussion in this PR there was no agreement that String.intern() should be used.
But it would be a trivial task to replace the `ConcurrentHashMap` based `STRING_POOL` by `String.intern()` (just replace `EclipsePreferences.intern()` by `String.intern()`).
@jukzi since it seems you have the analysis already set up, would you be interested in trying-out/comparing the previous and current state of this PR and how it runs if String.intern() is used instead of a custom String-pool?

vogella:
@HannesWell Lets merge and test the String.intern() change as a next step? Any idea why these TCK are failing?

iloveeclipse:
> Use String.intern() instead of own StringPool

Can we please update PR title & description (commit message too?) with what we **actually** intend to do here? As it is NOT about String.intern() anymore, if I see it right?

HannesWell:
> > Use String.intern() instead of own StringPool
> 
> Can we please update PR title & description (commit message too?) with what we **actually** intend to do here? As it is NOT about String.intern() anymore, if I see it right?

Done that, thanks for the heads-up.


> > Not using `String.intern` means you can dispose of the string cache if you need to. Using `String.intern` means it lasts forever. I think this is the primary distinction.
> 
> For prefs, I would choose not using `String.intern` since prefs is a bundle that can be installed, started and stopped. So the lifecycle of the string cache should be tied to the lifecycle of the bundle.


In order to reflect that I changed the static final ConcurrentHashMap by a synchronized WeakHashMap so that the shared string can be GCed if they are not referenced anymore.

In general the performance of String.intern(), at least used to be not very good:

> I'm not sure we should make assumptions that the JVM does it better:
> 
> https://www.baeldung.com/java-string-performance https://itecnote.com/tecnote/java-performance-penalty-of-string-intern/
> 
> If that assumption is wrong, how and when will we notice?
> 
> I do recall that there have been major improvements in how it's implemented in the JVM but in the past it was really horrible...

I cannot say if that changed, but at SO the over all opinion seems to be similar (however the accepted answer seem to have the same author like the linked blog-post):
https://stackoverflow.com/questions/10624232/performance-penalty-of-string-intern


As an alternative mentioned above: using the JVM's String deduplication might also be a solution that the JVM can take over. But it might also have its drawbacks.

jukzi:
I tested this change
before:
image
after:
image
https://github.com/eclipse-equinox/equinox/pull/420 :
image

This change also solves https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1692, but has some few more duplicates then 420 in my test scenario (start egit workspace, wait 5 sec, take heapdump)

jukzi:
btw the title has a typo, i can not edit it
> prefernece

bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/EclipsePreferences.java
```diff
@@ -629,15 +628,26 @@ protected String internalPut(String key, String newValue) {
...
+			} else if (DEBUG_PREFERENCE_SET) {
 				PrefsMessages.message("Setting preference: " + absolutePath() + '/' + key + '=' + newValue); //$NON-NLS-1$
-			properties = properties.put(key, newValue);
+			}
+			properties = properties.put(intern(key), intern(newValue));
 			return oldValue;
 		}
 	}
 
+	private static final Map<String, String> STRING_POOL = new WeakHashMap<>();
```
tjwatson, line 641: I believe this will pin the String for all time because `WeakHashMap` only keeps weak references to the keys.  The values continue to have strong references if I recall correctly.  So using the cached string as both the key and the value will causing that string to be pinned and never removed.  I'm unsure why we don't just use `String.intern`.  I'm fairly certain modern VMs have solved the old performance concerns with using intern.

iloveeclipse: While agreeing with the first part, I'm not confident that "modern VMs have solved the old performance concerns with using intern." I only trust facts & numbers :-)

Regarding the first part, see also yet another place where we do String caching: `org.eclipse.jdt.internal.core.util.DeduplicationUtil`

jukzi:
i think here is the history of openjdks string.intern()
https://github.com/openjdk/jdk/commits/7fbfb3b74a283261027e6c293e1a5dbc354cf0af/src/hotspot/share/classfile/stringTable.cpp
Seems active but mostly securityfixes and GC issues.

Here are numbers from jmh benchmark i made for DeduplicationUtil
openjdk **15.0.2** 2021-01-19
```
Benchmark                                 Mode  Cnt    Score   Error  Units
Deduplication.ConcurrentHashMap             ss   50   48,947 ± 4,154  ms/op
Deduplication.DeduplicationUtil             ss   50   56,004 ± 4,835  ms/op
Deduplication.DeduplicationUtilinternMap    ss   50   52,069 ± 4,298  ms/op
Deduplication.HashMap                       ss   50   43,158 ± 3,854  ms/op
Deduplication.LRUCache                      ss   50   60,375 ± 4,173  ms/op
Deduplication.WeakHashMap                   ss   50   42,413 ± 4,359  ms/op
Deduplication.WeakHashSet                   ss   50   45,319 ± 4,171  ms/op
Deduplication.charDeduplicationUtil         ss   50  102,697 ± 2,723  ms/op
Deduplication.guava                         ss   50   50,724 ± 6,487  ms/op
Deduplication.intern                        ss   50  248,533 ± 2,885  ms/op
```

**openjdk-22** -ea+26_windows-x64_bin:
```
Benchmark                                 Mode  Cnt    Score   Error  Units
Deduplication.ConcurrentHashMap             ss   50   45,703 ± 4,785  ms/op
Deduplication.DeduplicationUtil             ss   50   55,352 ± 3,900  ms/op
Deduplication.DeduplicationUtilinternMap    ss   50   42,606 ± 4,212  ms/op
Deduplication.HashMap                       ss   50   34,847 ± 4,474  ms/op
Deduplication.LRUCache                      ss   50   58,511 ± 4,849  ms/op
Deduplication.WeakHashMap                   ss   50   33,776 ± 4,654  ms/op
Deduplication.WeakHashSet                   ss   50   42,029 ± 4,152  ms/op
Deduplication.charDeduplicationUtil         ss   50   71,593 ± 3,281  ms/op
Deduplication.guava                         ss   50   47,871 ± 5,825  ms/op
Deduplication.intern                        ss   50  254,158 ± 5,848  ms/op
```

=> String.intern() still slow

Nervertheless i still think that EclipsePreferences are still not a hotspot anywhere and the focus should be more on memory then to add few preferences. JDTs DeduplicationUtil which is used during compile is called way more often and therefor more performance sensitive.

tjwatson:
> => String.intern() still slow

I'm getting old.  As I recall the `String.intern()` issue.  It wasn't about it being slow to `intern`, it was about it causing memory issues because it would pin large amounts of memory that would never get reclaimed.  One example I recall was if you did a String.substring we would have a new String, but that string would just be a pointer into an existing char[] of the original (larger) String.  If you then interned a string like that it would pin the complete char[] in memory even if you released the original String for GC.  Parsing large amounts of data from a single String that split it up into multiple strings would cause this common issue.

I would not think the code that actually does the `intern` calls would be performance sensitive and that we would not care so much about how much time it took to `intern` a string.

laeubi:
> I would not think the code that actually does the intern calls would be performance sensitive and that we would not care so much about how much time it took to intern a string.

I agree with @tjwatson that performance for preferences are not really an issue or hotspot and doubt the now thread-safe and weak variant is actually faster and the general rule applies here if you do it in a safe and consistent way it is probably not faster than the built in. And using the `intern()` would allow the JVM to even share more things and manage it better than we can.

The speedup in benchmarks usually seen is because they often work on local data (where `intern()` needs to work on _shared_ global data) so no synchronization is needed at all, and often such benchmarks simply create a lot of unique strings, what is usually not the case for when we want to intern items.

vogella:
To be fair we must apply @HannesWell change, if he agrees and updates it. It was his initial proposal

mickaelistria:
> To be fair we must apply @HannesWell change, if he agrees and updates it. It was his initial proposal

I think the EDP just mandates that we take the patch that is best for the project. Whichever gets in doesn't mean that don't acknowledge and recognize the work done by all committers on previous patches and discussions to reach the solution.

vogella:
> I think the EDP just mandates that we take the patch that is best for the project. 

Yes, but @HannesWell initial patch used String.intern() and some additional clean-up and he was ask to change it. So I think it is the better patch and I think "being nice"  is also something we should try to be.

merks:
My "sense" is that in this context, performance is likely not a major concern, and accumulating lots of garbage strings is also not likely a major concern.   (I was never suggesting that EMF's string pool should be used or copied; that pool exists because URIs are and endless set of strings and the underlying parts need sharing to solve that problem.)

HannesWell:
Two Different things have to be considered here:
- Possible Memory problems due to too many Strings being interned and interned Strings not being GCed
Since JDK 7 at least Hotspot keeps interned Strings at the heap (before it was at perm-gen)
  - https://www.oracle.com/java/technologies/javase/jdk7-relnotes.html#jdk7changes
  - https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6962931
When using just `String.intern()` in my little snippet above I can confirm that interned Strings are also GCed (at least String.intern() can later return another reference for an equal object if the prvious one was GCed).
So interning Strings should not cause memory shortage in the Perm-Gen anymore and it looks like interning also does not prevent the GC from reclaiming the String.
Therefore from a memory usage point of view I see no drawbacks in using `String.intern()` and the advantage to have a 'common' pool that is also used by other components (in Eclipse or third-party plugins/libraries).

- Performance of interning a string and looking up an interned a string
Probably not a hotspot, but when using `String.intern()` in my snippet above and increase the number of iterations to 10^7 the runtime is about four to five times faster compared to the custom String pool.
Of course this is just a poor-mans benchmark, but the numbers are relatively significant. But if I look at the size of the custom pool's WeakHashMap in the end of the benchmark, it is very large although its entries could be garbage collected. If I then add a `System.gc()` between the first and second loop the runtimes of the program using the custom String pool goes up in similar areas.
Therefore I have the impression, that interned String are GCed more aggressively.
This also corresponds to https://shipilev.net/jvm/anatomy-quarks/10-string-intern/, which also warns that String.intern() can increase the GC work significantly (of course depending on how many Strings are eventually interned). This also contraticts the assumption (which I usually share, that in most cases the JVM and the built-in default is faster/smarter/better than a custom solution). But it uses Java 1.8 as reference, so maybe this has been improved too (I haven't investigated Jörgs link in depth).
Btw. adding the VM argument `-XX:+PrintStringTableStatistics` with 10^7 Strings interned (and some being GCed again), this is the print-out for me:
```
SymbolTable statistics:
Number of buckets       :     32768 =    262144 bytes, each 8
Number of entries       :      3296 =     52736 bytes, each 16
Number of literals      :      3296 =    116104 bytes, avg  35.000
Total footprint         :           =    430984 bytes
Average bucket size     :     0.101
Variance of bucket size :     0.100
Std. dev. of bucket size:     0.316
Maximum bucket size     :         4
StringTable statistics:
Number of buckets       :   1048576 =   8388608 bytes, each 8
Number of entries       :   1382094 =  22113504 bytes, each 16
Number of literals      :   1382094 =  88460136 bytes, avg  64.000
Total footprint         :           = 118962248 bytes
Average bucket size     :     1.318
Variance of bucket size :     2.572
Std. dev. of bucket size:     1.604
Maximum bucket size     :        12
```
I also ran the JMH Benchmark from that article again using temurin-jdk-21.0.1+12 (in the article Java-1.8 is used) and got the following numbers (chm - `ConcurrentHashMap`, hm - `HashMap`):
```
Throughput mode:
Benchmark             (size)   Mode  Cnt         Score        Error  Units
StringIntern.chm           1  thrpt    5  68490243,982 ± 707782,826  ops/s
StringIntern.chm         100  thrpt    5    678656,634 ±   4520,300  ops/s
StringIntern.chm       10000  thrpt    5      3830,558 ±    202,857  ops/s
StringIntern.chm     1000000  thrpt    5        13,061 ±      0,122  ops/s
StringIntern.hm            1  thrpt    5  74991634,225 ± 130467,591  ops/s
StringIntern.hm          100  thrpt    5    719297,684 ±  19457,429  ops/s
StringIntern.hm        10000  thrpt    5      4280,492 ±     67,033  ops/s
StringIntern.hm      1000000  thrpt    5        15,163 ±      0,171  ops/s
StringIntern.intern        1  thrpt    5  20199803,090 ± 828703,038  ops/s
StringIntern.intern      100  thrpt    5    190436,286 ±   4641,474  ops/s
StringIntern.intern    10000  thrpt    5      1688,377 ±    109,560  ops/s
StringIntern.intern  1000000  thrpt    5         6,106 ±      0,136  ops/s
```

Looks like the results got better with more recent Java versions, but at least in the benchmarked case there is still a difference at the factor three to four.

Of course it is always the question how much these test scenarios match the real use case with regards to single/multi-threading and the amount of Strings to intern (when I started an eclipse with three simple projects in the workspace the pool contained about ~4000 elements). And of course it is to question if the runtime of the interning itself is significant part of the overall runtime or if it is neglectable.
And I agree on this with the others that the interning performance probably it isn't a concern. Nevertheless the increased GC work is something I find not ideal, but again the question is how bad the impact really is and if modern collectors can handle it better. But for both I don't have significant numbers.

> So it looks like @jukzi @laeubi and @tjwatson all agree #420 is a better PR for this case? What do you think @HannesWell ? If you agree that #420 is good enough in term of performance, please close this one in favor of the other simpler proposal.

In fact this PR started with using `String.intern()`, but when concerned were raised https://github.com/eclipse-equinox/equinox/pull/33#issuecomment-1158519665 and https://github.com/eclipse-equinox/equinox/pull/33#issuecomment-1161531519. Therefore I implemented the alternative. If we eventually agree that `String.intern()` is the way to go, yes I would like to return to this initial state and just use `String.intern()` in this PR, especially since this PR cleans up much more, although this renders #420 obsolete.

@iloveeclipse and @merks with the updated information are you fine with using `String.intern()`?
My suggestion is to use `String.intern()` for now and if it later turns out that in fact `String.intern()` is too much of a performance hit, we can still use the custom pool presented here (I would say that it is not too complicated and for that should have a good performance and memory characteristics).

merks (approved):
Looks fine.

vogella:
Thanks a bunch @HannesWell for updating this again and again and performing detailed testing. Also thanks to @jukzi for providing test data and all others which raised concerns and gave input.

The failing TCK test seem unrelated (we see them in all PR for Equinox).

I merge this now, so that we can see this in action. This should leave us sufficient time to revert or improve in case we identify additional issues. From @jukzi testing we know that this should remove at least~ 20 000 ("do not insert") and "insert" strings from the heap of his workspace.

Issue #27, opened 2022-06-16
Produce correct module-info class for org.eclipse.osgi using BND


akurtakov:
Log is:
```
Comparator differences from current build
	/home/jenkins/agent/workspace/I-build-4.25/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/siteDir/eclipse/downloads/drops4/I20220616-0910
compared to reference repo at 
	https://download.eclipse.org/eclipse/updates/4.25-I-builds

1.  equinox/bundles/org.eclipse.osgi
   no-classifier: different
      module-info.class: present in baseline only
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]
```
at https://download.eclipse.org/eclipse/downloads/drops4/I20220616-0910/buildlogs/comparatorlogs/buildtimeComparatorUnanticipated.log.txt

akurtakov:
@tjwatson is module-info.class in o.e.osgi needed ? what could have caused it to no longer be generated/added

sravanlakkimsetti:
module-info definition is provided in pom.xml https://github.com/eclipse-equinox/equinox/blob/48152cb7d8bbd2a2cedb253e57d53ccf6d5e9e77/bundles/org.eclipse.osgi/pom.xml#L51. This is not getting generated after equinox merge.

sravanlakkimsetti:
There are extra maven tasks after merge. These extra maven tasks are causing the trouble.
Here ate the task lists
Old  : [log_old.txt](https://github.com/eclipse-equinox/equinox/files/8924894/log_old.txt)
New : [log_new.txt](https://github.com/eclipse-equinox/equinox/files/8924895/log_new.txt)

sravanlakkimsetti:
specifically there are two tasks that are running extra now 
```
[INFO] --- tycho-source-plugin:3.0.0-SNAPSHOT:feature-source (feature-source) @ org.eclipse.osgi ---
[INFO] 
[INFO] --- tycho-p2-plugin:3.0.0-SNAPSHOT:p2-metadata (attach-p2-metadata) @ org.eclipse.osgi ---
[INFO] 
```
they are running just before add-module-info task.

laeubi:
> they are running just before add-module-info task.

these should be independet from any other task generating sources, the more interesting would be if the moditec is running and if it issues any warning/errors.

laeubi:
If I build it local I see a `module-info.class` included...

[...]

PR #34, opened 2022-06-17
Update moditect-maven-plugin to 1.0.0.RC2

akurtakov:
Provoked by https://github.com/eclipse-equinox/equinox/issues/27 so
build uses the latest plugin version.


akurtakov:
Change in last commit is only to fix a typo in commit message thus pushing.

PR #36, opened 2022-06-17
Use BND to generate the JPMS info automatically

laeubi:
This replaces the manual crafted moditec JPMS info with the automatically generated one by BND.

As the new tycho-extension for bnd is not yet published, we simply use an explicit maven execution in the pom right now.

FYI @stbischof

PR #41, opened 2022-06-18
Improve the output of the automatic JPMS generation

laeubi:
This is for further improving the work started at https://github.com/eclipse-equinox/equinox/pull/36.


laeubi:
@rotty3000 we have an issue with the generated class see https://github.com/eclipse-equinox/equinox/issues/42 is it possible to adjust this?

HannesWell:
Btw. couldn't we use the `bnd-process` goal with an bnd element in the `bnd-maven-plugin` configuration that contains the content of the `bnd.bnd` file, just like we do it in m2e?
https://github.com/eclipse-m2e/m2e-core/blob/4c1f0dc8667f4e0a4e3f6871b9b0e3d91a6e7ac2/m2e-maven-runtime/pom.xml#L61

This would save us the extra bnd.bnd file.

laeubi:
> This would save us the extra bnd.bnd file.

An require us to embedd non XML data into XML ;-)

beside that, my goal is more to get rid of the pom here after the next tycho release...

laeubi:
@tjwatson I'm not sure about the state of this, do you plan to take a look or should we close this?

akurtakov:
Looks like it's good enough so closing. @tjwatson can reopen/push it if still needed.

PR #43, opened 2022-06-18
Simplify equinox.cm.test

HannesWell:


bundles/org.eclipse.equinox.cm.test/src/org/eclipse/equinox/cm/test/Activator.java
```diff
@@ -13,52 +13,31 @@
...
-		packageAdmin = service;
+	static BundleContext getBundleContext() {
+		return FrameworkUtil.getBundle(Activator.class).getBundleContext();
 	}
 
-	static synchronized BundleContext getBundleContext() {
-		return bundleContext;
-	}
+	public static Bundle getBundle(String symbolicName) {
+		FrameworkWiring fkw = getBundleContext().getBundle(Constants.SYSTEM_BUNDLE_LOCATION)
```
HannesWell, line 31: This is basically what `Platform.getBundle(String)` is doing. Do you think it would be ok for this test to require o.e.core.runtime?

akurtakov:
Be careful with Suite classes removal , it would prevent running these tests on I-builds.

HannesWell:
> Be careful with Suite classes removal , it would prevent running these tests on I-builds.

This project does not have a test.xml so I suspect it is not executed as part of the I-build and removing the suit isn't an issue?

HannesWell:
> > Be careful with Suite classes removal , it would prevent running these tests on I-builds.
> 
> This project does not have a test.xml so I suspect it is not executed as part of the I-build and removing the suit isn't an issue?

@akurtakov because there is no `test.xml` in this project is the assumption correct that it is not executed during the I-builds and the suite can therefore be removed without problems?

akurtakov:
> @akurtakov because there is no `test.xml` in this project is the assumption correct that it is not executed during the I-builds and the suite can therefore be removed without problems?

It is not run in nightly builds but please don't remove it but rather give me some time to make it run.

HannesWell:
> > @akurtakov because there is no `test.xml` in this project is the assumption correct that it is not executed during the I-builds and the suite can therefore be removed without problems?
> 
> It is not run in nightly builds but please don't remove it but rather give me some time to make it run. The rest of the patch is fine so please merge it without the suite removal.

Done. Reverted the removal of `AllTests` and left it untouched. The test-classes are simplified/modernized further but I think that is fine as well.

PR #44, opened 2022-06-18
Simplify org.eclipse.equinox.common.tests

HannesWell:
Simply run all tests in the plug-in instead of specifying suits that
contain all tests. This also simplifies the addition of new tests.


bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/AllTests.java
```diff
@@ -1,36 +0,0 @@
-/*******************************************************************************
```
akurtakov, line 1: If you remove this class tests can no longer be run in I-builds as it's waht is set in test.xml https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.equinox.common.tests/test.xml#L48

akurtakov (changes requested):
Tests should be runnable without tycho/maven and this is what I-builds does to run on all supported OS/arch combos. 
Either test framework should be extended to work without having suite or this class should stay.

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/AllTests.java
HannesWell: I'm a bit surprised. I assumed that the test.xml are just leftovers from the past where Maven was not used and wanted to suggest to remove them.
How are these test.xml files picked up in an I-build? AFAIK Tycho-surefire does not respect them and either runs the entire plug-in or the suite specified in corresponding properties.

akurtakov: https://wiki.eclipse.org/Platform-releng/Eclipse_Test_Framework (although probably outdated in places) should explain things. https://github.com/eclipse-platform/eclipse.platform.releng/blob/master/bundles/org.eclipse.test/src/org/eclipse/test/EclipseTestRunner.java is the key part IIRC.

HannesWell:
> Tests should be runnable without tycho/maven and this is what I-builds does to run on all supported OS/arch combos. Either test framework should be extended to work without having suite or this class should stay.

I have to admit I never noticed that most (but not all?) test plug-ins have a test.xml, I noticed them occasionally but assumed they are just ancient leftovers. Can you explain why they are necessary and why Tycho cannot simply run them just with the default tycho:surefire that just executes all tests?

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/AllTests.java
akurtakov: Maven is not used for running the tests for I-builds as they are run on already provisioned target and this is done on 6 different os/arch/jvm combos without duplicating the whole rebuild process on each combo as maven would do.

mickaelistria:
"Official" tests, the ones that sends the emails, are run with antrunner against the whole SDK. Related jobs are available at https://ci.eclipse.org/releng/view/Automated%20tests/ . Tycho may be able to do the job, but that would require some more tweaks which so far never managed to become a priority

HannesWell:
Thank you both for the explanation. I wasn't aware of that.
But looking at it naively I suspect that it should be possible to run the entire plug-in by using `selectDirectory()` and passing the test plug-ins root path (is it a directory or a jar?) if no `testClassName` is specified here:
https://github.com/eclipse-platform/eclipse.platform.releng/blob/7768dbe43bbb01fbbb095ed0b79911522af819c7/bundles/org.eclipse.test/src/org/eclipse/test/EclipseTestRunner.java#L214

I will have a more detailed look at it tomorrow.

akurtakov:
It's quite possible for sure. There are just so many other things that this one stay as it. Last time @LucasBullen worked on making it support JUnit 5 which unfortunately I never managed to finish up.

akurtakov:
@HannesWell I'm closing this one as  test framework can not work without suite to drive it.

HannesWell:
> @HannesWell I'm closing this one as test framework can not work without suite to drive it.

Yes.
I started to work on an enhancement of the `EclipseTestRunner`, but didn't have the time yet to try it out or test if it really works. But I have the intention to come back to that topic later. If that works one day, I will reopen this.

PR #45, opened 2022-06-19
Apply new JDT clean-up "Convert while loops to enhanced for loops"

HannesWell:
Apply the new JDT clean-up [Convert while loops to enhanced for loops](https://www.eclipse.org/eclipse/news/4.25/jdt.php#while-to-enhanced-for).

The result is erroneous at the moment. Nevertheless I create this PR so that the JDT devs can have a look at it.

### DIFF
--- a/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/internal/app/AppCommands.java
+++ b/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/internal/app/AppCommands.java
@@ -113,9 +113,7 @@ public class AppCommands implements CommandProvider {
 		}
 
 		addHeader(Messages.console_help_app_commands_header, sb);
-		Iterator<Entry<String, String[]>> i = commandsHelp.entrySet().iterator();
-		while (i.hasNext()) {
-			Entry<String, String[]> entry = i.next();
+		for (Entry<String, String[]> entry : commandsHelp.entrySet()) {
 			String command = entry.getKey();
 			String[] attributes = entry.getValue();
 			addCommand(command, attributes, sb);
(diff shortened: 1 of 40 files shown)


HannesWell:
A general question to me is if we can gently clean up and fix warnings in the embedded `org.osg.service.application` sources?
Actually embedded (OSGi) sources should only be changed in the original, but given that the `org.osg.service.application` specification is dead there is no way to evolve it elsewhere.

@tjwatson are you ok with that?

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
```diff
@@ -1598,9 +1598,7 @@ private ResolutionError permuteUsedBlames(ResolveSession session,
                 perm2 = allCandidates.copy();
                 permRef2.set(perm2);
             }
-            for (int reqIdx = 0; reqIdx < usedBlame.m_reqs.size(); reqIdx++)
-            {
-                Requirement req = usedBlame.m_reqs.get(reqIdx);
+            for (Requirement req : usedBlame.m_reqs) {
```
tjwatson, line 1601: I prefer we first make this change back at the felix project.

tjwatson (commented):
To answer the original question.  I still prefer to keep the org.osgi API classes in line with the last version that was published.  That opinion obviously could change if we find something that is a bug or a substantial improvement.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
HannesWell: Yes. Thank you for mentioning that. It was not my intention to apply it there, I just oversaw that class when reverting the clean-up for OSGi/Felix sources.

Reverted that.

HannesWell:
> To answer the original question. I still prefer to keep the org.osgi API classes in line with the last version that was published. That opinion obviously could change if we find something that is a bug or a substantial improvement.

Acknowledge. I did not encounter a bug, just a bunch of warnings we then have to continue to suppress.

akurtakov:
This caused the following compile warning in I-build https://download.eclipse.org/eclipse/downloads/drops4/I20220721-0550/compilelogs/plugins/org.eclipse.equinox.registry_3.11.200.v20220720-2012/@dot.html#OTHER_WARNINGS .
```
1. WARNING in /src/org/eclipse/core/internal/registry/TableReader.java
 (at line 274)
for (int child : children) {
The value of the local variable child is not used
```

iloveeclipse:
It might also caused https://github.com/eclipse-platform/eclipse.platform.ua/issues/42?

PR #46, opened 2022-06-19
Fix SystemBundleTests testBug405919 to not compare inner resources with themself

HannesWell:
From the variable names I'm confident that the assignment is wrong in the tests.
@tjwatson is that correct?


tjwatson:
Good catch.  Thanks for fixing.

PR #47, opened 2022-06-19
Replace deprecated PackageAdmin service with Wiring-API

HannesWell:
Work in progress of the replacement of the deprecated PackageAdmin service within equinox.

The main work that is missing is the replacement of the `ExportPackages` inside `EquinoxCommandProvider`.
This requires some more code and I suspect that in the end a significant part of the PackageAdminImpl implementation as to be copied.

However I would like to discuss if the general direction is fine.


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/framework/util/Wirings.java
```diff
@@ -0,0 +1,120 @@
...
+import org.osgi.resource.Requirement;
+
+public class Wirings { // Wiring Util
+
+	private Wirings() {
+	}
+
+	// TODO: check usage
+
+	public static Stream<Bundle> getBundles(String symbolicName) {
```
tjwatson, line 47: We cannot have static API like this to get the Bundle objects because multiple instances of the framework can be loaded from the same class loader.

laeubi, line 47:
```suggestion
	public static Stream<Bundle> getBundles(BundleContext context, String symbolicName) {
```
I would suggest to simply force users passing a BundleContext, given that one has to fetch the `PackageAdmin` before anyways there must be a bundlecontext somewhere around, and the BC can be queried for the systembundle like BundleContext.getBundle(Constants.SYSTEM_BUNDLE) ...

laeubi:
I really like the idea of having some "WiringHelper" standardized, bnd also contains such helpers already and some common task are a bit tendentious to write "by hand".

@HannesWell as this has many conflicts now, maybe it would be worth to just replace one single (hopefully trivial) case where it can be discussed where to place such things and then do it incrementally?

tjwatson:
> I really like the idea of having some "WiringHelper" standardized, bnd also contains such helpers already and some common task are a bit tendentious to write "by hand".

I think we should do that in https://github.com/osgi/osgi/issues/425 like I suggested in my review comment.

HannesWell:
Sorry, I neglected this PR but agree that it is still relevant.
I'll look into the comments and will see if I can come up with a smaller set to start.
In general some utilities would probably be also helpful for PDE in order to migrate off the equinox-resolver API. But what would be helpful exactly will only become visible when going down that path.

PR #52, opened 2022-06-20
URIUtil: Fix UNC-related issues and provide UNC-safe URI-factories

### LINKED ISSUE #6: Trouble with UNC-Uris: Should I add URI-factories to URIUtil?

cmorty:
I had some issues with dollar-signs in UNC-Paths (e,.g. `\\WSL$`). There are others, too. [CDT](https://www.eclipse.org/cdt/) already has some special handling of UNC-Paths when creating URIs, but not everywhere. Thus the intention of this PR is to provide UNC-safe factories that can be used as drop-in replacement for `new java.net.URI()`. 

fixes #6

### DIFF
--- a/bundles/org.eclipse.equinox.common.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.common.tests/META-INF/MANIFEST.MF
@@ -8,7 +8,7 @@ Automatic-Module-Name: org.eclipse.equinox.common.tests
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Bundle-ActivationPolicy: lazy
 Require-Bundle: org.junit,
- org.eclipse.equinox.common;bundle-version="3.15.100",
+ org.eclipse.equinox.common;bundle-version="3.17.0",
  org.eclipse.core.tests.harness;bundle-version="3.11.400",
  org.eclipse.equinox.registry;bundle-version="3.8.200"
 Import-Package: org.eclipse.osgi.service.localization,
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/URIUtilTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/URIUtilTest.java
@@ -13,9 +13,18 @@
  *******************************************************************************/
 package org.eclipse.equinox.common.tests;
 
-import java.io.*;
-import java.net.*;
-import org.eclipse.core.runtime.*;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import org.eclipse.core.runtime.FileLocator;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.URIUtil;
 import org.eclipse.core.tests.harness.CoreTest;
 import org.osgi.framework.FrameworkUtil;
 
@@ -550,6 +559,37 @@ public class URIUtilTest extends CoreTest {
 			URI expected = data[i][2];
 			URI actual = URIUtil.makeRelative(location, root);
 			assertEquals("2." + Integer.toString(i), expected, actual);
+		}
+	}
+
+	/**
+	 * Test UNC-Paths containing a $.
+	 */
+	public void testDollar() throws URISyntaxException {
+		var relative = "SomePath";
+		URI expectedResolved = new URI("file:////WSL$/Ubuntu/SomePath");
+		String[] uris = {
+				// @formatter:off
+				"file:////WSL$/Ubuntu",
+				"file:////WSL$/Ubuntu/",
+				"file://////WSL$/Ubuntu",
+				"file://////WSL$/Ubuntu/"
+				// @formatter:on
+		};
+
+		for (int i = 0; i < uris.length; i++) {
+			URI base = new URI(uris[i]);
+			URI resolved = URIUtil.append(base, relative);
+			assertEquals("1." + Integer.toString(i), expectedResolved, resolved);
+		}
+
+		var relative2 = "SomePath/";
+		URI expectedResolved2 = new URI("file:////WSL$/Ubuntu/SomePath/");
+
+		for (int i = 0; i < uris.length; i++) {
+			URI base = new URI(uris[i]);
+			URI resolved = URIUtil.append(base, relative2);
+			assertEquals("2." + Integer.toString(i), expectedResolved2, resolved);
 		}
 	}
 }
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/URIUtilTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/URIUtilTest.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.equinox.common.tests;
 
+import static org.junit.Assert.assertThrows;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -566,9 +568,10 @@ public class URIUtilTest extends CoreTest {
 	 * Test UNC-Paths containing a $.
 	 */
 	public void testDollar() throws URISyntaxException {
-		var relative = "SomePath";
-		URI expectedResolved = new URI("file:////WSL$/Ubuntu/SomePath");
-		String[] uris = {
+		final var relative = "SomePath";
+		final String[] uris = {
+				// there must be four slashes because otherwise "WSL$" would be the host name
+				// and it is not allows to use $ in the host name
 				// @formatter:off
 				"file:////WSL$/Ubuntu",
 				"file:////WSL$/Ubuntu/",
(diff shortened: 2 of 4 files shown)


cmorty:
Can someone explain to me what I'm doing wrong? The Console just says that the test failed (https://ci.eclipse.org/equinox/job/equinox/job/PR-52/3/console) and the log something about the Workspace, which I don't really understand (https://ci.eclipse.org/equinox/job/equinox/job/PR-52/3/artifact/apiAnalyzer-workspace/data/.metadata/.log)

I assume that I did something wrong when adding the public methods, but I don't see an error that points to that.  :(

iloveeclipse:
> Can someone explain to me what I'm doing wrong?

```
1 API ERRORS
* Marker [on: /org.eclipse.equinox.common/META-INF/MANIFEST.MF, id: 206, type: org.eclipse.pde.api.tools.version_numbering, attributes: [apiMarkerID: 4, charEnd: 172, charStart: 154, description: - The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String, String) has been added
- The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String, int, String, String, String) has been added
- The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String) has been added
- The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String, String, String) has been added
, lineNumber: 5, message: The minor version should be incremented in version 3.16.200, since new APIs have been added since version  
```

Explanation: you've added 3 new public API methods, and that means, if the minor version segment wasn't changed in current release (it wasn't), it should be increased. 
So you have to change `3.16.100` to `3.17.0` in MANIFEST.MF and update `since` javadoc tags.

cmorty:
> ```
> 1 API ERRORS
> * Marker [on: /org.eclipse.equinox.common/META-INF/MANIFEST.MF, id: 206, type: org.eclipse.pde.api.tools.version_numbering, attributes: [apiMarkerID: 4, charEnd: 172, charStart: 154, description: - The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String, String) has been added
> - The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String, int, String, String, String) has been added
> - The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String) has been added
> - The non-abstract method org.eclipse.core.runtime.URIUtil.toURI(String, String, String, String, String) has been added
> , lineNumber: 5, message: The minor version should be incremented in version 3.16.200, since new APIs have been added since version  
> ```
> 
@iloveeclipse : Thanks a lot! Yet I couldn't find it in the logs. Was I looking at the wrong place?

iloveeclipse:
> Was I looking at the wrong place?

Yes, As first time contributor not all builds were executed for you without approval. I've approved and now you can see it in "continuous-integration/jenkins/pr-head" above: https://ci.eclipse.org/equinox/job/equinox/job/PR-52/3/display/redirect

mickaelistria:
@cmorty: your baseline (which is used for comparison) in the IDE, defined in Windows > Preferences > Plugin Development > API Baselines is most likely older than latest release. Please verify latest release is being used as baseline to get accurate versioning assistance.

bundles/org.eclipse.equinox.common.tests/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,7 @@ Automatic-Module-Name: org.eclipse.equinox.common.tests
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Bundle-ActivationPolicy: lazy
 Require-Bundle: org.junit,
- org.eclipse.equinox.common;bundle-version="3.15.100",
+ org.eclipse.equinox.common;bundle-version="3.17.000",
```
iloveeclipse, line 11: `3.17.000` is wrong. I told you to use `3.17.0`

bundles/org.eclipse.equinox.common/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.common; singleton:=true
-Bundle-Version: 3.16.200.qualifier
+Bundle-Version: 3.17.000.qualifier
```
iloveeclipse, line 5: `3.17.0.qualifier` please!

bundles/org.eclipse.equinox.common.tests/META-INF/MANIFEST.MF
cmorty: Ah ok. That was my mistake. Thanks.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
```diff
@@ -285,18 +299,105 @@ public static URI toURI(URL url) throws URISyntaxException {
...
+	 * @param scheme   Scheme name
+	 * @param ssp      Scheme-specific part
+	 * @param fragment Fragment
+	 * @return The new URI from the given components
+	 * @throws URISyntaxException If the URI string constructed from the given
+	 *                            components violates RFC 2396
+	 * @see java.net.URI#URI(String, String, String)
+	 * @since 3.17
+	 */
+	public static URI toURI(String scheme, String ssp, String fragment) throws URISyntaxException {
```
iloveeclipse, line 325: Is there any reason to have **3 new API** methods with lot of String arguments that could be wrongly used? Would it be enough to have one, or better leave all of them **private** to avoid confusion for API clients?

cmorty: The motivation is, that these are drop-in replacements for the original constructors which have lots of issues with UNC-Paths - thus they must have the same interface.

cmorty: Would it help to document that more thoroughly? I can also add tests that expect that exception. :)

cmorty:
I also incremented the minor of the test-bundle. Is that the right thing to do based on the error in https://ci.eclipse.org/equinox/job/equinox/job/PR-52/9/execution/node/49/log/ ?

cmorty:
@iloveeclipse Any chance of getting this merged? I know, it needs rebasing and the version-stuff probably needs adjusting, too. But is it worth putting an effort into it? Otherwise I'll probably put it somewhere into CDT, where it IMO does not belong.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
mickaelistria: As it seems mostly to be a fix, and that UNC is stanard enough, can't we just improve the existing methods to work as expected instead of introducing new ones?

cmorty: @mickaelistria  I'd love to. But this is a java-lib issue since forever (I'd have to check, but I think one of the bugs I found was filed against Java 5.0). All the tickets are wontfix, because it might break things. I've found multiple local workarounds in the CDT and Eclipse Codebase and it would be nice to have this at a central place.

mickaelistria: It's still not clear to me why new public method entry points in URIUtil are necessary. Can the existing ones in URIUtil be fixed to work better with UNC?

cmorty: I think we have a misunderstanding here. This is not about improving the existing implementation (which already has a spacial handling to convert URLs into URIs), but extending it, so it can also be used as drop-in replacement for the broken [java.net.URI-ctors](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#constructor_summary).

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/URIUtilTest.java
```diff
@@ -552,4 +561,74 @@ public void testMakeRelative() throws URISyntaxException {
...
+				"file:////WSL$/Ubuntu/",
+				"file:////WSL$/Ubuntu",
+				"file:////WSL$/Ubuntu/"
+				// @formatter:on
+		};
+
+		for (int i = 0; i < uris.length; i++) {
+			URI base = new URI(uris[i]);
+			URI resolved = URIUtil.append(base, relative);
+			assertEquals("1." + Integer.toString(i), expectedResolved, resolved);
```
laeubi, line 591: Can we please use **meaningfull** messages here? and of course also below?

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
laeubi: Looking at the referenced Bug descriptors, I can't see that anything is broken with URI constructor per se and it seems the only "value" is that you add more slashes. So why are these slashes missing in the first place? The referenced bugs seem to indicate inappropriate use of URI.normalize() and suggest using Path#normalize instead...

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/URIUtilTest.java
cmorty: Of course. But I oriented myself by the testcases that were already there....

laeubi: Please don't do so, that are just "bad" examples that are hard to understand if something fails, so you better adjust these so you would understand failures without the code...

cmorty: Would you be happy with inline comments?  I think explaining things in the exception is rather longish. And one would have to look at the code anyway to understand what is going on.

laeubi: I don't think there is any length restriction in the message :-)
In this case I think something like "Appending "+relative+" to base URI "+base+" yields wrong results" (if I understand correctly what is going on here!)

cmorty: OK. I tried to make the exceptions more expressive and also extended the test-cases to show that the URI-ctors actually fail.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
cmorty: I improved the commit message and the tests. I hope things are clearer now.

cmorty:
@laeubi , @mickaelistria , @iloveeclipse : In the latest version I tried to make the issue (buggy URI-ctor) clearer by adjusting the tests . I'd like to know whether there is a chance to get this merged. And I'm not talking about the code-quality, but the general issue of adding working URI-factories as drop-ins for all URI-ctors, although only some of them are broken. I don't think it is a good idea to only add the broken ones, as one has to remember when to use the factory and when to use the ctor.
I'm happy to discuss code quality, renaming the methods or improving the commit message and documentation. But if there is no realistic chance to get this merged, I'll just spend my limited free time elsewhere (Please do not understand this as criticism for potentially not merging this, I just want to avoid spending more time on this MR and then not get it merged after all.).

laeubi (approved):
Looks good from my side. But requires to resolve the conflicting files and a rebase.

vogella:
LGTM too, please update the PR

cmorty:
> LGTM too, please update the PR

Sorry. Limited time. ;) I'll try to get this updated this week.

cmorty:
@laeubi , @vogella : I finally got around rebasing. I hope I got the versioning right. For me as C++-Developer that's still a bit of magic. :)

vogella:
Thanks @cmorty , merging based on the +1 from @laeubi

Issue #56, opened 2022-06-23
Compilation failure in 4.25 I- and Y-Build: Y20220623-0900


sravanlakkimsetti:
Build failed with below error
```
 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0-SNAPSHOT:compile (default-compile) on project org.eclipse.equinox.console.ssh.tests: Compilation failure: Compilation failure: 
 [ERROR] /home/jenkins/agent/workspace/Y-build-4.25/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/gitCache/eclipse.platform.releng.aggregator/equinox/bundles/org.eclipse.equinox.console.ssh.tests/src/org/eclipse/equinox/console/ssh/SshCommandWithConfigAdminTests.java:[49] 
 [ERROR] 	import org.osgi.service.cm.ManagedService;
 [ERROR] 	       ^^^^^^^^^^^^^^^^^^^
 [ERROR] The import org.osgi.service.cm cannot be resolved
 [ERROR] /home/jenkins/agent/workspace/Y-build-4.25/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/gitCache/eclipse.platform.releng.aggregator/equinox/bundles/org.eclipse.equinox.console.ssh.tests/src/org/eclipse/equinox/console/ssh/SshCommandWithConfigAdminTests.java:[70] 
 [ERROR] 	private ManagedService configurator;
 [ERROR] 	        ^^^^^^^^^^^^^^
 [ERROR] ManagedService cannot be resolved to a type
 [ERROR] /home/jenkins/agent/workspace/Y-build-4.25/eclipse.platform.releng.aggregator/eclipse.platform.releng.aggregator/cje-production/gitCache/eclipse.platform.releng.aggregator/equinox/bundles/org.eclipse.equinox.console.ssh.tests/src/org/eclipse/equinox/console/ssh/SshCommandWithConfigAdminTests.java:[97] 
 [ERROR] 	when(context.registerService(any(String.class), any(ManagedService.class), any(Dictionary.class)))
 [ERROR] 	             ^^^^^^^^^^^^^^^
 [ERROR] The method registerService(String[], Object, Dictionary<String,?>) in the type BundleContext is not applicable for the arguments (String, any(ManagedService.class), Dictionary)
[...]

PR #59, opened 2022-06-24
 Import org.osgi.service.cm in console.ssh.tests 

akurtakov:
Bundle o.e.equinox.console.ssh has optional import but tests need it
unconditionally.
Fixes https://github.com/eclipse-equinox/equinox/issues/56

### DIFF
--- a/bundles/org.eclipse.equinox.console.ssh.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.console.ssh.tests/META-INF/MANIFEST.MF
@@ -13,6 +13,7 @@ Import-Package: org.apache.sshd.client,
  org.junit;version="4.8.1",
  org.mockito,
  org.mockito.stubbing,
- org.mockito.invocation
+ org.mockito.invocation,
+ org.osgi.service.cm
 Fragment-Host: org.eclipse.equinox.console.ssh, org.eclipse.equinox.console
 Automatic-Module-Name: org.eclipse.equinox.console.ssh.tests


laeubi:
> Bundle o.e.equinox.console.ssh has optional import

So they seem not really optional, given that org.osgi.service.cm is a standad osgi service, why not make it non-optional here?

akurtakov:
> > Bundle o.e.equinox.console.ssh has optional import
> 
> So they seem not really optional, given that org.osgi.service.cm is a standad osgi service, why not make it non-optional here?

https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.equinox.console.ssh/src/org/eclipse/equinox/console/ssh/SshCommand.java#L69 makes it look really optional and enabled via a flag in the console.ssh bundle. It's the test fragment that needs it unconditionally.
If you want to drive bigger change in console to always use - please do via another PR.

laeubi:
> f you want to drive bigger change in console to always use - please do via another PR.

nope, just wondering if it is really worth the effort to have it optional...

akurtakov:
> > f you want to drive bigger change in console to always use - please do via another PR.
> 
> nope, just wondering if it is really worth the effort to have it optional...

That's a good question but my knowledge on the topic is really low to have a say.

bundles/org.eclipse.equinox.console.ssh.tests/META-INF/MANIFEST.MF
```diff
@@ -13,6 +13,7 @@ Import-Package: org.apache.sshd.client,
  org.junit;version="4.8.1",
  org.mockito,
  org.mockito.stubbing,
- org.mockito.invocation
+ org.mockito.invocation,
+ org.osgi.service.cm
 Fragment-Host: org.eclipse.equinox.console.ssh, org.eclipse.equinox.console
```
HannesWell, line 18: Does it really work to specify multiple host bundles for a fragment?
I've never seen this before but when thinking about it, at least the Wiring API allows an arbitrary number of host wires. But until now I just thought it was an hypothetical scenario due to the generality of the Wiring API.

HannesWell (commented):
> > > Bundle o.e.equinox.console.ssh has optional import
> > 
> > 
> > So they seem not really optional, given that org.osgi.service.cm is a standad osgi service, why not make it non-optional here?
> 
> https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.equinox.console.ssh/src/org/eclipse/equinox/console/ssh/SshCommand.java#L69 makes it look really optional and enabled via a flag in the console.ssh bundle. It's the test fragment that needs it unconditionally. If you want to drive bigger change in console to always use - please do via another PR.

Considering that I agree that this is the way to go.

bundles/org.eclipse.equinox.console.ssh.tests/META-INF/MANIFEST.MF
akurtakov: Last time I touched that part was for another build breakage probably. I wish I could help more but ... :)

rotty3000:
Multiple hosts is completely legitimate.

Issue #7, opened 2022-06-01
Exception saving preferences in p2 tasks


iloveeclipse:
While executing p2 tasks from command line we see on 4.24 M3 following errors reported:
```
!ENTRY org.eclipse.equinox.preferences 4 4 2022-06-01 08:42:59.310
!MESSAGE Exception saving preferences to: /workspaces/socbm775/sandreev/ws-r7.6-ewc425/opt/93000/src/segments/eclipse/eclipseRE/p2/org.eclipse.equinox.p2.engine/profileRegistry/SDKProfile.profile/.data/.settings/org.eclipse.equinox.p2.artifact.repository.prefs.
!STACK 0
java.nio.file.NoSuchFileException: /workspaces/socbm775/sandreev/ws-r7.6-ewc425/opt/93000/src/segments/eclipse/eclipseRE/p2/org.eclipse.equinox.p2.engine/profileRegistry/SDKProfile.profile/.data/.settings/org.eclipse.equinox.p2.artifact.repository.prefs
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
        at java.base/sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:468)
        at java.base/sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:267)
        at java.base/java.nio.file.Files.move(Files.java:1422)
        at org.eclipse.core.internal.preferences.EclipsePreferences.write(EclipsePreferences.java:289)
        at org.eclipse.core.internal.preferences.EclipsePreferences.save(EclipsePreferences.java:1066)
        at org.eclipse.equinox.internal.p2.engine.ProfilePreferences.doSave(ProfilePreferences.java:141)
        at org.eclipse.equinox.internal.p2.engine.ProfilePreferences$SaveJob.run(ProfilePreferences.java:54)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

Most likely regression from https://github.com/eclipse-equinox/equinox.bundles/pull/8

tjwatson:
@HannesWell can you provide a PR to address this issue?

iloveeclipse:
The problem is still there. 
```
[...]

PR #65, opened 2022-06-27
Move preference files as atomic-operation

HannesWell:
Fixes https://github.com/eclipse-equinox/equinox/issues/7

The strategy to fix the `IOException` due to concurrent access is to synchronize writes of multiple threads for the same `EclipsePreferences` object. Otherwise it is not possible to ensure that the backup file is written correctly and then moved to replace the previous file withouth another thread interferring with that operation.

Is it correct that each EclipsePreferences instance represents on one preference file? If yes we can simply use one lock object per Preferences-instance to syncronize writes. Instead of a new object the `childAndPropertyLock` could be reused.

Looking into the call hierarchy reveals that `EclipsePreferences.write()` is only called via `EclipsePreferences.flush()`, `ProjectPreferences.flush()` and `ProfilePreferences.SaveJob.run()`. The first one already holds the `childAndPropertyLock` when calling `EclipsePreferences.write()`. The latter two don't.
The stack-trace Andrey presented shows that in his case the write was initiated from `ProfilePreferences.SaveJob.run()`, which I think is an indication that synchronization prevent the problem.


iloveeclipse:
Sounds reasonable. 
If merged I could verify (we build & run tests every week on latest nightly platform build from Thursday).

iloveeclipse (approved):
LGTM

HannesWell:
> Sounds reasonable. If merged I could verify (we build & run tests every week on latest nightly platform build from Thursday).

Did this solve the issue?

iloveeclipse:
I will check it on Monday.

PR #66, opened 2022-06-28
Touch o.e.equinox.server.jetty for Jetty 10.0.11 update

akurtakov:


akurtakov:
Verification build can't be successful without touching the jetty feature.

laeubi:
@akurtakov instead of `<plugin>` you might want to use (probably with upper version range)
```
<requires>
  <import plugin="... the jetty bundle..."/>
</requires>
```

This is a new feature (or fix) in Tycho that selects the bundles at install time rather than compile time making it unnecessary to bump such features...

akurtakov:
I'll give it a go.

PR #67, opened 2022-06-28
Switch Jetty to requires in feature.xml

akurtakov:
That should avoid having to touch the feature everytime Jetty version is
changed.

### DIFF
--- a/features/org.eclipse.equinox.server.jetty/feature.xml
+++ b/features/org.eclipse.equinox.server.jetty/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.equinox.server.jetty"
       label="%featureName"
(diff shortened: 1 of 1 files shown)


akurtakov:
@sravanlakkimsetti Any reason no to go that path?

merks:
Might this result in these being missing from the repository?  I'd test that to be sure that's not a problem like perhaps this one is  problem:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/365

akurtakov:
Pushing this one and starting I-build - hopefully everything works that way.

laeubi:
@akurtakov Thanks for trying this out! This should greatly simplify referencing non-reactor dependencies in features now!

PR #68, opened 2022-06-30
Address encountered deadlock issues with service use.

castortech:
This code was added at our end over 10 months ago, and has had 1 minor fix 5 months ago and been used in production, as we wanted to make sure that no issue were seen over time. Now that we have a new release to sync with, it's time to create a formal PR.

Using a reentrant lock (extended default only for special toString) on ServiceUse instead of simple synchronized blocks.

Also added quite a bit of logging to help tracing.


HannesWell:
Did you create this PR in order to address https://github.com/eclipse-equinox/equinox/issues/15 respectively https://github.com/eclipse-equinox/equinox/pull/31?
Because it sounds very related.

castortech:
> Did you create this PR in order to address #15 respectively #31? Because it sounds very related.

I didn't, as I said for us this was about 10 months ago, but it does look very very similar. Call that a coincidence.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -28,22 +32,66 @@
...
 	/** ServiceDescription of the registered service */
 	final ServiceRegistrationImpl<S> registration;
+	final Debug debug;
 
 	/** bundle's use count for this service */
 	/* @GuardedBy("this") */
 	private int useCount;
 
+	/** lock for this service */
+	private ReentrantLockExt lock = new ReentrantLockExt();
```
bjhargrave, line 46:
```suggestion
	private final ReentrantLock lock = new ReentrantLockExt();
```

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -28,22 +32,66 @@
...
+	public static final int DEADLOCK = 1001;
+
+	private int lockTimeout = 10;
```
bjhargrave, line 49:
```suggestion
	/** in seconds */
	private final int lockTimeout = 10;
```

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceRegistrationImpl.java
```diff
@@ -535,28 +537,54 @@ S getService(BundleContextImpl user, ServiceConsumer consumer) {
...
+				throw new ServiceException(
+						"Failed to acquire lock within try period, Lock, id:" + System.identityHashCode(useLock) + ", " //$NON-NLS-1$ //$NON-NLS-2$
+								+ useLock.toString(),
+						ServiceUse.DEADLOCK);
+			} catch (InterruptedException e) {
+				Thread.currentThread().interrupt(); // To appease sonar
+				return null;
+			} finally {
+				if (useLock.isHeldByCurrentThread()) {
+					useLock.unlock();
```
bjhargrave, line 586: This finally should be part of a try/finally block in the if tryLock block and always call unlock. It is too risky here.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -28,22 +32,66 @@
...
 	final ServiceRegistrationImpl<S> registration;
+	final Debug debug;
 
 	/** bundle's use count for this service */
 	/* @GuardedBy("this") */
 	private int useCount;
 
+	/** lock for this service */
+	private ReentrantLockExt lock = new ReentrantLockExt();
+	public static final int DEADLOCK = 1001;
```
bjhargrave, line 47: This needs javadoc. It is a new type for ServiceException.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -28,22 +32,66 @@
...
+				lock.unlock();
+			}
+		}
+	}
+
+	int getLockTimeout() {
+		return lockTimeout;
+	}
+
+	ReentrantLockExt getLock() {
```
bjhargrave, line 93:
```suggestion
	ReentrantLock getLock() {
```

There is no need to expose the impl details of this lock (that it is some specific subclass.)

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -28,22 +32,66 @@
...
+			throw new ServiceException(
+					"Failed to acquire lock within try period, Lock, id:" + System.identityHashCode(lock) + ", " //$NON-NLS-1$ //$NON-NLS-2$
+							+ lock.toString(),
+					DEADLOCK); // $NON-NLS-1$
+		} catch (InterruptedException e) {
+			Thread.currentThread().interrupt(); // To appease sonar
+			return null;
+		} catch (Exception e) {
+			throw new RuntimeException("Callable threw exception", e); //$NON-NLS-1$
+		} finally {
```
bjhargrave, line 82: Same as my other comment. The finally block needs to be in a try/finally inside the if tryLock and always call unlock.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceRegistrationImpl.java
```diff
@@ -535,28 +537,54 @@ S getService(BundleContextImpl user, ServiceConsumer consumer) {
...
 
 			/* Obtain and return the service object */
-			synchronized (use) {
-				/* if another thread removed the ServiceUse, then
-				 * go back to the top and start again */
-				synchronized (servicesInUse) {
-					user.checkValid();
-					if (servicesInUse.get(this) != use) {
-						continue;
+			ReentrantLockExt useLock = use.getLock();
```
bjhargrave, line 540:
```suggestion
			ReentrantLock useLock = use.getLock();
```

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceRegistrationImpl.java
```diff
@@ -627,9 +657,16 @@ boolean ungetService(BundleContextImpl user, ServiceConsumer consumer, S service
 			}
 		}
 
-		boolean result;
-		synchronized (use) {
-			result = consumer.ungetService(use, serviceObject);
+		return use.callWithLock(() -> {
+			if (registry.debug.DEBUG_SERVICES) {
+				ReentrantLockExt useLock = use.getLock();
```
bjhargrave, line 662:
```suggestion
				ReentrantLock useLock = use.getLock();
```

bjhargrave (changes requested):
Some requested changes.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -28,22 +37,68 @@
...
+		}
+	}
+
+	Duration getLockTimeout() {
+		return lockTimeout;
+	}
+
+	ReentrantLock getLock() {
+		return lock;
 	}
```
HannesWell, line 102:
```suggestion
	interface Locked extends AutoCloseable {
		@Override
		void close();
	}

	Locked lock() throws InterruptedException {
		if (lock.tryLock(lockTimeout, TimeUnit.MILLISECONDS)) {
			if (debug.DEBUG_SERVICES) {
				Debug.println('[' + Thread.currentThread().getName() + "] getServiceLock[" //$NON-NLS-1$
						+ context.getBundle() + "](" + this + "), id:" + System.identityHashCode(lock) //$NON-NLS-1$ //$NON-NLS-2$
						+ ". holdCount:" + lock.getHoldCount() //$NON-NLS-1$
						+ ", queued:" + lock.getQueueLength()); //$NON-NLS-1$
			}
			return lock::unlock;
		}
		throw new ServiceException(
				"Failed to acquire lock within try period, Lock, id:" + System.identityHashCode(lock) + ", " //$NON-NLS-1$ //$NON-NLS-2$
						+ lock.toString(),
				ServiceException.DEADLOCK);
	}
```
What do you and the others think about using the `lock()` method proposed above. It would allow to use the 'locked-state' as resource of a try-block in a safe manner, at the three-locations where `use.getLock()` is now used:
```
try (Locked locked = use.lock()) {
 // perform the protected action
} catch (InterruptedException e) {
  Thread.currentThread().interrupt(); // To appease sonar
  // if necessary return a reasonable default
}
```
More or less `try (Locked locked = use.lock()) {` would simply replace `synchronized (use) {`.
This would make the `runWithLock()` methods obsolete since other exceptions can simply be propagated outside the try-block and would also simplify the lock's use in `ServiceRegistrationImpl.getService()`. Furthermore code duplications would be avoided (e.g. the 'DeadLock'-`ServiceException`).

bundles/org.eclipse.osgi/osgi/src/org/osgi/framework/ServiceException.java
```diff
@@ -80,6 +80,12 @@ public class ServiceException extends RuntimeException {
 	 * @since 1.8
 	 */
 	public static final int	ASYNC_ERROR			= 7;
+	/**
+	 * A deadlock occurred during service registration.
+	 * 
+	 * @since 1.8
+	 */
+	public static final int DEADLOCK = 8;
```
HannesWell, line 88: This class is a copy of the OSGi Framework reference implementation and we don't want to modify it in Equinox-specific ways.
That being said I think it would be useful to have a dedicated error code for this situation so that SCR-Actors like org.apache.felix.scr could handle this situation gracefully.
But this has to be added to the OSGi spec first.

HannesWell (changes requested):
> > Did you create this PR in order to address #15 respectively #31? Because it sounds very related.
> 
> I didn't, as I said for us this was about 10 months ago, but it does look very very similar. Call that a coincidence.

Looks like a coincidence, yes. :)
Great, I just thought that we should soon work to fix that. Didn't expect to have a PR that fast. So thanks already.

Can you fetch https://github.com/eclipse-equinox/equinox/pull/31 and report if this _fixes_ it?

You solve the deadlock by throwing a ServiceException if the lock cannot be obtained after a fixed timeout has elapsed. IIRC this was also an approach discussed in https://github.com/eclipse-equinox/equinox/issues/15.
So does the start-up then simply fail or will the SCR-actor perform another attempt to active a component later?
@tjwatson just in case it is not yet implemented: Do you think Felix-SCR could handle a Service-Exception caused by a dead-lock specifically by just re-attempting to activate the root component?

Besides that I think it would be even better if we had a 'real' dead-lock detection, meaning that the chain of _uses_ in this thread is tracked and if one lock cannot be obtained it is checked if there is another thread with a 'crossing' chain in the opposite direction.
I think this could be achieved by maintaining a set of locked uses per thread that is updated when the lock is acquired and released. When a lock cannot be acquired it can then be checked if the thread holding the desired lock is waiting for a lock held by this thread. The latter would also require a Mapping lock->owner, but since you already subclass `ReentrantLock` the protected method can be made visible.

Having a 'real' dead-lock detection would be more reliable (maybe some components take just very long to activate) and often faster (for actually quickly activating components waiting for 10s is just waste of time).

HannesWell:
It is not tested and maybe I have missed something, but I think the following code in `ServiceUse` could be a real deadlock detection:
```
	private final ReentrantLockExt lock = new ReentrantLockExt(); // needs an additional override of getOwner() to make it visible

	interface Locked extends AutoCloseable {
		@Override
		void close();
	}

	private static final ConcurrentMap<Thread, ReentrantLock> AWAITING_LOCKS = new ConcurrentHashMap<>();

	private final ThreadLocal<Set<ReentrantLock>> ownedLocks = ThreadLocal.withInitial(HashSet::new);

	Locked lock() throws InterruptedException {
		boolean setAwaiting = false;
		do  {
			if (lock.tryLock(10, TimeUnit.MILLISECONDS)) { // use a short time-out to not spin-wait for the lock to become available
				ownedLocks.get().add(lock);
				if (setAwaiting) {
					AWAITING_LOCKS.remove(Thread.currentThread());
				}
				if (debug.DEBUG_SERVICES) {
					Debug.println('[' + Thread.currentThread().getName() + "] getServiceLock[" //$NON-NLS-1$
							+ context.getBundle() + "](" + this + "), id:" + System.identityHashCode(lock) //$NON-NLS-1$ //$NON-NLS-2$
							+ ". holdCount:" + lock.getHoldCount() //$NON-NLS-1$
							+ ", queued:" + lock.getQueueLength()); //$NON-NLS-1$
				}
				return () -> {
					ownedLocks.get().remove(lock);
					lock.unlock();
				};
			}
			AWAITING_LOCKS.put(Thread.currentThread(), lock);
			setAwaiting = true;
			Thread owner = lock.getOwner();
			if (owner != null) { // lock could be released in the meantime
				ReentrantLock crossingLock = AWAITING_LOCKS.get(owner);
				if (ownedLocks.get().contains(crossingLock)) {
					throw new ServiceException(
							"Failed to acquire lock within try period, Lock, id:" + System.identityHashCode(lock) + ", " //$NON-NLS-1$ //$NON-NLS-2$
									+ lock.toString(),
							ServiceException.DEADLOCK);
				}
			}
			// Not a dead-lock. Lock was regularly hold by another thread. Just try again
		} while (true);
	}
```

bjhargrave:
I have forced pushed to this PR with a number of updates and improvements.

bjhargrave:
FYI, I authored all this ServiceUse code, so I am very familiar with it.

bjhargrave:
I don't understand enough to know what the build error means and if it related to changes in this PR.

> Error:  Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:3.0.0-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.equinox.supplement: Only qualifier changed for (org.eclipse.equinox.supplement/1.10.500.v20220630-2245). Expected to have bigger x.y.z than what is available in baseline (1.10.500.v20220516-2155) -> [Help 1]

laeubi:
@bjhargrave you need to increment the version of org.eclipse.equinox.supplement in the maninfest (and probably pom if there is any) from 1.10.500 -> 1.10.600

bjhargrave:
> java.lang.AssertionError: The StaticCircuit2 component shouldn't have bound objects expected:<0> but was:<1>
	at org.eclipse.equinox.ds.tests.tbc.DSTest.testCircularityHandling(DSTest.java:1426)

This test failure may be the result of the PR changing to break the deadlock using the timeout on the ReentrantLock.

tjwatson:
> This test failure may be the result of the PR changing to break the deadlock using the timeout on the ReentrantLock.

This test seems to fail only in the github actions, no one has been able to determine why.  The test run on eclipse CI looks more suspicious https://ci.eclipse.org/equinox/job/equinox/job/PR-68/7/

But I don't think it is related. It has this deadlock caused by some issue with closing the framework storage:

```
    java.base@17.0.1/java.io.FileDescriptor.close0(Native Method)
    java.base@17.0.1/java.io.FileDescriptor.close(FileDescriptor.java:297)
    java.base@17.0.1/java.io.FileOutputStream$1.close(FileOutputStream.java:392)
    java.base@17.0.1/java.io.FileDescriptor.closeAll(FileDescriptor.java:355)
    java.base@17.0.1/java.io.FileOutputStream.close(FileOutputStream.java:390)
    org.eclipse.osgi.storagemanager.StorageManager.closeOutputStream(StorageManager.java:912)
    org.eclipse.osgi.storagemanager.ManagedOutputStream.close(ManagedOutputStream.java:56)
```

This has been a known issue from the past, but I've not seen it happen on the eclipse CI for a long time.

tjwatson:
I worry about the performance penalty of this change for such a high use code path.  Particularly the move away from simple synchronization and the use of many new lambdas.

We will need to do some performance analysis on this change @jhanders34

HannesWell:
> I worry about the performance penalty of this change for such a high use code path. Particularly the move away from simple synchronization and the use of many new lambdas.
> 
If we use the AutoClosable-lock approach suggested above (https://github.com/eclipse-equinox/equinox/pull/68#discussion_r911444032) there is only one location with a lambda, which can be easily replaced by a corresponding anonymous class.

laeubi:
As mentioned in the other change, I think it would be better trying to move to a more lock free approach than using time-outs.
I even wonder if according to the OSGi spec a caller has to expect such "unusual" exceptions? I think none of the documented options in `org.osgi.framework.BundleContext.getService(ServiceReference<S>)` allows for such, even returning `null` might be arguable as the server obviously *is* registered.

castortech:
> 

Regarding the timeout, I can only say that when we captured debugging and analyzed the results, we could never see any delay during normal use and even cases that would produce deadlock would have very low delay. The original synchronized would obviously have delays of its own. IMHO the biggest question is the choice of the 10 secs timeout, which was arbitrary and might be better at a much lower value.

bjhargrave:
> Particularly the move away from simple synchronization

I am not sure this is wise either. But calling ServiceFactories has always been an issue. We must prevent multiple concurrent calls to a factory (so I am not sure how a lock free design can handle that), so we hold a mutex lock. Using intrinsic locks is simple but provides no way to back off when a timeout elapsed which can be evidence of a deadlock. Using a ReentrantLock allows a timeout so we can backoff and throw an exception indication the ServiceFactory could not be called.

 

> the use of many new lambdas.

There are now 3 lambda expressions. They can be replaced with anonymous classes if necessary.

> I even wonder if according to the OSGi spec a caller has to expect such "unusual" exceptions? I think none of the documented options in `org.osgi.framework.BundleContext.getService(ServiceReference<S>)` allows for such, even returning `null` might be arguable as the server obviously _is_ registered.

ServiceException can always be thrown. I think returning `null` is wrong for a timeout as it wallpapers over a real issue. Throwing an exception allows the caller to know something is wrong and tell someone (e.g. log it).


> the biggest question is the choice of the 10 secs timeout, which was arbitrary and might be better at a much lower value.

The challenge is to distinguish between a real deadlock rather then a very slow ServiceFactory implementation which is running on another thread. For the latter, we want to wait for it. For the former, the deadlock will never resolve (or it wouldn't be a deadlock), so waiting a long time is of no great harm.  I would be more concerned that 10 seconds is not long enough to be convinced that it is not just a very slow ServiceFactory implementation.

HannesWell:
> > the biggest question is the choice of the 10 secs timeout, which was arbitrary and might be better at a much lower value.
> 
> The challenge is to distinguish between a real deadlock rather than a very slow ServiceFactory implementation which is running on another thread. For the latter, we want to wait for it. For the former, the deadlock will never resolve (or it wouldn't be a deadlock), so waiting a long time is of no great harm. I would be more concerned that 10 seconds is not long enough to be convinced that it is not just a very slow ServiceFactory implementation.

As mentioned before I share this concern, therefore I suggested to use a 'real' deadlock detection, for which I described a prototype above: https://github.com/eclipse-equinox/equinox/pull/68#issuecomment-1171721317
I have neither tested nor used it, so it might be flawed, but from my theoretical thinking it should work. But let me know if you see a problem.

bjhargrave:
> But let me know if you see a problem.

The main issue is that it is a big hammer to hit an very rare problem. It uses N+1 maps (N being the number of threads which call this code) and the maps are pinned to the thread via ThreadLocals. With changes coming to Java for virtual threads, we should steer clear of adding more ThreadLocals.

laeubi:
> so I am not sure how a lock free design can handle that

@HannesWell has made some experiments, and the deadlock could be prevented by putting a break point at the right place, so it is not really an unavoidable dead-lock but somewhere in the locking-chain there is an incomplete locking of required objects.

From the last investigations, I found that the sync is just done for too many things (e.g. service use increments) and also for factories and for service, while e.g. for a `ServiceFactory` one should actually lock on the `BundleContext` as far as I can see and for a plain service there need no sync at all ...

@bjhargrave I wonder if you as the author of this code could tell how well it is covered by test, e.g. if one wants to adjust the locking model used here, do we need manual tests or can we rely on the automatic ones so if build is okay we are quite certain nothing breaks here?

bjhargrave:
> I wonder if you as the author of this code could tell how well it is covered by test

I don't know how extensive the build tests are.

The code has been used for many years in many production environments. So deep care must be taken with any changes here.

> while e.g. for a `ServiceFactory` one should actually lock on the `BundleContext` as far as I can see

I assume you mean the user's bundle context. That is too coarse. That means a bundle can only get one service at a time (SCR might be upset). ServiceUse is the combination of the user of the service and the service which is why it is the lock object. It provides the finest grain for locking and, in general, should not be contended.

laeubi:
> I assume you mean the user's bundle context. That is too coarse. That means a bundle can only get one service at a time (SCR might be upset)

Better slow/upset than dead-locking ;-)
The problem arise if two requestor of a service crossing dependency chains, then it is not a problem of contention... see https://github.com/eclipse-equinox/equinox/issues/15 for more details.

HannesWell:
> > But let me know if you see a problem.
> 
> The main issue is that it is a big hammer to hit an very rare problem. It uses N+1 maps (N being the number of threads which call this code) and the maps are pinned to the thread via ThreadLocals. With changes coming to Java for virtual threads, we should steer clear of adding more ThreadLocals.


It is indeed a big hammer. Do you know if there is a better approach? I mean deadlock detection is actually a quite generic programming problem (not even Java specific), so I assume that some smart people already got better ideas than I did when thinking ~15min about the problem.

IIRC Eclipse Jobs can detect deadlocks? But probably only between jobs-rules? Or is this something we can inherit.


> > so I am not sure how a lock free design can handle that
> 
> @HannesWell has made some experiments, and the deadlock could be prevented by putting a break point at the right place, so it is not really an unavoidable dead-lock but somewhere in the locking-chain there is an incomplete locking of required objects.

I want to clarify that I had to put the break-point and with that hold back thread B back before it obtained the lock that is later wanted by thread A. So basically I manually scheduled the threads so that one can complete its work before the other ones starts it.
If we want to implemented that it will probably be more complicated and has a worse performance (is my guess, but I might be wrong) than just detecting it and aborting with an exception.

laeubi:
> If we want to implemented that it will probably be more complicated and has a worse performance (is my guess, but I might be wrong) than just detecting it and aborting with an exception.

I don't think that this is a performance problem

1. we don't see heavy multi-treading here, actually two threads, bigger setups might have a few more but realistically acquiring services with hundreds of threads concurrently is not really a usual use-case
2. Often one service acquires the other and activates and we have more a linear chain
3. the performance penalty of starting all over again (**if the code actually is able to recover!**) is much higher, e.g. how/when should a failed lockup retried? Maybe with a backoff-algorithm? Compared to waiting a bit more for something I need to wait anyways (given we not have dead-locked) is nothing much to gain

laeubi:
> That cannot be known until we run performance measurements.

As a deadlock has a performance of `O(∞)` we can't get worse here ;-)

> Only that we need to understand the cost of this for something that is arguably a bug in the design of service components involved in the deadlock and perhaps should be addressed by SCR itself

It is only a bug if it is forbidden by the spec, I don't see why it should be forbidden if you claim that we should not make any assumptions about number and/or concurrency level...

castortech:
> On the other hand I see the appeal to having the deadlock break out from the framework level. But the result isn't going to fix the components involved in the deadlock. They will still be in a broken state, correct?

Since I'm not the Equinox expert here, I might be missing the point, but I can say that with the patch that I submitted, we have never again see any deadlock being reported or reproduced. The deadlock exception has never occurred and so we never had to deal with recovery from the deadlocks that are now gone.

bjhargrave:
> They will still be in a broken state, correct?

Yes.

> So if equinox decides to use to fine grained sync to boost performance,

To coarsen the grain will have some effects beyond possibly avoiding deadlocks. It is hard to know what they will be without extensive testing in a variety of production use cases.

It is also unclear that coarsening the grain will solve all deadlock cases since a service factory implementation can do anything including using threads which may call back into the service registry and bump into the coarse lock held by the framework when calling the service factory.

I would not support changing to coarsen the grain. At most, we can attempt to break the deadlock by throwing an exception to indicate a deadlock was broken. But I am not even sure this is worth it since the components are still not in the desired state and someone will have to sort it out.

bjhargrave:
> no "bad" service factory involved

I am not talking about a bad service factory implementation here. The use of thread calling back into the service registry can be indirect based upon code that the factory implementation calls. A service factory implementation can depend upon other services and it had no control or idea over what those services do.

bjhargrave:
> The deadlock described [here](https://github.com/eclipse-equinox/equinox/issues/15#issuecomment-1157432597) even do not involve **any** `ServiceFactory`, just plain services!

SCR is involved and SCR uses service factories for all components.

HannesWell:
> > But let me know if you see a problem.
> 
> The main issue is that it is a big hammer to hit an very rare problem. It uses N+1 maps (N being the number of threads which call this code) and the maps are pinned to the thread via ThreadLocals. With changes coming to Java for virtual threads, we should steer clear of adding more ThreadLocals.

I thought about it a bit more and actually the `ThreadLocal` is not necessary. What we want to know is if the current thread is waiting for a lock, which is currently hold by another thread, that in turn waits for a lock that is hold by this thread. To check the last part, in my first propsal, a thread-local set of acquired locks was used. But actually that is not necessary because we can simply check if the current thread is the owner of the lock that the other thread is waiting for.
Consequently only a `ConcurrentMap<Thread, ServiceUseLock> AWAITED_LOCKS` remains and that map is only touched if the lock could not be acquired at the first attempt. Since we have to wait here any ways we cannot optimize the performance and effectively get the deadlock detection for free.

Besides that I think the code will be easier and no lambdas are necessary if the `ServiceUseLock` implements `AutoCloseable` and gets unlocked in its `close()` method and we provide a corresponding `lock()` method in `ServiceUse` that returns the ServiceUseLock as soon as it is successfully acquired.

We can then use `try (ServiceUseLock locked = use.lock()) {` as drop in replacement for `synchronized (use) {`. This also behaves the same regarding interruption.
I prepared an updated version of this PR that contains all the described changes:
https://github.com/HannesWell/equinox/commits/fix15_SCR_Deadlock

On that branch commits of each author are squashed to not record the back and forth but to maintain the original authorship as far as possible.
Feel free to (force-)push it to this PR branch if you and the others think that's suitable.

We should also try this change with the tests I provided in https://github.com/eclipse-equinox/equinox/pull/31
I'm on vacation until Friday but when I'm back I can do that.

bjhargrave:
> Feel free to (force-)push it to this PR branch if you and the others think that's suitable.

I really like the change to use try-with-resources. It is much cleaner than using callbacks. So I would go ahead and force push your latest work to this PR so I can easily comment on some things.

As for the deadlock detection logic, it only detects a two-party deadlock. A->B->A. It could be enhanced to look for larger party deadlocks. A->B->C->A.

bjhargrave:
> Glad you like it and please go ahead.

OK, I took your work and updated it. There were some issues such as improper handling of the thread interrupt status.



> Good Point, that should be simply feasible when further following the chain and considering that we can hit a cycle/deadlock the current thread ist not involved.

I also made this change. The detection code now walks a chain if locks looking for the current thread.

bjhargrave:
@castortech Your ECA is not on file at Eclipse. https://accounts.eclipse.org/legal/eca/validation/110972

Can you please take are of getting this done?

castortech:
> @castortech Your ECA is not on file at Eclipse. https://accounts.eclipse.org/legal/eca/validation/110972
> 
> Can you please take care of getting this done?

I had at the beginning updated it. Just did again, Redid the ECA and got successful message and also updated profile to add github account info if needed and updated with profile.

bjhargrave:
> I had at the beginning updated it. Just did again, Redid the ECA and got successful message and also updated profile to add github account info if needed and updated with profile.

Something is still wrong with your ECA. The commit author `Alain Picard <apicard@benchmarkconsulting.com>` is not understood by the ECA validator.

castortech:
> > I had at the beginning updated it. Just did again, Redid the ECA and got successful message and also updated profile to add github account info if needed and updated with profile.
> 
> Something is still wrong with your ECA. The commit author `Alain Picard <apicard@benchmarkconsulting.com>` is not understood by the ECA validator.

Emailed you directly to avoid polluting this thread

HannesWell:
> > Glad you like it and please go ahead.
> 
> OK, I took your work and updated it. There were some issues such as improper handling of the thread interrupt status.

What specifically was wrong with the previous approach?
Now the current thread is interrupted again after the lock was released. With my previous approach the thread was interrupted as soon as the lock method returns, to effectively preserve the interrupted state of the thread. The intention was to mimic an intrinsic synchronized that also does not alter the interrupted state of the thread, i.e. of the thread is interrupted before or while acquiring the lock the thread is also interrupted in the beginning of the synchronized-block?

With the current change this behavior is changed and given that we potentially have deeply nested locks I'm not sure if this is a good idea.

> 
> > Good Point, that should be simply feasible when further following the chain and considering that we can hit a cycle/deadlock the current thread ist not involved.
> 
> I also made this change. The detection code now walks a chain if locks looking for the current thread.

The way it is currently implemented let the current thread spin in an endless loop if the awaited locks depends on other locks dead-locked in a cycle but the current-thread does not participate in the actual cycle: e.g. A->B->C->B.

That loop is only broken if the deadlock is detected and resolved by throwing the ServiceException.
Nevertheless I think we should consider to check if the current thread really participates in the deadlock-cycle or only depends on the cycle. I implemented a solution based on the algorithms used to find strongly connected components in cyclic-graphs.
I'm about to push that commit to this PR.


For the further procedure: I intentionally worked out my suggestions as first commit of the commit chain and squashed the previous four commits into two, to avoid the back and forth in the git history and assumed you would use that.
Or  do you prefer to keep it as it is? IMHO preserving the current (five) commit history only makes it harder to follow the changes later. I only kept the three to reflect the original (main) authorship of each change.

bjhargrave:
> With the current change this behavior is changed and given that we potentially have deeply nested locks I'm not sure if this is a good idea.

It is not entirely clear to me what is correct here. Normally in these situations, I restore the interrupt state after then locked operation completes. This way the locked operation is not "broken" by the interrupt state. But with intrinsic locks this just flows in so I can see your argument also has merit.

> The way it is currently implemented let the current thread spin in an endless loop

It was a first attempt :-) Your change though makes the detection much more expensive. Sets and queues are created, manipulated, and iterated over. I think we need a cheaper test (or accept we cannot detect all possible deadlocks).

I think we still also have to agree that this PR is worth doing. I am still on the fence.

> For the further procedure: I intentionally worked out my suggestions as first commit of the commit chain and squashed the previous four commits into two, to avoid the back and forth in the git history and assumed you would use that.

I am not sure what you mean here? Your commits were not on the tip of the PR branch, so I manually pulled in your changes and went from there. I also rebased on the tip of master and also rebased to fix the author for @castortech to pass the ECA validation. 

As we incrementally improve, we should probably just add commits to the tip and not squash arbitrary commits. Once we get close to done and also agree that this PR should even be merged, then we should squash all the commits into a single commit before merging.

bjhargrave:
> Your change though makes the detection much more expensive. Sets and queues are created, manipulated, and iterated over. I think we need a cheaper test (or accept we cannot detect all possible deadlocks).

I think a simple change to my loop will avoid the infinite looping issue:

```diff
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
@@ -260,7 +260,9 @@ public class ServiceUse<S> {
                                clearAwaitingLock = true;
                                // Check if the current thread is in the lock chain.
                                Thread owner = useLock.getOwner();
-                               for (ServiceUseLock crossingLock; (owner != null) // lock could be released in the meantime
+                               int maxCycles = AWAITED_LOCKS.size();
+                               for (ServiceUseLock crossingLock; (maxCycles-- > 0) // Prevent infinite loop
+                                               && (owner != null) // lock could be released in the meantime
                                                && ((crossingLock = AWAITED_LOCKS.get(owner)) != null);) {
                                        owner = crossingLock.getOwner();
                                        if (owner == Thread.currentThread()) {
```

bjhargrave:
There is also another bug I fixed where we did not clear the awaiting lock when throwing the deadlock exception. See https://github.com/bjhargrave/equinox/commit/66774452a4ed61eae170594f50fc870b45629d78 for fix. My latest branch is at https://github.com/bjhargrave/equinox/commits/pull/68.

HannesWell:
> I think we still also have to agree that this PR is worth doing. I am still on the fence.

I have the same position here like @laeubi, we definitely have to address the deadlock even if it comes with a performance penalty. When using the current m2e-2.0 snapshot I encountered that dead-lock multiple times already on different computers and I can just say that it is very nasty. After I selected the workspace the splash-screen goes down and just nothing happens. On Linux not even an Eclipse icon was shown in the tool-bar and the only could kill the process from the task manager. On Windows that was similar. I mainly encountered that after running updates. I don't want to imagine what happens if only 1% of Eclipse users encounter that problem and how many complaints we get then. I think that won't be beneficial for Eclipse.
And as far as I can tell m2e does nothing wrong here and it is clearly an issue of Equinox.
I replicated the M2E scenario with a test in https://github.com/eclipse-equinox/equinox/pull/31 and that does nothing forbidden (just enforces some unlucky timing). I'm about to integrate that case into this PR.

Furthermore when DS is used more this scenario can happen more and more often in other places too.

> > With the current change this behavior is changed and given that we potentially have deeply nested locks I'm not sure if this is a good idea.
> 
> It is not entirely clear to me what is correct here. Normally in these situations, I restore the interrupt state after the locked operation completes. This way the locked operation is not "broken" by the interrupt state. But with intrinsic locks this just flows in so I can see your argument also has merit.

To me it seems to be a design decision how to handle interruptions, but since that decision was already made in the past, I suggest we simply stick to it.

> > The way it is currently implemented let the current thread spin in an endless loop
> 
> It was a first attempt :-) Your change though makes the detection much more expensive. Sets and queues are created, manipulated, and iterated over. I think we need a cheaper test (or accept we cannot detect all possible deadlocks).

Sure, but I think we are well in improving them step by step. :)
It is correct that the check is more expensive than before, but since it is only carried out after the lock-timeout has elapsed after 100ms I think it does not matter because we are waiting here anyway and one cannot wait more performant. The only case were this would be relevant is if the lock was released immediately after the timeout has elapsed and will be acquired in the next attempt. Only then a quick check is beneficial. For that one has to be unlucky.

However, the only performance penalty of this change only comes from the replacement of the intrinsic `synchronized` by a `ReentrantLock`, but as said above I think we have to accept that. But even if directly compared the Lock is slower the question is if the overall impact is noticeable.

> > Your change though makes the detection much more expensive. Sets and queues are created, manipulated, and iterated over. I think we need a cheaper test (or accept we cannot detect all possible deadlocks).
> 
> I think a simple change to my loop will avoid the infinite looping issue:

That being said your suggestion for the improved cycle-detection looks good to me since it should work reliably and should not loop forever. Since we only want to know if the current thread is part of the cycle and are not interested in the cycle itself this should be sufficient. And we don't have to worry about a concurrent change of the map size, because if we are really in a dead-lock that number will never go below the number elements in the cycle.

> 
> As we incrementally improve, we should probably just add commits to the tip and not squash arbitrary commits. Once we get close to done and also agree that this PR should even be merged, then we should squash all the commits into a single commit before merging.

Alright, that sounds good. 👍🏽 I first assumed we maybe want to preserve the commits and therefore reordered them. Then sorry for the extra work for you (and me) and lets stick with the current approach.




> There is also another bug I fixed where we did not clear the awaiting lock when throwing the deadlock exception. See [bjhargrave@6677445](https://github.com/bjhargrave/equinox/commit/66774452a4ed61eae170594f50fc870b45629d78) for fix. My latest branch is at https://github.com/bjhargrave/equinox/commits/pull/68.

Good catch! Your branch looks good. I'll force push it to this PR and then apply the test-addition on top of that.

HannesWell:
The good news is that the test-cases to reproduce https://github.com/eclipse-equinox/equinox.framework/issues/55 that I just integrated here from #31 now passes.
The bad news is that the it looks like that for `org.eclipse.equinox.ds.tests` the tests are not executed in the CI. But that is another issue.

However at least (as suspected) this change resolves the deadlock properly. If we now find a strategy to even recover from it we would have an ideal solution.
One naive approach would be to check if we are at the first call of `ServiceRegistrationImpl.getService()` of the current thread and simply try again after some time to lock the use and run the same code.

But we have to ensure that the rest of Equinox and Felix-SCR is ready for that.
Furthermore I wonder if `releaseService()` and `ungetService()` can realistically deadlock too?

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -267,21 +267,21 @@ ServiceUseLock lock() {
...
+				AWAITED_LOCKS.remove(awaitingThread);
 			}
 			if (interrupted) {
 				Thread.currentThread().interrupt();
 			}
 		}
 	}
 
-	private static boolean isDeadLocked(ServiceUseLock lock) {
+	private static boolean isDeadLocked(ServiceUseLock lock, Thread currentThread) {
```
bjhargrave, line 279: I would make this `void checkDeadLocked` and have it throw any exception. Then all the deadlock code is together.

bjhargrave:
I added a commit to move the awaiting map into an instance field of ServiceRegistry since we do not want to use statics since multiple frameworks may co-exist in the same JVM.

I also had to rebase on master due to conflicts.

HannesWell:
Sorry for the late response.

> I added a commit to move the awaiting map into an instance field of ServiceRegistry since we do not want to use statics since multiple frameworks may co-exist in the same JVM.

I would have said that multiple frameworks in the same JVM are not a problem, since the static Set does not carry any 'framework-information'? But I have no objection against your change.

> I also had to rebase on master due to conflicts.
>
Thanks.

Are you in general fine with this change? I am so far.
I want to stream-line/complete the test code soon (hopefully until tomorrow). Then from my side the only pending point is (maybe) fine-tuning.

bjhargrave:
> Are you in general fine with this change?

I am but I want to get some feedback from @tjwatson testing for performance/correctness before we merge. I asked Tom again today about doing this.

HannesWell:
> > Are you in general fine with this change?
> 
> I am but I want to get some feedback from @tjwatson testing for performance/correctness before we merge. I asked Tom again today about doing this.

Sounds good. 👍🏽
I just wanted to have general issues discussed better sooner than later. If we agree to add this, I think it would be good add it for M2 (which is due Friday in a week, 29th of July).
I'm on vacation from Monday on for two weeks and can therefore not assist more then until M2.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceUse.java
```diff
@@ -171,8 +210,147 @@ void decrementUse() {
...
+	 *                          detected.
+	 */
+	ServiceUseLock lock() {
+		Thread awaitingThread = null;
+		boolean interrupted = false;
+		try {
+			final ServiceUseLock useLock = getLock(); // local var to avoid multiple getfields
+			while (true) {
+				try {
+					if (useLock.tryLock(100_000_000L, TimeUnit.NANOSECONDS)) { // 100ms (but prevent conversion)
```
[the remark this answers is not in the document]
bjhargrave: I think the current code is correct. It takes the optimistic approach that the thread is not interrupted (which is the most common case). If the thread is interrupted, then the tryLock will throw an InterruptedException which will be caught and we record the interrupt status for setting upon return. The throwing of the exception clears the interrupt status and we loop back up to call tryLock again.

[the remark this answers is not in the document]
HannesWell: Yes that was the intention.
Furthermore it is the only reliable approach because if we are unlucky the thread gets interrupted just after we called `Thread.interrupted()` and before `tryLock()` is called. So we have to be prepared anyway.

HannesWell:
The tests are now complete and can be reviewed as well.

tjwatson:
I ran the proposed changes through an Open Liberty build and it passed all but one of our 86246 tests (the one failure appears to be an infrastructure issue).  This helps confirm the fix does not have any functional issues, although we should get a run against the OSGi core TCK also.

I still need to get startup performance measurements done on a configuration that enables a very large set of service components during startup.  I hope to have something in the next day or so.

tjwatson (approved):
The performance run across a wide range of configurations in Open Liberty showed no measurably significant slow down.  In some cases we saw a small performance improvement.  

I suspect the improvements observed are due to other improvements that we have yet to consume in Open Liberty though.  For example https://github.com/eclipse-equinox/equinox.framework/pull/54 and https://github.com/eclipse-equinox/equinox.framework/pull/46.  It is still possible that this change has some measurable cost but it could be getting offset by the other improvements.  Regardless, it seems that the cost is not significant enough to be concerned about at this point.

Another observation is that the measurable CPU time did see a slight increase in some configurations, but that increased CPU utilization did not seem to impact the overall startup time of that configuration when 4 cores are assigned to the test run, implying that a bit more work is happening in the background.  When restricting the run to a single core this did convert to a slightly slower startup time for these configurations.  Regardless, that still seems to be a small cost to the overall startup time performance.

tjwatson:
I should also state that in Open Liberty we use `equinox.start.level.thread.count` to enable parallel bundle activation by setting it to something > 1.  This causes a flood of service registry activity from multiple threads when we are starting up.  This in turn should exercise the locking mechanism here to a great extent.

vogella:
@tjwatson with this fix, can you activate equinox.start.level.thread.count by default (setting it to > 1)?

laeubi:
@vogella AFAIK this is non standard behavior and should best be activated on the EPP if desired.

tjwatson:
> @tjwatson with this fix, can you activate equinox.start.level.thread.count by default (setting it to > 1)?

Recall my comments from https://bugs.eclipse.org/bugs/show_bug.cgi?id=540507#c17

> You should know that only that first number from "Starting application" can really be influenced by the parallel bundle activation from the start-level launch.  By this time all the bundles have been activated by the framework.  The rest of the activation that occurs in a typical Eclipse launch happens in response to a class load with lazy activation.

> This is one reason I think you will find very little gain in parallel bundle activation at this level, because most of Eclipse bundles are lazy activated with a trigger class load.  That activation then happens on the thread that is loading the class.

> Furthermore, I really only think this parallel bundle activation from the framework launch helps in cases where many bundles are installed that are eagerly activated.

I see very little advantage to using this option by default for any Eclipse RCP usage as long as most all bundles are lazy activated.

> @vogella AFAIK this is non standard behavior and should best be activated on the EPP if desired.

There is nothing non-standard here.  The OSGi specification does not mandate that bundles within the same start-level be started by a single thread or in any particular order within the start-level.  The implementation is free to start in parallel or not.

HannesWell:
> @tjwatson with this fix, can you activate equinox.start.level.thread.count by default (setting it to > 1)?

Can't you simply test with the latest I-build if a value greater one for that flag speeds up the Start-up?

tjwatson:
> Can't you simply test with the latest I-build if a value greater one for that flag speeds up the Start-up?

Yes, that is what I did when I first put this function into Equinox.  I saw no improvement at all with it enabled.  After some investigation it was because most everything is lazy-activated so all the real work only happens when classes are loaded.  The majority of class loading does not happen until all the bundles have been started and control has been returned to the thread that runs the SWT display thread.  Because DS is lazy also, even the registration of the DS components do not load any classes until the service is actually used (unless the component is immediate).

Let me clarify the term "started".  That is really just Bundle.start.  But for lazy activated bundles that does not mean the `BundleActivator.start` gets called immediately.  Instead that will be delayed until the first class is defined using the class loader of the bundle.  The changes to do parallel starts for start-level really just make sure Bundle.start is called.  But for the vast majority of our bundles that results in very little work to be done until a class load is triggered.

I personally do not plan to retest if starting in parallel helps for Eclipse RCP/IDE bring up again because I don't think anything has changed that would show it helping.

PR #70, opened 2022-07-04
Reuse the Manifest data in the BND file

laeubi:
This works except the additional Require-Capability see https://github.com/bndtools/bnd/issues/5297

@bjhargrave @rotty3000 any idea how to help BND here to merge both entries?

### DIFF
--- a/bundles/org.eclipse.osgi/bnd.bnd
+++ b/bundles/org.eclipse.osgi/bnd.bnd
@@ -1,61 +1,9 @@
 -jpms-module-info: ${project.artifactId};access='OPEN';version=${versionmask;===;${version_cleanup;${project.version}}}
 -jpms-module-info-options: jdk.unsupported;static="true",java.management;ignore="true"
 
-Export-Package: \
- org.eclipse.equinox.log, \
- org.eclipse.osgi.container, \
- org.eclipse.osgi.container.builders, \
- org.eclipse.osgi.container.namespaces, \
- org.eclipse.osgi.framework.console, \
- org.eclipse.osgi.framework.eventmgr, \
- org.eclipse.osgi.framework.internal.reliablefile, \
- org.eclipse.osgi.framework.log, \
- org.eclipse.osgi.framework.util, \
- org.eclipse.osgi.launch, \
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.osgi/bnd.bnd
```diff
@@ -1,60 +1,7 @@
...
- org.osgi.service.permissionadmin, \
- org.osgi.service.resolver, \
- org.osgi.service.startlevel, \
- org.osgi.service.url, \
- org.osgi.util.tracker
-
-Provide-Capability: \
- osgi.serviceloader;osgi.serviceloader="org.osgi.framework.connect.ConnectFrameworkFactory";register:="org.eclipse.osgi.launch.EquinoxFactory", \
- osgi.serviceloader;osgi.serviceloader="org.osgi.framework.launch.FrameworkFactory";register:="org.eclipse.osgi.launch.EquinoxFactory"
+-include: META-INF/MANIFEST.MF
```
bjhargrave, line 4: If you do 

`-include ~META-INF/MANIFEST.MF`

it will not override any values set in the including bnd file. So the bnd file would need to specify the complete desired Require-Capability header.

laeubi: Great this works :+1:  would be cool to have a modifier to actually merge duplicate directives but for this case preferring BND file instructions already works!

> So the bnd file would need to specify the complete desired Require-Capability header.

In this case its even simpler as the BND generated manifest is actually not used and this is just a vehicle to generate the JPMS infos.

laeubi:
This now works as expsected, the module-info.class now contains only some additional things that are most probably missing before:

- [old_module_info.txt](https://github.com/eclipse-equinox/equinox/files/9041551/old_module_info.txt)
- [new_module_info.txt](https://github.com/eclipse-equinox/equinox/files/9041550/new_module_info.txt)

laeubi:
@tjwatson as you are fine with the additional exports (and BND do not plan to change anything) are you fine with this change? I think this will prevent manual work and associated errors in the future.

tjwatson (approved):

PR #71, opened 2022-07-09
New source for sshd

mickaelistria:


laeubi:
@mickaelistria as mentioned in the other PR it might be more valuable to migrate to a version range here instead of forcing qualifier updates every time.

laeubi:
See for example https://github.com/eclipse-equinox/equinox/pull/67

PR #72, opened 2022-07-10
Apply jdt clean ups

HannesWell:
Apply JDT clean ups

- "Use diamond operator"
- "Use try-with-resource"
- "Use try with multi-catch"

to Equinox-sources and ignore embedded felix and osgi sources.

The last clean up required manual removal of now unused Exception type imports but that problem has been reported to JDT.

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/FrameworkExtensionInstaller.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/FrameworkExtensionInstaller.java
@@ -80,8 +80,6 @@ public class FrameworkExtensionInstaller {
 				result.setAccessible(true);
 			}
 			return result;
-		} catch (SecurityException e) {
-			// if we do not have the permissions then we will not find the method
 		} catch (NoSuchMethodException | RuntimeException e) {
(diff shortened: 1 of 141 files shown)


HannesWell:
The test failures is only for the GH workflow. Jenkins and local execution of that test pass.
Merging.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/FrameworkExtensionInstaller.java
```diff
@@ -80,8 +80,6 @@ private static Method findMethod(Class<?> clazz, String name, Class<?>[] args, C
 				result.setAccessible(true);
 			}
 			return result;
-		} catch (SecurityException e) {
-			// if we do not have the permissions then we will not find the method
```
tjwatson, line 84: Why did you remove this?  I believe this will result in blowing up `<clinit>` if the framework does not have permission to reflect.

tjwatson: Nevermind.  SecurityException is a RuntimeException!

PR #74, opened 2022-07-13
Move more parts to pomless, use structured layout

laeubi:


bundles/org.eclipse.equinox.slf4j.stub/build.properties
```diff
@@ -15,3 +15,4 @@ bin.includes = META-INF/,\
                .,\
                about.html
 src.includes = about.html
+pom.model.property.skipAPIAnalysis=true
```
HannesWell, line 18: This should not be necessary anymore.
The api-check profile from the parent-platform now 'computes' `skipAPIAnalysis` based on the presence of the API-builder.

HannesWell (changes requested):
I think Tycho and Equinox are not yet ready for this change.
Tycho's pomless-aggregator generation for the structured layout that does not work for I builds since Equinox then is not the top-level project.
However I already thought about enhancing Tycho to allow arbitrary aggregator directories specified in the parent pom (not only `bundles`, `features`, `tests`, ...) without the need the also specify them in the corresponding Tycho-property.

Furthermore not all test projects are currently build and executed and if they are they fail.
I already started to make them work, like in https://github.com/eclipse-equinox/equinox/pull/29, but there are one or two more to do.

Additionally this change reveals some suppressed warnings, that we should first fix before we remove their suppression. Otherwise some people will let you know that the I-build now has warnings. I already started some work in this direction.

So I suggest to gradually use Tycho-pomless more (I have also prepared some changes for that as well but didn't want to start another PR before my vacation) and after all test projects are runnable and Tycho can also handle arbitrarily nested aggregator directories we change to the structured layout.

laeubi:
> Tycho's pomless-aggregator generation for the structured layout that does not work for I builds since Equinox then is not the top-level project.

It don't need to be on the top level.

> However I already thought about enhancing Tycho to allow arbitrary aggregator directories specified in the parent pom

That would be too broad and one would not be able to have folders that are ignored by Tycho.

> Additionally this change reveals some suppressed warnings, that we should first fix before we remove their suppression.

This is intentional, as discussed we want to use an additional Jenkins plugin once it is enabled see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1422

laeubi:
Replaced by https://github.com/eclipse-equinox/equinox/pull/75 @HannesWell  I think we still can move things, if you have concerns about the automatic module pom, we can just keep the current explicitly named modules, but I'll leave this to you, just let me know when you made progress here as I wanted to try enabling snapshot deployments for equinox once we do not have too much poms to configure, making it pomless simplifies the task a lot.

HannesWell:
> > Tycho's pomless-aggregator generation for the structured layout that does not work for I builds since Equinox then is not the top-level project.
> 
> It don't need to be on the top level.

You are right. They only need to be top-level if there are custom aggregator directories specified in the maven.config, don't they?

> 
> > Additionally this change reveals some suppressed warnings, that we should first fix before we remove their suppression.
> 
> This is intentional, as discussed we want to use an additional Jenkins plugin once it is enabled see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1422

I agree that the discussed one is the better approach. But shouldn't we wait until that plug-in is enabled?



> Replaced by #75 @HannesWell I think we still can move things, if you have concerns about the automatic module pom, we can just keep the current explicitly named modules, but I'll leave this to you, just let me know when you made progress here as I wanted to try enabling snapshot deployments for equinox once we do not have too much poms to configure, making it pomless simplifies the task a lot.

That's actually right. I have to admit I somehow didn't realize that it only moved the tests into a dedicated folder.
But yes that could be done already if we keep the explicit aggregator-pom for the tests folder. We could then just include the `bundles` module in the root.
Just let me complete #75, which already removes many poms and then have a dedicated PR to move the tests into the `tests` folder. Does this sound like a good plan for you? Do you want to revive this PR for that?

laeubi:
> You are right. They only need to be top-level if there are custom aggregator directories specified in the maven.config, don't they?

I don't understand ... it does not matter where the directory is, you just

1. use one of the "magic" names
2. reference that directory somewhere (e.g. module, -f, ...)

> I agree that the discussed one is the better approach. But shouldn't we wait until that plug-in is enabled?

It has proven that "waiting" for something that is not a problem because of hacks, disabled things, whatever mostly leads to "wait forever" ...

> Do you want to revive this PR for that?

Not really, recreate it from scratch is mot probably much easier than rebase this one and resolve the conflicts :-)

Issue #9, opened 2022-05-02
Use tycho pomless?


laeubi:
I noticed most bundles use trivial pom files so it might be a goof idea to use the pomless feature of tycho to e.g. prevent the need to adjust the version in two different files.

HannesWell:
All pom files that can be removed (at the moment) have been removed, so I think we can consider this as completed.

PR #75, opened 2022-07-14
Simplify poms and delete (now) trivial poms to use Tycho-pomless more

HannesWell:
Remove '<skipAPIAnalysis>true</skipAPIAnalysis>' properties that are now obsolete since the platform-build detects automatically if the API-check should run or not.

For o.e.osgi.util the additional compiler-args are not necessary anymore because this plug-in does not contain any sources(in that process it was forgotten to remove them). Their removal makes the pom trivial so it can be removed to use Tycho-pomless instead.

Part of https://github.com/eclipse-equinox/equinox/issues/9

@laeubi as mentioned in https://github.com/eclipse-equinox/equinox/pull/74 this is my current work to use more pomless in equinox. The current state preservers the suppressed warnings.
If we want we can remove the suppression now or gradually later with the corresponding fixes. At least for equinox.app I have started to prepare some fixes.


HannesWell:
Reminder for myself: At https://github.com/eclipse/tycho/wiki/Tycho-Pomless `tycho.pomless.parent` should be described too.

laeubi (approved):
Go for pomless :-)

HannesWell:
The build looks good so far. The test failures should vanish as soon as https://github.com/eclipse-equinox/equinox/pull/43 and https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/389 (i.e. https://github.com/eclipse-platform/eclipse.platform/pull/46) are merged and available. The former also depends on the latter.

HannesWell:
Build succeeded, all tests pass (at least in Jenkins and GH is only one that does regularly). The number of build modules and executed tests is the same.
-> Ready for submission.

PR #76, opened 2022-07-15
Use slf4j from Maven Central

mickaelistria:


laeubi (changes requested):
Please use
```
   <requires>
    <import plugin="slf4j.api"/>
    <import plugin="slf4j.api.source"/>
   </requires>
```
with a suitable version range that matches the version ranges used in equinox when referencing SLF4J

mickaelistria:
I prefer not mixing different concerns in the same commit. And this particular case is not working anyway for the moment (see build failure)

laeubi:
> I prefer not mixing different concerns in the same commit. And this particular case is not working anyway for the moment (see build failure)

Yes it is not working because a version range would be required to not select the slf4j 2.0 API :-)

mickaelistria:
That's not how I interpret `Missing requirement: slf4j.api 1.7.32 requires 'java.package; org.slf4j.impl 1.6.0' but it could not be found`. The Orbit bundle doesn't add such Import-Package requirement

laeubi:
> requires 'java.package; org.slf4j.impl 1.6.0'



> That's not how I interpret `Missing requirement: slf4j.api 1.7.32 requires 'java.package; org.slf4j.impl 1.6.0' but it could not be found`. The Orbit bundle doesn't add such Import-Package requirement

You could include
```
<dependency>
    <groupId>org.osgi</groupId>
    <artifactId>slf4j.osgi</artifactId>
    <version>1.7.1</version>
</dependency>

```

mickaelistria:
> Missing requirement: slf4j.api 2.0.0.alpha6 requires 'osgi.extender; (osgi.extender=osgi.serviceloader.processor)' but it could not be found

That one I don't really mind. I would rater stick to 1.7.x at the moment to remain close to the current state.

> You could include [...]

Maybe yes, but I first focus on libs that can be directly replaced. I'll leave this PR as it for the moment and will come back to it later.

laeubi:
> I would rater stick to 1.7.x at the moment to remain close to the current state.

Yes but with your current state you pull in the highest version (=2.0.0.alpha6).

akurtakov:
Closing as it has been removed in https://github.com/eclipse-equinox/equinox/commit/de39e61efda1564f07a5b9b0933a9002b8348876

PR #79, opened 2022-07-17
Align project-name of o.e.equinox.launcher.cocoa.macosx

HannesWell:
The name of the project does not match the bundle symbolic name and prevents importing the actual `o.e.equinox.launcher.cocoa.macosx.aarch64` fragment.

In general I wonder what the difference between `o.e.equinox.launcher.cocoa.macosx` and `o.e.equinox.launcher.cocoa.macosx.aarch64` is? The only difference seems to be the name.

Furthermore I noticed that the `o.e.equinox.launcher.gtk.linux.loongarch64` is not build and not published. So can it be removed?


rgrunber:
@HannesWell I was just noticing these 2 fragments were both being included when looking into https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/2784 . At first I though maybe one fragment somehow depend on the other. Then I noticed they have the exact same platform filter `Eclipse-PlatformFilter: (& (osgi.ws=cocoa) (osgi.os=macosx) (osgi.arch=a
 arch64) )` . In the I-builds, even the shared `eclipse_11801.so` is identical. Why do both exist ? CC: @akurtakov  ?

akurtakov:
Mac OS has always been total mystery for me. @sravanlakkimsetti should be able to give some info.

merks:
It looks to me like they have different filters:

> https://github.com/eclipse-equinox/equinox/blob/cfed903b766475c1710d7ae208853f0ba5f1d9cd/bundles/org.eclipse.equinox.launcher.cocoa.macosx.aarch64/META-INF/MANIFEST.MF#L8
> https://github.com/eclipse-equinox/equinox/blob/cfed903b766475c1710d7ae208853f0ba5f1d9cd/bundles/org.eclipse.equinox.launcher.cocoa.macosx.x86_64/META-INF/MANIFEST.MF#L8

I.e., one for each of the supported macosx architectures.  I expect  that when org.eclipse.equinox.launcher.cocoa.macosx.aarch64 was created, that org.eclipse.equinox.launcher.cocoa.macosx already existed and was copied to org.eclipse.equinox.launcher.cocoa.macosx.aarch64 and the filter was changed.  But the original org.eclipse.equinox.launcher.cocoa.macosx was not renamed until @HannesWell did that.

I expect they should generate libraries for the different architectures...

HannesWell:
I assume the same as Ed.
We could probably gain more confidence by checking the git history (I cannot do that at because I'm on vacation).

If that's really the case we could consider to deprecate the fragment without architecture for removal.

See also: https://github.com/eclipse-pde/eclipse.pde/issues/406

sravanlakkimsetti:
@HannesWell @merks 
When we developed aarch64 support, We had couple of approaches we could take have individual fragments for each of the native components or have universal binaries(these includes binaries for both architectures in the form of slices and mac will determine which one to use) 

We decided to go with individual way. We created `org.eclipse.equinox.launcher.cocoa.macosx.aarch64`. But when we did a full build including by including arm64 bundles we tycho builder did not recognize `org.eclipse.equinox.launcher.cocoa.macosx.aarch64` we had a lot of errors as `org.eclipse.equinox.launcher.cocoa.macosx` not found. 

The root of this error is from p2 or pde as I recollect not tycho. We had universal binaries when eclipse supported x86_64 and power_pc architectures in the past. There are some specific changes done to support universal libraries at that time. Those are causing problems now.

To quickly resolve the issue I created `org.eclipse.equinox.launcher.cocoa.macosx` as a symbolic link to `org.eclipse.equinox.launcher.cocoa.macosx.aarch64` just to satisfy the tycho builder. There is purpose for ``org.eclipse.equinox.launcher.cocoa.macosx` to exist other than satisfying the p2/pde builder. 

If possible please get rid of `org.eclipse.equinox.launcher.cocoa.macosx`. Please make sure the aggregator build succeeds by removing this fragment.

Issue #81, opened 2022-07-17
10 failing osgi.tests since I20220715-1800


iloveeclipse:
https://download.eclipse.org/eclipse/downloads/drops4/I20220715-1800/testresults/html/org.eclipse.osgi.tests_ep425I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

Probably side effect of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/389 changes.
All test hit an assert with this message: `Failure | 0.0.1.org.junit.platform.commons`.

HannesWell:
This will likely fixed as soon as the pending https://github.com/eclipse-platform/eclipse.platform/pull/46 is available from I-builds.
I think it depends on the other two pending PRs in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/389.

mickaelistria:
This seems all good now: https://download.eclipse.org/eclipse/downloads/drops4/I20220718-1800/testresults/html/org.eclipse.osgi.tests_ep425I-unit-cen64-gtk3-java11_linux.gtk.x86_64_11.html

PR #82, opened 2022-07-18
Leverage enhanced ConfigurationSessionTestSuite

HannesWell:
Use the enhancements of https://github.com/eclipse-platform/eclipse.platform/pull/54

Use addMinimalBundleSet(), addThisBundle() and addBundle(Class) and leverage the fact that o.e.core.tests.harness now pulls all requirements into the test-runtime. Use the same pattern for the additionally required osgi bundles, which happens automatically if addBundle(Class) is used.

In combination with https://github.com/eclipse-platform/eclipse.platform/pull/54 this also fixes https://github.com/eclipse-equinox/equinox/issues/81.

It will only succeed after the next I-build once https://github.com/eclipse-platform/eclipse.platform/pull/54 is available.

### DIFF
--- a/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
@@ -6,11 +6,16 @@ Bundle-Version: 3.16.800.qualifier
 Bundle-Vendor: Eclipse.org
 Require-Bundle: 
  org.eclipse.core.runtime,
- org.eclipse.core.tests.harness,
+ org.eclipse.core.tests.harness;bundle-version="3.13.0",
  org.eclipse.test.performance,
  org.junit
 Bundle-ActivationPolicy: lazy
-Import-Package: org.osgi.service.event; resolution:="optional"
+Import-Package: org.osgi.service.event,
+ org.osgi.util.function,
+ org.osgi.util.measurement,
+ org.osgi.util.position,
+ org.osgi.util.promise,
+ org.osgi.util.xml
 Export-Package: org.eclipse.osgi.tests.bundles,
  org.eclipse.osgi.tests.appadmin
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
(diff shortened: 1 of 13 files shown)


HannesWell:
It looks like this build does not pick up the latest changes of tonight's I-build.
@laeubi can this be a caching issue?

laeubi:
> It looks like this build does not pick up the latest changes of tonight's I-build. @laeubi can this be a caching issue?

Are there any? I don't see any imports changed ...

HannesWell:
> > It looks like this build does not pick up the latest changes of tonight's I-build. @laeubi can this be a caching issue?
> 
> Are there any? I don't see any imports changed ...

I added some methods to `ConfigurationSessionTestSuite` in https://github.com/eclipse-platform/eclipse.platform/pull/54, which is already imported. So now import changes but existing imports change.
But the build fails saying those methods are undefined:
```
[ERROR] The method addMinimalBundleSet() is undefined for the type ConfigurationSessionTestSuite
[ERROR] /home/jenkins/agent/workspace/equinox_PR-82/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/services/datalocation/LocationAreaSessionTest.java:[96] 
[ERROR] 	sessionLock.addThisBundle();
[ERROR] 	            ^^^^^^^^^^^^^
```

laeubi:
> which is already imported

Adding methods require at least a minor version increase and you should increase the minimum version of that import as well, otherwise OSGi (or P2/Tycho) is free to choose any older version.

HannesWell:
> > which is already imported
> 
> Adding methods require at least a minor version increase and you should increase the minimum version of that import as well, otherwise OSGi (or P2/Tycho) is free to choose any older version.

Aww. Right. Stupid mistake. For once, you don't have API tools watching your back....
Thanks. Created https://github.com/eclipse-platform/eclipse.platform/pull/57

HannesWell:
Great. It finally worked. The test-failure in the GH build is unrelated.
-> Ready for submission.

PR #87, opened 2022-07-24
Recover from Service-Factory deadlock

HannesWell:
Follow up of https://github.com/eclipse-equinox/equinox/pull/68, with a first attempt to implement a recovery strategy to resolve an encountered deadlock during the use of service-factories.

In case of an encountered Deadlock between Service-Factories, this approach simply checks if the current service-factory invocation is the first one of this thread and then makes a new attempt to call the factory.
In order to make a new deadlock more unlikely the thread sleeps for a random period that becomes even longer with further attempts, hoping that the other threads involved the deadlock sleep for another time that is different enough so that the race-condition does not occur anymore.

This PR is a first draft with a relatively naive approach, but at least it worked in the test-cases added for #68:
- Assumes it is possible without restrictions to 
- Currently requires Java-9's `StackWalker`, but this can be reworked to use means available in Java-8.

Alternatively creates of Service-factories could be adjusted to handle encountered deadlocks, but this would likely require changes outside of Equinox and maybe some spec-work in OSGi. 

I'm on vacation for the next two weeks, but wanted to share this current state/the basic idea already.

### DIFF
--- a/bundles/org.eclipse.equinox.ds.tests/.gitignore
+++ b/bundles/org.eclipse.equinox.ds.tests/.gitignore
@@ -1,1 +1,2 @@
 /scr_test
+/OSGI-INF/org.eclipse.*.xml
--- a/bundles/org.eclipse.equinox.ds.tests/OSGI-INF/org.eclipse.equinox.ds.tests.LazyServiceComponentActivationDeadLockTest$Component1.xml
+++ b/bundles/org.eclipse.equinox.ds.tests/OSGI-INF/org.eclipse.equinox.ds.tests.LazyServiceComponentActivationDeadLockTest$Component1.xml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.3.0" name="org.eclipse.equinox.ds.tests.LazyServiceComponentActivationDeadLockTest$Component1">
-   <service>
-      <provide interface="org.eclipse.equinox.ds.tests.LazyServiceComponentActivationDeadLockTest$Component1"/>
-   </service>
-   <reference cardinality="1..1" field="a" interface="org.eclipse.equinox.ds.tests.LazyServiceComponentActivationDeadLockTest$Component2" name="a"/>
-   <reference cardinality="1..1" field="b" interface="org.eclipse.equinox.ds.tests.LazyServiceComponentActivationDeadLockTest$Component4" name="b"/>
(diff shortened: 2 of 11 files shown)


bjhargrave:
I am unclear how waiting will solve the deadlock. The previous fix can detect the deadlock and then throw an exception to alert that there was a deadlock (versus hanging). But just catching the deadlock exception and then blocking the thread for some random time before retrying does not seem to be able to "solve" the deadlock. If A->B->C->A is the deadlock, then no amount of time waiting for A will fix anything.

bjhargrave:
> IIRC the idea is that If Thread 1 is starting at C with: C->A->B->C and Thread 2 is starting at A with A->B->C->A and we detect a deadlock then if we are lucky enough to where Thread 1 is at C or Thread 2 is at A (their starting point) then we force them to wait without the lock on their starting point to allow the other thread to continue. But it seems like a narrow gain that still can allow deadlock.

This can't work since neither C nor A can be available to another thread until it is completely constructed (that is, the service factory returns an object) and neither can completely construct because of the cycle.

HannesWell:
> > I am unclear how waiting will solve the deadlock. The previous fix can detect the deadlock and then throw an exception to alert that there was a deadlock (versus hanging). But just catching the deadlock exception and then blocking the thread for some random time before retrying does not seem to be able to "solve" the deadlock. If A->B->C->A is the deadlock, then no amount of time waiting for A will fix anything.
> 
> IIRC the idea is that If Thread 1 is starting at C with: C->A->B->C and Thread 2 is starting at A with A->B->C->A and we detect a deadlock then if we are lucky enough to where Thread 1 is at C or Thread 2 is at A (their starting point) then we force them to wait without the lock on their starting point to allow the other thread to continue. But it seems like a narrow gain that still can allow deadlock.

That is the intention yes. Of course it is not a perfect solution that works for all cases, but it works for cases like the test case, where the cycle contains optional links. Of course a cycle of required links can never be resolved. But IIRC Felix warns/errors about that if one models that. But for optional-links this is a possible way as you can see in the adjusted test-cases `testLateBindingInSameBundleDeadLock`, where all components gets their optional links set eventually.

HannesWell:
Do you have further remarks on this one?
For the case of deadlocks introduced by cyclic optional/multi links I think this can be beneficial.

laeubi:
@HannesWell as we have not seen any deadlock problems in the day passed and it seems to be there is not full agreement how to handle this maybe just postpone this?

HannesWell:
> @HannesWell as we have not seen any deadlock problems in the day passed and it seems to be there is not full agreement how to handle this maybe just postpone this?

Yes, I'll close this for now but will leave the branch intact, just in case this becomes relevant again and somebody remembers this.

PR #91, opened 2022-07-30
Fix NPE by RegistryFactory.getRegistry

laeubi:
`RegistryFactory.getRegistry` can return null, but the static methods to not check for null.


laeubi:
@HannesWell Any idea why the `testCircularityHandling` fails?

HannesWell:
> @HannesWell Any idea why the `testCircularityHandling` fails?

Not really but ist used to fail for some time already even before the detection of deadlocks was merged.
But it only fails in the GH build.
Unfortunately I cannot help further in vacation.

HannesWell:
But the failure is likely not related to this change.

akurtakov:
Should this one be pushed?

akurtakov:
I don't see issue with preventing this NPE.

PR #92, opened 2022-08-01
The org.eclipse.equinox.jsp.jasper bundle should allow a newer version

merks:
of javax.servlet.jsp than 2.3.

The Orbit 2.2.0 version of javax.servlet.jsp has been removed from the
target platform by the following PR and the newer version 2.3.3 version
is now available, so the org.eclipse.equinox.jsp.jasper should
accommodate that new version.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/412

### DIFF
--- a/bundles/org.eclipse.equinox.jsp.jasper/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.jsp.jasper/META-INF/MANIFEST.MF
@@ -4,13 +4,13 @@ Bundle-Name: %bundleName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-SymbolicName: org.eclipse.equinox.jsp.jasper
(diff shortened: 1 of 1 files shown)


merks:
@akurtakov Is this an occasional problem that a build is abort?

akurtakov:
I've retriggered the build. I don't recall anything like this, maybe @laeubi has an idea.

laeubi:
I only see that everything seems "slow" at the moment (e.g. I also see timeouts in GH actions and some read-timeouts in jenkins ci builds) but everytime I complain at EF infra about anything similar it takes too long for them to look at it and the incident gets closed with "please report back if it occurs again" ... as you can see the GH action here succeeds within 22m so it *should* work in a similar timeframe on the jenkins CI...

akurtakov:
Rebuild succeeded and patch merged.

PR #96, opened 2022-08-11
Equinox.setup needs to enable linked resources.

merks:


HannesWell:
Thank you! This has disturbed regularly and I didn't know where it were coming from (but is was not annoying enough to start searching).

PR #104, opened 2022-08-18
Path: avoid String search before replace.

jukzi:
String.replace already has a fast path to return this if pattern is not
found.


jukzi:
unrelated error: java.lang.AssertionError: The StaticCircuit2 component shouldn't have bound objects expected:<0> but was:<1>
	at org.eclipse.equinox.ds.tests.tbc.DSTest.testCircularityHandling(DSTest.java:1425)

mickaelistria (approved):

jukzi:
unrelated error org.eclipse.equinox.ds.tests.tbc.DSTest.testCircularityHandling

mickaelistria:
Thanks

jukzi:
thanks @mickaelistria for taking the time

PR #113, opened 2022-09-06
Path: Make Path implementation immutable

jukzi:
Prefer constructor over initialize() to make fields final.
This allows to safely share the Path instance across threads.


mickaelistria:
That is very welcome, thanks!

PR #114, opened 2022-09-07
Path: lazy compute hash value

jukzi:
Computing hash is slow while the hash value is rarely used since eclipse organizes the workspace in a sorted tree rather then hash map.

Implementation as in java.util.String.

### DIFF
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- *  Copyright (c) 2000, 2018 IBM Corporation and others.
+ *  Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -13,6 +13,8 @@
  *     Patrick Tasse - Add extra constructor to Path class (bug 454959)
  *******************************************************************************/
 package org.eclipse.equinox.common.tests;
+
+import static org.junit.Assert.assertNotEquals;
 
 import java.util.ArrayList;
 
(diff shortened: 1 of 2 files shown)


mickaelistria:
Doesn't it make Path less thread-safe?

merks:
It seems to me that it's just a race condition but no thread can see an uninitialized value.  One potential problem though is that if the hash is actually 0 it will be recomputed on each call.   Maybe that should be prevented...

merks:
Note that String.hashCode is like this:
```
    public int hashCode() {
        int h = hash;
        if (h == 0 && value.length > 0) {
            hash = h = isLatin1() ? StringLatin1.hashCode(value)
                                  : StringUTF16.hashCode(value);
        }
        return h;
    }
```
And of course String is thread safe (and hash is not declared volatile).  I'm not sure if maybe the zero length string is the only String with hashCode 0...

laeubi:
> Computing hash is slow while the hash value is rarely used since eclipse organizes the workspace in a sorted tree rather then hash map.

If it is slow (how slow?) but rarely used, should we really "optimize" that and can not simply compute the hash on the fly for the rare cases where it is used?

merks:
> > Computing hash is slow while the hash value is rarely used since eclipse organizes the workspace in a sorted tree rather then hash map.
> 
> If it is slow (how slow?) but rarely used, should we really "optimize" that and can not simply compute the hash on the fly for the rare cases where it is used?

Let's not assume that how the platform uses it is how everyone uses it...

laeubi:
> Let's not assume that how the platform uses it is how everyone uses it...

I'm all for that, but that's the assumption/rationale for this change, so either we want this change (regardless of ho often hash code is needed) or it should be taken into account that this is premature optimization.

> I'm not sure if maybe the zero length string is the only String with hashCode 0...

At least the contract of hashCode/equals is that if equals returns true the hashcode should be equal and that the hashcode has at best no collisions. From these both contracts you can assume that:

- the only string that always produces hashcode=0 is the empty string
- it is very unlikely that any other string produces hashcode=0 (because that would be a collision that should be avoided anyways) 
- If there is a collusion, the impact compared to the whole set of possible strings is rather irrelevant as you already get worse performance due to the collision itself

merks:
I assume that @jukzi has done profiling to see how much computing the hash code costs and can see the call hierarchies from where it's called to conclude that it's primarily called during initialization.   Yes I **assume** this because I see him doing lots of measurements.  👅

jukzi:
Thread safety: the trick is that if one CPU does not already see the computed value it will just calculate it in that thread. As soon as the cache is flushed other CPUs will eventually see the result too but it does not matter if two CPUs come to the same conclusion. 
0 hash - there is a small chance (1/MAX_INTEGER) that instances will result in a hash of 0 which will always calculated again on demand - who care - in average it's still a win!

measurement: i have often seen this as hotspot, the last occurrence is documented here: https://github.com/eclipse-platform/eclipse.platform/issues/111#issuecomment-1218072249

image

It's only a small improvement compared to the overall too much computation done there. But i like to take a quick win by optimizing a very basic class that might also improve other usecases.

mickaelistria:
I personally like this change and want to reaffirm such changes are very welcome in Platform. If some internal and safe optimization is clearly a win just according to the code or the analysis with no risk of regression, even if it's a neglect-able win in current Platform state, it is really worth chasing it. As @merks mentioned, we never know what others are doing of the model Platform has built, and we also don't know what the future of Platform will be doing with those. Having such model optimized early makes them more sustainable or more likely for Platform to make progress.

On a more general aspect:
Even marginal wins are important wins. I remember a talk from Tracy Miranda at EclipseCon France where she mentioned some other presentation from the technical director of the Sky cycling team back when they were leading all challenge partly thanks to technical innovation. The take away was that on the long run, a marginal win (eg 0.1%) is more than what is likely to separate the winner of a major race from the 2nd. The gap between winner and 2nd on Tour de France this year is 0.06% of the overall race time, and some marginal wins have contributed to the win. Marginal wins are important wins; in mature ecosystem or technology, most of the wins are marging, yet the sum of marginal wins makes the final difference.

On a more focused aspect: 2 tests are now failing ;)

laeubi:
> i have often seen this as hotspot

One should take in min that profiling very fast methods can make them actually appear them slower as they are...

Funny enough I think that much more could be gained by caching instance of path (especially as they are final now), but this will require to compute the hash-code always ;-)

laeubi:
> there is a small chance (1/MAX_INTEGER) that instances will result in a hash of 0 which will always calculated again on demand - who care - in average it's still a win!

Please note that this is only true if has computation is proven to be well-distributed, looking as the current implementation I don't think this actually true!

merks:
> > i have often seen this as hotspot
> 
> One should take in min that profiling very fast methods can make them actually appear them slower as they are...
> 

This depends on things like whether tracing or sampling is used.  I personal do not to trust sampling because it often leads to misleading conclusions that some method is super expensive when it's not; at least that's been my experience.  But using tracing and looking at %, in my experience, leads to pretty much exactly the expected improvements in an actual running application...

laeubi:
> This depends on things like whether tracing or sampling is used. I personal do not to trust sampling because it often leads to misleading conclusions that some method is super expensive when it's not; at least that's been my experience. But using tracing and looking at %, in my experience, leads to pretty much exactly the expected improvements in an actual running application...

I assume you mean `tracing` = `instrumentation`?

If you are using instrumentation you add overhead e.g. the Profiler needs to add code to record method entry/exit and get timestamps for that and the JVM might not be able to properly inline/jit-compile the method so **especially** using instrumentation makes very fast methods seem slow if we are talking about w few percentages in the overall time slice.

jukzi:
To my experience sampling is good to find bad system calls while flight recording is more accurate if the hotspot is inside the JVM (but is totally blind against OS calls). In this case i did both, and both methods agreed it's a hotspot. And it makes perfectly sense since hashing with a multiply is well known to be slow (for example https://lemire.me/blog/2015/10/22/faster-hashing-without-effort/). Modern CPUs are superscalar but the algorithm is not parallelizable.
Caching is not an option there since the Pathes are constructed temporarily by appending a segment.

laeubi:
> Caching is not an option there since the Pathes are constructed temporarily by appending a segment.

But the caching the has-code is useless anyways if path is never cached ... ;-)

jukzi:
> But the caching the has-code is useless anyways if path is never cached ... ;-)

the hash it's never ever computed for those paths. that's the whole point of this change - just never compute the hash!

laeubi:
> > But the caching the has-code is useless anyways if path is never cached ... ;-)
> 
> the hash it's never ever computed for those paths. that's the whole point of this change - just never compute the hash!

And that's why I suggested to maybe simply recompute the hash on each call instead of just *caching the hash* if the assumption was that the hash is never actually needed (but you have already edited the original description).

Still the *construction* of a path do a lot more stuff than calculating the hash, so if *thats* a hotspot then even more can be gained not constructing so much path instances, because even if *some* are temporal, I really hope that most path instances in my workspace won't vanish for a long time...

mickaelistria:
Let's merge this. Further enhancement are welcome in other PRs.

HannesWell:
In case somebody ist interested, I just encountered the following article that discusses the Zero HashCode value of String and mentioned that a boolean flag was added in Java 13 to prevent recomputation in that Case:
https://www.javaspecialists.eu/archive/Issue277-Strings-with-Zero-HashCode.html

jukzi:
interesting

Issue #16, opened 2022-05-05
No clear instructions how to build & test launcher binaries


iloveeclipse:
To review https://github.com/eclipse-equinox/equinox.framework/pull/2, I'm trying to understand how to **build** and **use** the built library in SDK.

To build, I did
```
cd /data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/library/gtk
./build.sh clean
./build.sh
```
That seem to run C build successfully.

So I can just take installed SDK and replace `eclipse` in the root of the SDK with generated `/data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/library/gtk/eclipse` file. Starting it in the shell opens Eclipse workspace dialog, but is that correct procedure? Also the binary seem to be much smaller compared to original one.

So I've tried following:

```
cd /data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature
mvn clean verify -Pbuild-individual-bundles -Drt.equinox.binaries.loc=/data/git/equinox.binaries/ -Dnative=gtk.linux.x86_64
```

I see that during the build
```
Copying /data/git/equinox.binaries/org.eclipse.equinox.executable/bin/gtk/linux/x86_64/eclipse to /data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/bin/gtk/linux/x86_64/eclipse

...

Attempting to rename: /data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/bin/gtk/linux/x86_64/eclipse to /data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/bin/gtk/linux/x86_64/launcher

...

Building zip: /data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/target/org.eclipse.equinox.executable-3.8.1700-SNAPSHOT.jar
```
In that `/data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/target/` directory I see 
`/data/git/rt.equinox.framework/features/org.eclipse.equinox.executable.feature/target/tmp/bin/gtk/linux/x86_64/launcher` executable, which is **binary different** to `eclipse` executable I've created before, but nearly same size of the original executable from SDK.

```
[...]

PR #118, opened 2022-09-18
Fix console launcher not consulting eclipse.ini to read vmargs when there is no console window (#117)

### LINKED ISSUE #117: Windows console launcher (eclipsec.exe) not reading vmargs from eclipse.ini when there is no console window
When the console launcher `eclipsec.exe` on Windows is started without a console window, the `vmargs` specified in `eclipse.ini` are ignored. This can easily be tested e.g. by adding debugging arguments `-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=8000` to the `eclipse.ini` of an Eclipse IDE installation.

Starting the Eclipse IDE running `eclipsec.exe` on the command line works as expected: The JVM waits for a debugger to attach with the message `Listening for transport dt_socket at address: 8000`.

Starting the Eclipse IDE using `Runtime.getRuntime().exec("eclipsec.exe")` (easiest to do with `jshell`) does not work as expected: The JVM does not wait for a debugger to attach and just runs the Eclipse IDE. Apparently, the `vmargs` specified in `eclipse.ini` are ignored. In fact, using `procmon` from SysInternals shows that in this case `eclipsec.exe` only attempts to read `eclipsec.ini` but not `eclipse.ini`.

Whether `eclipse.ini` is consulted in addition to `eclipsec.ini` to read the `vmargs` is determined by the function `isConsoleLauncher` which checks whether there is a console window or not:

https://github.com/eclipse-equinox/equinox/blob/a3a9628da771382ba2821b0b61228eb9c9946b6e/features/org.eclipse.equinox.executable.feature/library/eclipse.c#L1609-L1618

It turns out that there are ways to start the console launcher without a console window. One such case is via `Runtime.getRuntime().exec("eclipsec.exe")` which apparently uses the `CREATE_NO_WINDOW` flag when creating the process (see https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags).

The issue came up when running an Ant script that executes an eclipse application using its console launcher in an `exec` task. Ant uses `Runtime.getRuntime().exec("<command_line>")` internally. The workaround in this case is to explicitly specify the `eclipse.ini` file on the command line: `eclipsec.exe --launcher.ini eclipse.ini`.

jonathan-meier:
Fixes #117. The fix is to properly propagate the launcher mode from `main` to the launcher DLL and removing the incomplete `isConsoleLauncher` function entirely.

### DIFF
--- a/features/org.eclipse.equinox.executable.feature/library/eclipse.c
+++ b/features/org.eclipse.equinox.executable.feature/library/eclipse.c
@@ -392,15 +392,15 @@ static _TCHAR** extractVMArgs(_TCHAR** launcherIniValues);
 #ifdef _WIN32
 static void     createConsole();
 static void		fixDLLSearchPath();
-static int 		isConsoleLauncher();
 #endif
-static int      consoleLauncher = 0;
+static int      isConsoleLauncher = 0;
 
 /* Record the arguments that were used to start the original executable */
-JNIEXPORT void setInitialArgs(int argc, _TCHAR** argv, _TCHAR* lib) {
+JNIEXPORT void setInitialArgs(int argc, _TCHAR** argv, _TCHAR* lib, int consoleLauncher) {
 	initialArgc = argc;
 	initialArgv = argv;
 	eclipseLibrary = lib;
+	isConsoleLauncher = consoleLauncher;
 }
 
 #ifdef MACOSX
@@ -571,9 +571,6 @@ static int _run(int argc, _TCHAR* argv[], _TCHAR* vmArgs[])
    	 */
     initWindowSystem( &argc, argv, !noSplash );
 #elif _WIN32
-    /* this must be before doing any console stuff, platforms other than win32 leave this set to 0 */
-    consoleLauncher = isConsoleLauncher();
-
     /*fix the DLL search path for security */
     fixDLLSearchPath();
 #endif
@@ -1017,7 +1014,7 @@ static _TCHAR** getConfigArgs() {
 	int configArgc = 0;
 	int ret = 0;
 
-	configFile = (iniFile != NULL) ? iniFile : getIniFile(program, consoleLauncher);
+	configFile = (iniFile != NULL) ? iniFile : getIniFile(program, isConsoleLauncher);
 	ret = readConfigFile(configFile, &configArgc, &configArgv);
 	if (ret == 0)
 		return configArgv;
@@ -1603,20 +1600,6 @@ static void createConsole() {
 	}
 }
 
-/* Determine if the launcher was the eclipsec.exe or not based on whether we have an attached console.
- * This will only be correct if called before createConsole.
- */
-static int isConsoleLauncher() {
-	HWND (WINAPI *GetConsoleWindow)();
-	void * handle = loadLibrary(_T_ECLIPSE("Kernel32.dll"));
-	if (handle != NULL) {
-		if ( (GetConsoleWindow = findSymbol(handle, _T_ECLIPSE("GetConsoleWindow"))) != NULL) {
-			return GetConsoleWindow() != NULL;
-		}
-	}
-	return 0;
-}
-
 static void fixDLLSearchPath() {
 #ifdef UNICODE
 	_TCHAR* functionName = _T_ECLIPSE("SetDllDirectoryW");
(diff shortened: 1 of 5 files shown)


akurtakov:
@niraj-modi would you please review this patch?

akurtakov:
@nnemkin Is it possible that you review this patch ?

Torbjorn-Svensson (approved):

Torbjorn-Svensson:
@vogella are you planning to do more work/review on this PR or was your force-push just a rebase?
Do you know anyone that can take a look at this PR and hopefully get it merged?

vogella:
> @vogella are you planning to do more work/review on this PR or was your force-push just a rebase? Do you know anyone that can take a look at this PR and hopefully get it merged?

Just a rebate. Can you check the failing tests?

Torbjorn-Svensson:
> > @vogella are you planning to do more work/review on this PR or was your force-push just a rebase? Do you know anyone that can take a look at this PR and hopefully get it merged?
> 
> Just a rebate. Can you check the failing tests?

I have no idea why the test [testCircularityHandling ](https://github.com/eclipse-equinox/equinox/blob/33ad07188e546f4e5f2757ebd6f2c52648fbcea5/bundles/org.eclipse.equinox.ds.tests/src/org/eclipse/equinox/ds/tests/tbc/DSTest.java#L1375)(org.eclipse.equinox.ds.tests.tbc.DSTest) would fail with this change.
To me, this test has nothing to do with what .ini file to read in the eclipsec.exe file. Are you of a different opinion?
Does this work for other PR's?

iloveeclipse:
> Can you check the failing tests?

I'm pretty sure the native libraries code isn't built & tested anyway on PR's, and there are no tests for that change either.

The binaries are supposed to be built manually via magic jenkins jobs, see https://github.com/eclipse-equinox/equinox/issues/16#issuecomment-1157432730

More important question is: has anyone tried to build the code and test if that change actually works?

Torbjorn-Svensson:
> > Can you check the failing tests?
> 
> I'm pretty sure the native libraries code isn't built & tested anyway on PR's, and there are no tests for that change either.
> 
> The binaries are supposed to be built manually via magic jenkins jobs, see [#16 (comment)](https://github.com/eclipse-equinox/equinox/issues/16#issuecomment-1157432730)
> 
> More important question is: has anyone tried to build the code and test if that change actually works?

No, I've not built it and I don't even have enough permission to do it on the Jenkins job.
All I've done is to review the change and based on that it relies on the define instead of doing a runtime check, makes it less subject to work some time and not some other time.

Torbjorn-Svensson:
Using the github action approach in #247, I've been able to validate the functionality using the following approach:

1. Created a small console application RCP in that simply prints the value of java property and exits.
2. Modified the eclipse.ini to contain the definition of the property
3. Created a small java application that runs the product exported in step 1.

The java application contains the following code:
```java
public class Main {
    public static void main(String[] args) throws IOException, InterruptedException {
        String path = "C:\\tmp\\eclipse";

        Process proc = new ProcessBuilder(path + "\\eclipsec.exe")
                .directory(new File(path))
                .start();
        
        System.out.println("Exit code: " + proc.waitFor());

        try (BufferedReader is = new BufferedReader(new InputStreamReader(proc.getInputStream()))) {
            String line;
            while ((line = is.readLine()) != null) {
                System.out.println(line);
            }
        }
    }
}
```

When running the application defined above with the Eclipse 2023-03 SimRel binaries, the property cannot be extracted:

```
Exit code: 0
Property value is: null
```

When running it with the binaries produces with pr/118, then I instead get the following result:

```
Exit code: 0
Property value is: foo
```

Other than a rebase and removing some trailing whitespace, I would say that this PR is ready for a merge (keep in mind that I'm not an equinox committer).

What's the next step to get this change merged?

vogella:
Merging based on [Torbjorn-Svensson](https://github.com/Torbjorn-Svensson) review.

vogella:
Thanks @jonathan-meier  and @Torbjorn-Svensson for working in this and testing it. We (PMC) decided a while ago that should remove the native support for splash screen from the launcher as the assumption was that this would greatly simplify the launcher code. Maybe something you could help with, in case you also think that simplification of the launcher code might be helpful.

HannesWell:
In the area of the launcher the following would be also an interesting issue.
- https://github.com/eclipse-equinox/equinox/issues/249

umairsair:
See https://github.com/eclipse-equinox/equinox/issues/117#issuecomment-1737305331

> is this fix part of latest eclipse releases?
> 
> IIUC, the binaries are picked from following location and I don't see a new commit after this fix.. am I missing something here?
> 
> https://github.com/eclipse-equinox/equinox.binaries/tree/master/org.eclipse.equinox.executable/bin/win32/win32/x86_64

PR #120, opened 2022-09-21
Test the warnings-checks-for-github-projects

laeubi:


laeubi:
See https://github.com/eclipse-equinox/equinox/pull/134

PR #124, opened 2022-10-01
Simplify and generalize project imports in P2's Oomph setup

HannesWell:
The root .project file of p2 is removed via https://github.com/eclipse-equinox/p2/pull/149

The abandoned project 'org.eclipse.equinox.p2.testserver' has been deleted as part of https://github.com/eclipse-equinox/p2/pull/49.

@merks or do you know another reason why multiple source-locators are required?

### DIFF
--- a/releng/org.eclipse.equinox.releng/Equinox.setup
+++ b/releng/org.eclipse.equinox.releng/Equinox.setup
@@ -269,15 +269,7 @@
         <requirement
             name="*"/>
         <sourceLocator
-            rootFolder="${github.clone.equinox.p2.location/bundles}">
-          <excludedPath>bundles/org.eclipse.equinox.p2.testserver</excludedPath>
-        </sourceLocator>
-        <sourceLocator
-            rootFolder="${github.clone.equinox.p2.location/examples}"/>
-        <sourceLocator
-            rootFolder="${github.clone.equinox.p2.location/features}"/>
-        <sourceLocator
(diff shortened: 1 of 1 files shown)


merks:
It looks like the products folder was deliberately excluded, but I don't see  a good reason to exclude it.

Note that in other Git repos that have a .project in the root, I've added filters to eliminate the duplicates (from folders that contain nested projects) while making the root files available as a project.

https://github.com/eclipse/birt/issues/699

https://github.com/eclipse/birt/blob/6dd4b76d1134ca9dfd67eaeda66a076a4b75b9eb/.project#L17-L27

HannesWell:
> It looks like the products folder was deliberately excluded, but I don't see a good reason to exclude it.

The products folder was just added recently via #114 and the setup existed before that, so there is probably no reason. :) 

> 
> Note that in other Git repos that have a .project in the root, I've added filters to eliminate the duplicates (from folders that contain nested projects) while making the root files available as a project.
> 
> [eclipse/birt#699](https://github.com/eclipse/birt/issues/699)
> 
> https://github.com/eclipse/birt/blob/6dd4b76d1134ca9dfd67eaeda66a076a4b75b9eb/.project#L17-L27

Thanks for that hint. At least for the other Eclipse Platform repos I found a .project file in the repo root (I checked a few but not all) and therefore went the simple way. But if the .project file in the root is needed, for a reason I don't see at the moment, we can use that filter. 👍🏽

PR #128, opened 2022-10-06
Add maven descriptors to equinox bundles

laeubi:
Reverse mapping of maven meta-data is errorprone and cumbersome, with the new tycho features it is now possible to create good meta-data at the first place.

FYI @mickaelistria @akurtakov @merks 

This also demonstrates the use of the new [verify-osgi-pom mojo](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-maven-dependency-consistency-check) to check for bad/missing OSGi dependencies.

This will most probably require a bump for every bundle affected, so I'd like to get some feedback first, also one should note that currently the group id `org.eclipse.equinox` is used as derived from the parent.

For local test one can run:
`mvn -Pbuild-individual-bundles -Dproject.build.sourceEncoding=UTF-8 -DskipTests -Drt.equinox.binaries.loc=../equinox.binaries clean verify -Dtycho.localArtifacts=ignore -e -Pverify-maven`

**Just keep in mind to bump the version of the bundle you like to investigate or it will be overwritten by the baseline replace!**

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,7 +23,7 @@ pipeline {
 			steps {
 				sh """
 				mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-					-Pbuild-individual-bundles -Pbree-libs -Papi-check -Pverify-maven\
+					-Pbuild-individual-bundles -Pbree-libs -Papi-check -Pverify-maven -Pjavadoc\
 					-Dcompare-version-with-baselines.skip=false \
 					-Dproject.build.sourceEncoding=UTF-8 \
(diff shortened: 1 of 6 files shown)


laeubi:
Together wit
- https://github.com/eclipse-tycho/tycho/pull/1475

I was able to generate javadoc artifacts with the default maven-javadoc-plugin as well, I added this as a new commit to this change.

About the consumer pom it seems we have two options:

1. bump all versions now
2. wait until someone changes a bundle, that could need some time 

Also the verify-check shows that equinox has a dependency to the eclipse.jobs that are not part of this repository...

akurtakov:
What is the status of this one?

PR #133, opened 2022-10-06
Enable the generation of javadoc and scan for issue

laeubi:
This will require:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/614

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,90 @@
         </repository>
       </repositories>
     </profile>
+	<profile>
+		<id>javadoc</id>
+		<build>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-javadoc-plugin</artifactId>
+					<configuration>
+						<!-- Special code only for J9 -->
+						<excludePackageNames>org.eclipse.osgi.internal.cds</excludePackageNames>
+						<!-- should be place probably in the clobal config ? -->
+						<doclint>reference,html,syntax</doclint>
+						<links>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -53,6 +53,21 @@
...
+	<profile>
+		<id>javadoc</id>
+		<build>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-javadoc-plugin</artifactId>
+					<configuration>
+						<!-- Special code only for J9 -->
+						<excludePackageNames>org.eclipse.osgi.internal.cds</excludePackageNames>
```
akurtakov, line 65: I think we should exclude *.internal.* as javadoc for internals is not generally required (yet).

laeubi: It is not strictly required, but internal code is still used, so bad javadoc hurts there too, also internal API is widly used across eclipse plugins, so I think its good to have it for now, if it really creates issues, we can later use more explicit excludes.

laeubi:
The test-failure of the Gh actions is unrelated...

pom.xml
akurtakov: OK. Only thing to consider is that if showing hundreds of warnings very few people (if anyone at all) will try to get a clean build as the battle looks lost but if the number is not that great there are people that try to achieve it. Reiteration with stricter rules can happen after that.

laeubi: I have now enabled some of the options from the isv build, that should reduce warnings/errors. Beside that I hope that at least no **new** ones are introduced that would already be a benefit.

PR #137, opened 2022-10-10
improve Dark Theme on Windows eclipse-platform/.github#27

betamaxbandit:
Using CTabFolder and CTabItem, instead of TabFolder/TabItem, improves the look when in Dark Theme; the background of the tab no longer has a non-dark background.

This relates to multiple commits to fix:

[Windows][Dark theme] Multiple preference pages use tab widgets, that are not dark theme aware eclipse-platform/.github#27

### DIFF
--- a/bundles/org.eclipse.equinox.security.ui/src/org/eclipse/equinox/internal/security/ui/storage/StoragePreferencePage.java
+++ b/bundles/org.eclipse.equinox.security.ui/src/org/eclipse/equinox/internal/security/ui/storage/StoragePreferencePage.java
@@ -16,9 +16,11 @@ package org.eclipse.equinox.internal.security.ui.storage;
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.preference.PreferencePage;
 import org.eclipse.swt.SWT;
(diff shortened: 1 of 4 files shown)


betamaxbandit:
Note, project's Save Actions include "Format all lines", so a number of unrelated lines contain formatting changes.

laeubi:
@jonahgraham @mickaelistria @akurtakov I think this clearly shows that the "bandaiding" do not work/scale well... We can't put a restriction on every eclipse plugin to use CTab(Folder) just to improve dark-mode users... That will effectively require deprecate usage of TabFolder in all Platform code and every plugin written for eclipse...

jonahgraham:
> @jonahgraham @mickaelistria @akurtakov I think this clearly shows that the "bandaiding" do not work/scale well... We can't put a restriction on every eclipse plugin to use CTab(Folder) just to improve dark-mode users... That will effectively require deprecate usage of TabFolder in all Platform code and every plugin written for eclipse...

Please continue this discussion on https://github.com/eclipse-platform/.github/issues/27

laeubi (changes requested):
Requires discussion at https://github.com/eclipse-platform/.github/issues/27

PR #149, opened 2022-11-02
Switching to macos sdk version 11

sravanlakkimsetti:
tracked in
https://github.com/eclipse-platform/eclipse.platform.swt/issues/452


iloveeclipse:
Could it be, this is the reason for gszillions of test fails in  https://download.eclipse.org/eclipse/downloads/drops4/I20221102-0910/testResults.php ?

Phillipus:
Perhaps best not to change to SDK 11 if this is not the cause of https://github.com/eclipse-platform/eclipse.platform.swt/issues/452 ? Many users are still using Catalina.

sravanlakkimsetti:
Reverted the change to SDK 11 now

Issue #159, opened 2022-11-15
Open File Handle on Bundle


juergen-albert:
With bnd we make have use of the on the fly Bundle reload after sources are build. Up to equinox 3.17.100.v20211104-1730 this works flawless. Starting with 3.17.200.v20220215-2237 until the latest release version (3.18.100 I believe), bndtools starts to complain under Windows that it can't the delete the bundle jar anymore as another process is still using it. 

It seems that the FileHandel is not released after install. As it works with Felix and with the mentioned older Versions, I believe it to be something like an unclosed InputStream in equinox as the bnd code here looks pretty solid in that regard.

HannesWell:
Can you trace somehow where the such unclosed resource is opened? It will then likely be trivial to fix it.

laeubi:
Not sure if OSGi is required to release any file handles (e.g. it might optimize to not read everything into memory and thus keep the jar open to keep reading stuff from it if requested for a while), but why not just using the `org.osgi.framework.BundleContext.installBundle(String, InputStream)` and `org.osgi.framework.Bundle.update(InputStream)` then you have full control of the closing of the stream (even though the API states the stream is always closed anyways).

juergen-albert:
@HannesWell Not really. I can only say that bnd uses the various `BundleContext.installBundle` methods and Windows and Java is usually not really forthcoming with the culprit. 

The only thing we see is:
![image002](https://user-images.githubusercontent.com/8199490/201981265-45d2422a-9d09-4fd9-9e67-747781be8f9c.png)

and that the file is held by javaw.exe

HannesWell:
> @HannesWell Not really. I can only say that bnd uses the various `BundleContext.installBundle` methods and Windows and Java is usually not really forthcoming with the culprit.

To bad, because I have a similar problem with Xtext and don't know a elegant way to find such resource leaks either.

The only answswer I found in the internet was to do a heap analysis (e.g. with VisualVm) and search for open Stream objects.

So unless somebody else knows a good way to trace them, we probably have to check what has changed since then.

juergen-albert:
@tjwatson The Bnd launcher explicitly deactivates references when it is running under windows. Don't ask my why, maybe @pkriens can tell us why.

@bjhargrave I have tried the file leak detector activ in eclipse and in the running framework that seems to be the culprit. Both constantly say `0 descriptors are open` even while I can see the open file in the Windows ResourceMonitor.

I can see the issue here at least.
3.17.100
```
[...]

PR #160, opened 2022-11-15
Ensure InputStream passed to StorageUtil.readFile() is closed again

HannesWell:
Fixes https://github.com/eclipse-equinox/equinox/issues/159


HannesWell:
After analyzing the problem a bit more it looks like this fixes a not to be neglected resource-leak and it would probably be good to have that in RC1.

@tjwatson can you please review and approve/commit this, if you think it is good and more approvals are not required.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
```diff
@@ -1115,7 +1115,7 @@ File stageContent(final InputStream in, final URL sourceURL) throws BundleExcept
 
 	File stageContent0(InputStream in, URL sourceURL) throws BundleException {
 		File outFile = null;
-		try {
+		try (InputStream stream = in) {
```
HannesWell, line 1118: Before 68fa0c03747cf77307616b0fc644757a81e39578 the given `InputStream` was closed, in case the `sourceURL` uses the file protocol and the specified file (`inFile`) is not a directory.
It wasn't closed if `inFile` was a directory (was this a bug?).
After the mentioned commit the `InputStream` was only closed if another protocol than `file` was used.
With this change it is always closed-

laeubi (approved):
Looks good

vogella:
Approved for RC1

bjhargrave (approved):

tjwatson (approved):
LGTM, thanks!

HannesWell:
Great. Thanks!

PR #164, opened 2022-11-24
Remove "javax.servlet.*" upper version limit

### LINKED ISSUE #163: Make "org.eclipse.equinox.http.servlet" bundle compatible with Tomcat Migration Tool
Apache Tomcat provides a [Tomcat Migration Tool for Jakarta EE Software](https://tomcat.apache.org/download-migration.cgi) for automatic migration from "javax.servlet" to "jakarta.servlet" namespace. To make "org.eclipse.equinox.http.servlet" bundle compatible with this tool, the upper version limit for "javax.servlet.*" packages must be removed from the [MANIFEST.MF](https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.equinox.http.servlet/META-INF/MANIFEST.MF#L17).

A 5+ version of "javax.servlet" is extremely unlikely, and therefore the upper version constraint is useless.

This change is very important for Eclipse RAP project in order to make RAP application running on Jakarta EE 9+ container.

ifurnadjiev:
Apache Tomcat provides Tomcat Migration Tool for Jakarta EE Software that automatically changes "javax" to "jakarta" namespace. In order to provide a compatible bundle (Jakarta EE 9+ API has version 5+) the upper version limit must be removed.
More over, a 5+ version of "javax.servlet" is extremely unlikely, and therefore the upper version constraint is useless.

Fix #163


laeubi (changes requested):
I'm not sure if it is the right approach to remove valid version ranges to make it compatible with a "migration tool" (why do I need to change my code if it "automatically" changes things as claimed here?)

> More over, a 5+ version of "javax.servlet" is extremely unlikely, and therefore the upper version constraint is useless.

Version ranges are not about how "likely" a new version is and we can't change them just to make some tool happy, so you should describe what is the exact issue here.

laeubi:
@tjwatson maybe this is also a candidate for bundle deprecation? `HttpService` will no longer evolve and if people like to use it they can use older releases... or are there any consumers in platform that require this? If yes we should consider to update the spec to jakarta, probably adding "extended" httpservice methods as well.

ifurnadjiev:
Apache Tomcat 10 is based on Servlet API with "jakarta" namespace instead "javax". In order to use "old" web apps (war files) that use "javax" namespace in Tomcat 10, there is a special folder "for deployment "webapp-javaee" or standalone tool.  In both approaches the result is a replacement of "javax" namespace with "jakarta" namespace in all the classes, manifests, but does not change the version constraints. As a result in the "migrated" war file there is a wrong version range like:
`jakarta.servlet;version="[3.1.0,5.0.0)"`
and the bundle "org.eclipse.equinox.http.servlet" can not be resolved. (jakarta.servlet package has version 5).

By removing the upper version limit from "javax.servlet.*" packages in MANIFEST.MF allows as to use war files, based on "javax" namespace directly in Tomcat 10 by putting them in deployment folder "webapp-javaee". For other containers, based on Jakarta EE 9+, the standalone Tomcat migration tool can be used to have a deployable application (war file) with "jakarta" namespace.

laeubi:
> In both approaches the result is a replacement of "javax" namespace with "jakarta" namespace in all the classes, manifests, but does not change the version constraints.

Then probably the tool should better be fixed to also insert an appropriate version range.

mknauer:
Typically, I completely agree with you, because version constraints are actually useful and helpful, and in this case also defined correctly. Actually. :slightly_smiling_face:

From the consumer side, version constraints should ensure that the bundle is not used outside its spec. But since there will be no Servlet API in the `javax` namespace with version 5+ in the future (perhaps we should have phrased it as "_never-ever_" rather than "_unlikely_"), there is no need to constrain anything.
De-facto, however, the existing version constraint hinders a consumer, an example of this is the automatic migration tooling, but perhaps there are other use cases for this.

Therefore, I believe that in this particular case, removing the upper version limit is not only a simple and safe solution, but also a pragmatic one. Especially since I would like to switch completely to the `jakarta` namespace in the medium term - who knows whether this bundle will still exist then.

laeubi:
> De-facto, however, the existing version constraint hinders a consumer, an example of this is the automatic migration tooling, but perhaps there are other use cases for this.

Don't think so if you say "never ever" no consumer can "never ever" be affected. A migration **must take changing version bounds into consideration** it is simply invalid to assume that the original author can anticipate any "valid" range in such a case.

Beside that, changing multiple items to conform with a converter do not scale well, while enhancing the converter itself will fix the problem for all consumer.

mknauer:
Don't get me wrong, I don't disagree with you. The automated migration should be improved, and we already opened issue apache/tomcat-jakartaee-migration/issues/39 there. But I see this as something that can be done additionally and independently of this change.

My intention is only to give consumers every possible option for easy migration and to facilitate it with a comparatively small change without losing too much. Surely, if you conclude that the risk of this change is greater than its benefit, then that is also a valid decision of the project.

I didn't really want to start a discussion on versions: My main goal remains - to really remove all remnants of `javax` in the HTTP/servlet area in the next few months so that all major Eclipse projects use `jakarta` exclusively. That would be cool for the community.

laeubi:
> But I see this as something that can be done additionally and independently of this change.

But why? It doesn't make sense to cripple valid version ranges to fix bugs in other tools.



> My main goal remains - to really remove all remnants of `javax` in the HTTP/servlet area in the next few months so that all major Eclipse projects use `jakarta` exclusively.

That would be great.

merks:
Sometimes the pragmatic thing and the right thing are not the same thing.

laeubi:
> Sometimes the pragmatic thing and the right thing are not the same thing.

And sometimes the pragmatic things are just wrong, you can't judge every hack as "pragmatic" and useful in general...

If I wanted to be pragmatic, I would download the jar, modify the manifest to whatever seems sufficient and use this artifact to "convert" it.

Beside that one could question if it makes any sense to convert this particular bundle, because it instantly would break every user of that service ... so one needs to "convert" each and every bundle that is using this service as well. Then one needs to check if it not uses DS or blueprint what most likely is not understood by the converter as well...

laeubi:
As the initial issue is now fixed upstream:
- https://github.com/apache/tomcat-jakartaee-migration/issues/39
I think there is no more reason for this.

@ifurnadjiev if you think there is still something to change let us know.

ifurnadjiev:
@laeubi Do you have any plans to make the plugins Jakarta compatible?
- org.eclipse.equinox.http.servlet
- org.eclipse.equinox.http.servletbridge
- org.eclipse.equinox.http.jetty

laeubi:
As these require the (deprecated) http service and the OSGi alliance don't plan to upgrade that specification its very unlikely do happen.

That said, as both JavaX and Jakarta live in different namespaces there are some options to support both but this will require a lot of efforts and you likely better migrate to a more modern approach / implementation e.g. [Apache Felix](https://felix.apache.org/documentation/index.html) Whiteboard Servlet.

PR #180, opened 2023-01-05
Use connect classloader for ConnectBundleEntry getLocalURL

### LINKED ISSUE #179: ConnectBundleEntry getLocalURL throws UnsupportedOperationException
When running Eclipse using the Atomos connector, `platform://` URLs get resolved as `ConnectBundleFile` rather than `FileBundleEntry`. That's not a problem in itself, except for this difference

https://github.com/eclipse-equinox/equinox/blob/bfbcf87790b03a42fa0b78cff2c65561208ba6a6/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/bundlefile/FileBundleEntry.java#L94-L97

https://github.com/eclipse-equinox/equinox/blob/bfbcf87790b03a42fa0b78cff2c65561208ba6a6/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/connect/ConnectBundleFile.java#L79-L83

I bumped into this problem while implementing

- https://github.com/equodev/equo-ide/pull/36

In case it's useful, a testcase which reproduces this is to clone https://github.com/equodev/equo-ide at `7b46b2a2bd5dcd6be32a9711915983f394aa55ce` and either

- run `./gradlew IdeMainTest`
- or `./gradlew eclipse`, import the Solstice project into Eclipse, and run the main in `IdeMainTest`

To work around the `UnsupportedOperationException` I added this little hack https://github.com/equodev/equo-ide/pull/36/commits/fd0bb6f1c2cc36368618f77684261e7c9973e50f

tjwatson:
Fixes #179

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/url/BundleURLConnection.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/url/BundleURLConnection.java
@@ -64,7 +64,7 @@ public class BundleURLConnection extends URLConnection implements BundleReferenc
 				in = bundleEntry.getInputStream();
 				connected = true;
 			} else {
-				throw new IOException(NLS.bind(Msg.RESOURCE_NOT_FOUND_EXCEPTION, url));
+				throw new IOException(NLS.bind(Msg.RESOURCE_NOT_FOUND_EXCEPTION, getURL()));
 			}
 		}
 	}
@@ -132,7 +132,7 @@ public class BundleURLConnection extends URLConnection implements BundleReferenc
 	 */
 	public URL getLocalURL() {
 		URL local = bundleEntry.getLocalURL();
-		return local == null ? url : local;
+		return local == null ? getURL() : local;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/url/BundleURLConnection.java
```diff
@@ -131,7 +131,8 @@ public long getLastModified() {
 	 * @return the local URL using a common local protocol
 	 */
 	public URL getLocalURL() {
-		return bundleEntry.getLocalURL();
+		URL local = bundleEntry.getLocalURL();
+		return local == null ? url : local;
```
bjhargrave, line 135:
```suggestion
		return local == null ? getURL() : local;
```

Wouldn't it be better to use the accessor method rather than a field reference? :-)

tjwatson: Its a protected field from super that is accessed directly from other code in this class.  I would like to stay consistent so perhaps use `getURL` in all locations `url` is accessed here?

PR #181, opened 2023-01-10
Status: make more threadsafe

jukzi:
Members are not final but used across threads.

### DIFF
--- a/bundles/org.eclipse.equinox.common/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.common/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.common; singleton:=true
-Bundle-Version: 3.17.0.qualifier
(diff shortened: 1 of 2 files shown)


jukzi:
There was an error that i cannot explain otherwise where a wrong "status" was read: https://github.com/eclipse-platform/eclipse.platform/pull/310#issuecomment-1377072070

laeubi:
I think the main question is if status should be threadsafe for concurrent access and shared between different threads concurrently... So I would assume that a status object is constructed, initlized and maybe modified and the passed alon and is the never modified...

laeubi:
By the way making it "volativle" does not makes is "more threadsafe.... that would only be relevant if different threads see different states, but still if one thread modifies two fileds another thread might see an intermediate state...

So I must confess changing code just for trial-and-error without any evidence that it helps seems not desirable.

laeubi:
e.g. `org.eclipse.core.runtime.MultiStatus.add(IStatus)` is more likely a source where concurrency can cause issues but then **all** methods must be synchronized!

laeubi:
Another one `org.eclipse.core.internal.databinding.BindingStatus.add(IStatus)` again, adding volatile won't help in any of those cases.

PR #184, opened 2023-01-17
Add the official OSGi TCKs to test the compliance of Equinox

laeubi:
This adds a new job that checks the compliance of equinox against the offical OSGi TCK test suite as applicaiable.


laeubi:
Some smaller cleanups and version bumps are required, but it in general it is working now @tjwatson do you have any concerns or remarks?

laeubi:
Jenkins is green so I assume the failure is unrelated.

laeubi:
This is now ready to be merged as a first step, if no one objects I'll merge this now so further improvements can base on this.

akurtakov:
If the failing tests are to be worked on this should be fine.

laeubi:
> If the failing tests are to be worked on this should be fine.

I don't know how "compliant" equinox is supposed to be but at least we should be aware that it currently fails the official tests in some cases.

Actually I though I have excluded them from being displayed in the Github result,  they are still picked up for some reasons here, I'll check that.

laeubi:
Alright there was a wrong condition on the junit publisher and thus the results here was not updated, fixed that lets see if it works on the next run...

laeubi:
This works now so it is ready to be merged!

tjwatson:
Can you point me to where we see the test results?

tjwatson:
I see this:

```
[INFO] [bundle] Coordinator 1.5.200-SNAPSHOT .............. SUCCESS [01:28 min]
[INFO] [bundle] Configuration Admin 1.5.200-SNAPSHOT ...... FAILURE [17:32 min]
[INFO] [bundle] Eclipse Preferences Mechanism 3.10.200-SNAPSHOT SUCCESS [01:14 min]
[INFO] [bundle] User Admin Service 1.2.400-SNAPSHOT ....... SUCCESS [  3.529 s]
[INFO] [bundle] Meta Type 1.6.300-SNAPSHOT ................ SUCCESS [  5.019 s]
[INFO] [bundle] Device Access Service 1.1.400-SNAPSHOT .... FAILURE [05:05 min]
[INFO] [bundle] Event Admin 1.6.200-SNAPSHOT .............. SUCCESS [ 29.873 s]
[INFO] parent 4.27.0-SNAPSHOT ............................. SUCCESS [  0.021 s]
```

IIRC when updating to the latest SCR we needed to also update config admin to the latest spec.  One aspect of that update was some integration with the coordinator.  I never implemented that aspect of config admin for Equinox.  I think that is probably why Configuration admion is failing the TCK.  I'm not sure why Device Access Service is failing.

laeubi:
It is a bit hidden at the moment I plan to improve this once the general setup works.

grafik
grafik
grafik

The CM test currently require an extension from Tycho, to enable security, but this should be available shortly, then only fife tests fail for cm...

laeubi:
We now have 3 tests failing on the device spec and 5 tests failed at the cm spec, so I think all is fine in general and we can investigate this further later on if this are real issues or require additional investigation.

PR #189, opened 2023-01-20
Create README.md

laeubi:
@tjwatson @akurtakov WDYT? At laest better than nothing :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,18 @@
+![](https://www.eclipse.org/equinox/images/logo.png)
+
+# Mission Statement
+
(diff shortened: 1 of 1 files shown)


laeubi:
Preview can be found here:
https://github.com/eclipse-equinox/equinox/blob/4f056d20dd4fe60219280147b40e4caffa0d1e6c/README.md

akurtakov:
I fully agree. Merging as otherwise discussions will prolong too much. Editing after that is simple enough even with a plain browser so improvements are more than welcome.

PR #198, opened 2023-01-23
Add core TCKs implemented by org.eclipse.osgi

laeubi:
FYI @tjwatson I think this should cover the most important parts equinox implements, is anything missing?

I also added resolver even though it seems the sources are imported from Felix (are these synchronized from time to time?) probably we should test SCR even if it is provided by Felix and not part of Equinox code base?

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -52,6 +52,10 @@ jobs:
       fail-fast: false
       matrix:
         tck: 
+          - { name: framework,   chapter:  '10', label: 'Framework',           suffix: ' API',                   pattern: "TEST-org.osgi.test.cases.framework-*.xml" }
+          - { name: url,         chapter:  '52', label: 'URL Handlers',        suffix: ' Service Specification', pattern: "TEST-org.osgi.test.cases.url-*.xml" }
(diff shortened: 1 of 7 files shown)


tjwatson:
> FYI @tjwatson I think this should cover the most important parts equinox implements, is anything missing?
> 
> I also added resolver even though it seems the sources are imported from Felix (are these synchronized from time to time?) probably we should test SCR even if it is provided by Felix and not part of Equinox code base?

Thanks!

Resolver tests are good to run.  That will verify we register the service correctly.  The source is kept in sync.  I maintain both sources.  Often making the fix in Equinox, then merging the fix into Felix later.  SCR tests are also good to run in my opinion.  Again I am a maintainer of the Felix SCR implementation.  But good to verify the TCK against the impl running on Equinox.

tjwatson:
If we are going to run the Declarative services TCK, perhaps it is time to delete our old ds.tests.  I am doubtful they cover anything useful beyond the tests already run in the felix project itself or the TCK.  Given they have a failing test in our github flow that we cannot track down also motivates me to delete the ds.tests.

akurtakov:
> If we are going to run the Declarative services TCK, perhaps it is time to delete our old ds.tests. I am doubtful they cover anything useful beyond the tests already run in the felix project itself or the TCK. Given they have a failing test in our github flow that we cannot track down also motivates me to delete the ds.tests.

Huge +1.

PR #199, opened 2023-01-24
Add a method that returns the registry as an optional

laeubi:
As per API contract the RegistryFactory can return null, this has some drawbacks:
- callers of the API might be not well aware of this
- additional null checks are required on the caller site

this adds a new registry() method that reaturns an optional and makes it clear to the caller that there might be a chance of getting "nothing" and also allows to convienetly hande the case using modern programming primitives like lamdas or handler chaning.

### DIFF
--- a/bundles/org.eclipse.equinox.registry/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.registry/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.registry;singleton:=true
-Bundle-Version: 3.11.200.qualifier
+Bundle-Version: 3.12.0.qualifier
 Bundle-Localization: plugin
 Export-Package: org.eclipse.core.internal.adapter;x-internal:=true,
  org.eclipse.core.internal.registry;x-friends:="org.eclipse.core.runtime",
--- a/bundles/org.eclipse.equinox.registry/src/org/eclipse/core/runtime/RegistryFactory.java
+++ b/bundles/org.eclipse.equinox.registry/src/org/eclipse/core/runtime/RegistryFactory.java
@@ -14,6 +14,7 @@
 package org.eclipse.core.runtime;
 
 import java.io.File;
+import java.util.Optional;
 import org.eclipse.core.internal.registry.ExtensionRegistry;
 import org.eclipse.core.internal.registry.RegistryProviderFactory;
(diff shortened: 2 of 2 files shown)


HannesWell:
While I agree with all your arguments in favour of returning an Optional instead of null, I'm not sure if it really helps users to just add another method that wraps the result of an existing method in an Optional without fading out the existing method (at least marking it as deprecated).
Furthermore ideally all methods should either return null or Optional (while today one would probably choose optional) but mixing null and Optional sporadically is IMHO not optimal.

I'm not against this, but I also don't think it is the best invention since sliced bread. :)

laeubi:
In the past, deprecation was not really desired (because of "gazillion" of warnings) so I currently see no way other than _recommend_ the new method in the javadoc...

vogella:
+1 and I think it is fine to deprecate, if we introduce new apis

mickaelistria:
> I think it is fine to deprecate, if we introduce new apis

As mentioned, the issue is that just deprecating would add tons of warnings into Platform code. IMO, if we go deprecating something, Platform must lead by example and -in the same release- remove all references to older APIs and use the newer one instead.
Migration of Platform to newer API is usually a part or even a requirement in the deprecation process.

HannesWell:
> In the past, deprecation was not really desired (because of "gazillion" of warnings) so I currently see no way other than _recommend_ the new method in the javadoc...

I'm probably pushing an open door, but isn't that exactly the purpose of deprecation to cause that warnings so that users become aware of the deprecation? If one uses the old method a new recommendation is probably never read in the Java-doc.
On the other hand for those that want to maintain a longer

On the other hand for those that don't want to use such new method immediately to maintain a longer backwards compatibility it can be annoying, but I would suggest to suppress the deprecation warning. Then each time one goes through that code one gets a visual hint to re-evaluate if the new method can be used then.

> > I think it is fine to deprecate, if we introduce new apis
> 
> As mentioned, the issue is that just deprecating would add tons of warnings into Platform code. IMO, if we go deprecating something, Platform must lead by example and -in the same release- remove all references to older APIs and use the newer one instead. Migration of Platform to newer API is usually a part or even a requirement in the deprecation process.

I don't see a problem in doing that. If one introduces a new API while deprecating another in favor of the new one it means that the new one is superior over the old one. I would then be more than happy to apply the new one in the platform to show-case its superiority. :)

laeubi:
I found that there are `@noreference` annotation as a hint to not reference a method (anymore).

mickaelistria:
> I found that there are @noreference annotation as a hint to not reference a method (anymore).

Those would also generate a warning, so they're not better than `@Deprecated` and have less discoverability, less semantics; they should only be used in the corner-case of an internal method which is made visible for a specific consumer in some different package... Usually a case that should be avoided.
So IMO, good old `@Deprecated` is correct here.

PR #204, opened 2023-01-30
Touch o.e.equinox.core.feature for felix.scr update

akurtakov:


akurtakov:
I tried similar approach with test-framework https://github.com/eclipse-platform/eclipse.platform.releng/commit/73371e99d6f2cd255e50aa5533b0697a3aa09562 and requirements didn't end up in the p2 site. I would really welcome if someone gives a hand on this one to investigate deeper.

laeubi:
I just checked equinox.event that has:

```
Require-Capability: osgi.extender;
 filter:="(&(osgi.extender=osgi.component)(version>=1.0)(!(version>=2.0)))"
```

this should pull in felix-scr (or more formally at least one scr implementation), so I think one should be able to even remove the scr dependency completely.

akurtakov:
OK, I'll give it a try.

PR #205, opened 2023-01-30
Do not list felix.scr expliictly.

akurtakov:
It should be pulled in as a provider of
osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.0)(!(version>=2.0)))"

Change is an effort to reduce the need to "touch" features everytime 3rd party dependency is being updated.

### DIFF
--- a/features/org.eclipse.equinox.core.feature/feature.xml
+++ b/features/org.eclipse.equinox.core.feature/feature.xml
@@ -101,13 +101,6 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.felix.scr"
-         download-size="0"
-         install-size="0"
-         version="0.0.0"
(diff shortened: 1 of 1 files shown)


laeubi:
@akurtakov what do you think about having a `site` in equinox produced (enabled by a profile) that builds an sdk like updatesite as part of the verification action? Then it would be possible to verify such changes much easier (no Ibuild required) and we can even publish a snapshot updatesite (like m2e / tycho) after each master build. I think @HannesWell has done the parts on m2e so probably we can use the same technique here?

akurtakov:
That would be very welcome .

akurtakov:
Jenkins build succeeds thus merge.

Issue #209, opened 2023-02-01
ServiceCaller should be either more robust or better documented


ruspl-afed:
`ServiceCaller` [states](https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/ServiceCaller.java#L134)
```
	 * @throws NullPointerException if any of the parameters are null
```
However, if all the parameters are not `null` it can still throw NPE from constructor, since `FrameworkUtil.getBundle(caller)` could return `null` [here](https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/ServiceCaller.java#L295).

I would consider it as a bug, since current behavior contradicts with declared API contract. But may be there is another point of view regarding where the change should be done to make both code an docs consistent.

laeubi:
> I would consider it as a bug, since current behavior contradicts with declared API contract.

I would consider it a but using `ServiceCaller` at all ... especially if it is used anywhere where `FrameworkUtil.getBundle(caller)` can return null, so probably it should better throw IllegalArgument Exception then.

tjwatson:
I agree, an `IllegalArgumentException` is more appropriate.  What calling class would be using `ServiceCaller` that has no `Bundle`?

laeubi:
> I agree, an `IllegalArgumentException` is more appropriate. What calling class would be using `ServiceCaller` that has no `Bundle`?

Some code uses `ServiceCaller` in stupid ways e.g. in an Activator(!) as a static initialized field, if you use such bundles outside OSGi and unluckily the Activator class is initialized it fails the class-loading with NPE ... but that's just a guess as I noticed that when working on Tycho.

ruspl-afed:
My use case is running the plain junit test for a type that triggers `ServiceCaller`-dependent code for one of its dependencies. 
Do we have a good way to understand does particular type requires OSGI-running or not?

I used to prefer classes from "org.eclipse.equinox.common" bundle, that are either OSGi-neutral or robust enough to not fail without it. `ServiceCaller` seems to be different.

laeubi:
I think the best would be to try fix that code, but `ServiceCaller` could also probably just return as if no service can be found if called outside OSGi.

ruspl-afed:
Of cause I'll fix the code, since I have no other options for now. 
But, the static methods of `ServiceCaller` could handle this case.
[...]

PR #210, opened 2023-02-01
ServiceCaller should be either more robust or better documented #209

ruspl-afed:
Throw IllegalStateException if FrameworkUtil#getBundle(Class) returns `null` for a given caller and add this to javadoc

Signed-off-by: Alexander Fedorov <alexander.fedorov@arsysop.ru>

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/ServiceCaller.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/ServiceCaller.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2022 IBM Corporation and others.
+ * Copyright (c) 2020, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/ServiceCaller.java
```diff
@@ -147,7 +150,9 @@ public static <S> boolean callOnce(Class<?> caller, Class<S> serviceType, Consum
 	 * @param <S>         the OSGi service type to look up
 	 * @return true if the OSGi service was located and called successfully, false
 	 *         otherwise
-	 * @throws NullPointerException if any of the parameters are null
+	 * @throws NullPointerException  if any of the parameters are {@code null}
+	 * @throws IllegalStateException if {@link FrameworkUtil#getBundle(Class)}
```
tjwatson, line 154: This exposes implementation details that could change in the future.  Instead can we have something like:

> if the bundle associated with the caller class cannot be determined

tjwatson (approved):
LGTM

tjwatson:
fixes #209

PR #212, opened 2023-02-07
Name of the repo has changed, move profile activation to maven.config

laeubi:


.mvn/maven.config
```diff
@@ -0,0 +1,2 @@
+-Pbuild-individual-bundles
+-Pbree-libs
```
akurtakov, line 2: Having -Pbree-libs in maven.config means every contributor has to setup toolchains.xml (and for equinox it even includes 1.8) which is too much of a hassle even more me to setup. Thus I believe this profile should be kept in build.yml and Jenkinsfile.
Others like -Dcompare-version-with-baselines.skip=false  and -Dproject.build.sourceEncoding=UTF-8 on the other side make perfect sense here.

laeubi: This reminds me about the following feature:
- https://github.com/eclipse-tycho/tycho/issues/1163

laeubi: @merks just wondering, could JustJ provide java 8 at all or is there any requirement it can only start from java 11?

akurtakov: This might be nice to have for some people but I think it's best to finish PRs with what's available today :).
P.S. JustJ doesn't provide 1.8 JVMs so that would not help here.

laeubi: I just opened:
- https://github.com/eclipse-justj/justj/discussions/7

I think it would be good to have local builds being as close as possible and if I remember right @tjwatson mentioned Java 8 should stay for the "core" OSGi for some time.

akurtakov: Back in the days there was no proper openjdk for Java 8 IIRC but things might have changed.

tjwatson: I don't think we have any plans to run tests on Java 8.  I also do not think Java 8 is necessary to compile against with modern compilers that should use `-release` to ensure you don't compile in dependencies to APIs that are not available on the target release.

If we had plans to run tests on Java 8 then it may be more of a requirement.  But I would not want to force such a burden onto the project.

merks: The process for building a JRE from a JDK relies on jlink which was introduced when the JRE was modularized, so there would be no automated way of provide the range of JREs that we do for Java 11 and higher.  Instead would be a significant hand-written effort to provide a single full JRE, and then it would be something that would not be widely consumed...  In older JDKs there was a JRE available directly nested:

https://adoptium.net/temurin/releases/

Jenkinsfile
```diff
@@ -23,10 +23,10 @@ pipeline {
 			steps {
 				sh """
 				mvn clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
-					-Pbuild-individual-bundles -Pbree-libs -Papi-check -Pjavadoc\
+					-Papi-check -Pjavadoc\
 					-Dcompare-version-with-baselines.skip=false \
 					-Dproject.build.sourceEncoding=UTF-8 \
```
HannesWell, line 28:
```suggestion
```
The `sourceEncoding` is specified in the parent-pom as well and therefore can be removed.

.mvn/maven.config
HannesWell: > Having -Pbree-libs in maven.config means every contributor has to setup toolchains.xml (and for equinox it even includes 1.8) which is too much of a hassle even more me to setup. Thus I believe this profile should be kept in build.yml and Jenkinsfile.

Agree on that.

.github/workflows/build.yml
```diff
@@ -36,13 +36,11 @@ jobs:
         --batch-mode
         --global-toolchains ${{ github.workspace }}/.github/workflows/toolchains.xml
         --threads 1C
-        -Pbuild-individual-bundles
-        -Pbree-libs
         -Papi-check
         -Dcompare-version-with-baselines.skip=false
         -Dproject.build.sourceEncoding=UTF-8
```
HannesWell, line 41:
```suggestion
```
The `sourceEncoding` is specified in the parent-pom as well and therefore can be removed.

PR #215, opened 2023-02-11
Disable Header processing for org.eclipse.osgi as it uses special filter

laeubi:
org.eclipse.osgi uses some very special filters that Tycho (currently) cannot merge successfully, so we should disable the processing for now.

### DIFF
--- a/bundles/org.eclipse.osgi/pom.xml
+++ b/bundles/org.eclipse.osgi/pom.xml
@@ -39,6 +39,14 @@
         </configuration>
       </plugin>
       <plugin>
+	    <groupId>org.eclipse.tycho</groupId>
+	    <artifactId>tycho-packaging-plugin</artifactId>
+	     <configuration>
+			<!-- equinox uses very specially crafted manifest headers -->
+			<deriveHeaderFromSource>false</deriveHeaderFromSource>
+		 </configuration>
+	  </plugin>
+      <plugin>
        <!--Unless next tycho release we need to explicitly enable this here, once 2.7.4 or 3.x is out we can switch to full pomless here! -->
        <groupId>biz.aQute.bnd</groupId>
         <artifactId>bnd-maven-plugin</artifactId>


laeubi:
FYI @akurtakov 

As in 
```
1.  equinox/bundles/org.eclipse.osgi/pom.xml
   no-classifier: different
      META-INF/MANIFEST.MF: different
         Require-Capability: baseline='osgi.ee; filter:="(| (&(osgi.ee=JavaSE)(version=1.8)) (&(osgi.ee=JavaSE/compact1)(version=1.8)) )"' != reactor=
                                      'osgi.ee;filter:="(| (&(osgi.ee=JavaSE)(version=1.8)) (&(osgi.ee=JavaSE/compact1)(version=1.8)) )",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"'
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]
```

probably I can find a way to properly merge filter-strings, but untill then I think it is more safe to simply keep it as it for org.eclipse.osgi.

@tjwatson is there any specific use-case for `JavaSE/compact1`?

merks:
Thinking ahead to when a person moves their project to a version of Tycho with this new support, how does one detect that the Require-Capability is not successfully merged?   This one is the only problematic one because it already uses osgi.ee?

image

laeubi:
I have searched the platform and only found this one actually using `osgi.ee` (your screenshot uses `osgi.extender`) and duplicates are filtered out, but as here the filter string diverged from that what is generated it can't be replaced.

I don't know if Equinox (@tjwatson) or BND (@bjhargrave) has a function to compute a compound filter that would absorb this because obviously the existing filter string (`(| (&(osgi.ee=JavaSE)(version=1.8)) (&(osgi.ee=JavaSE/compact1)(version=1.8)) )`) already includes the newly added (`(&(osgi.ee=JavaSE)(version=1.8))`)

laeubi:
I think it might even be valid to simply skip adding `osgi.ee` if there is already one ... I'll prepare a patch for that.

merks:
I see, so only if there is an osgi.ee requirement already is there potentially an issue.  That should be easy to detect, I guess, and then instead of merging or adding something, perhaps one should then do nothing with regard to producing or merging an osgi.ee requirement.   I.e., if one is already using the non-deprecated form in the MANFIEST, I wouldn't expect anything should be changed.  Or?

Given that this MANIFEST.MF does not contain `Bundle-RequiredExecutionEnvironment`, why does the issue arise at all?  Where is Tycho getting the information about the osgi.ee requirement other than from the `Bundle-RequiredExecutionEnvironment`?

laeubi:
Tycho uses bnd-lib to analyze the produced classfiles and the lowest class version number is then used as the required `osgi.ee`.

laeubi:
I now created
- https://github.com/eclipse-tycho/tycho/pull/2087

so this is not needed anymore.

tjwatson:
> @tjwatson is there any specific use-case for `JavaSE/compact1`?

I don't think so, not even sure `org.eclipse.osgi` can actual run on compact1 anymore.  Been a really long time since I've tried that.  IMO the compact profiles of Java 8 were a failure and only existed in Java 8 to be removed in Java 9+ when JPMS got added.  If it makes things easier we can make the ee filter of `org.eclipse.osgi` only reference full Java 8.

akurtakov:
At least it will be a prep for moving up one day so I say go for it.

PR #216, opened 2023-02-11
Fix for NPE in PackageSource.toString()

iloveeclipse:
Seen debugging JDT UI Leak test & printing the references

```
java.lang.NullPointerException: Cannot read the array length because
"value" is null
	at java.base/java.lang.String.<init>(String.java:275)
	at java.base/java.lang.String.valueOf(String.java:4229)
	at org.eclipse.osgi.internal.loader.sources.PackageSource.toString(PackageSource.java:90)
	at java.base/java.lang.String.valueOf(String.java:4215)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:173)
	at java.base/java.util.HashMap$Node.toString(HashMap.java:296)
	at java.base/java.lang.String.valueOf(String.java:4215)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:173)
	at org.eclipse.jdt.ui.leaktest.reftracker.ReferencedObject.toString(ReferencedObject.java:49)
	at java.base/java.lang.String.valueOf(String.java:4215)
	at java.base/java.lang.StringBuilder.append(StringBuilder.java:173)
	at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.followFieldReference(ReferenceTracker.java:75)
```

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/loader/sources/PackageSource.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/loader/sources/PackageSource.java
@@ -87,7 +87,7 @@ public abstract class PackageSource {
 		builder.append(id).append(" -> "); //$NON-NLS-1$
 		SingleSourcePackage[] suppliers = getSuppliers();
 		if (suppliers == null) {
-			return builder.append(String.valueOf(null)).toString();
+			return builder.append("null").toString(); //$NON-NLS-1$
 		}
 		builder.append('[');
 		for (int i = 0; i < suppliers.length; i++) {


tjwatson:
The change is probably better than before.  But I am curious what Java version and implementation this is on because the Javadoc for String.valueOf clearly states:

```
if the argument is null, then a string equal to "null"; otherwise, the value of obj.toString() is returned.
```

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#valueOf(java.lang.Object)

iloveeclipse:
This sounds surprising, but in fact both ecj and javac compile the code to use `String.valueOf(char[])` and **not**  `String.valueOf(Object)` 
```
public class Y {
    public static void main(String [] args) {
		System.out.println(String.valueOf(null));
	}
}
```

```
/usr/lib/jvm/java-17/bin/java /data/runtime-Java_11/Java17/src/Y.java
Exception in thread "main" java.lang.NullPointerException: Cannot read the array length because "value" is null
        at java.base/java.lang.String.<init>(String.java:275)
        at java.base/java.lang.String.valueOf(String.java:4232)
        at Y.main(Y.java:3)
```

So it is not about which JVM I've used but why compiler selects `String.valueOf(char[])` and **not**  `String.valueOf(Object)`?
As a naive Java developer I honestly would expect some warning like "ambiguous argument", but neither compiler emits any warning...

And this below is what makes the program work "as expected"
```
public class Y {
    public static void main(String [] args) {
		System.out.println(String.valueOf((Object)null));
	}
} 
```

@stephan-herrmann : could you enlighten please why the compiler do that above without warning?

laeubi:
Beside this interesting compiler issue, `StringBuilder.append(null)` *should* give the same result, or will it also select the wrong?

tjwatson:
> Beside this interesting compiler issue, `StringBuilder.append(null)` _should_ give the same result, or will it also select the wrong?

That would depend on which `append` method would get compiled into the class.  The one taking `char[]` would NPE.

stephan-herrmann:
> @stephan-herrmann : could you enlighten please why the compiler do that above ...

Yea, that's one of the reasons, why we all love overloading. The "simple" answer is in https://docs.oracle.com/javase/specs/jls/se19/html/jls-15.html#jls-15.12.2.5

`char[]` is a subtype of `Object`, hence `valueOf(char[])` is "better" ("more specific") than `valueOf(Object)`.

> ... without warning?

What should it say?
1. Warning: Your program uses overloading.
2. Warning: This overload resolution involves several applicable methods.
3. Warning: JLS 15.12.2.5 was needed to resolve this invocation

BTW, *if* ever you need to pass `null` to an overloaded method, perhaps always casting that argument to the desired type might be a reasonable practice.

laeubi:
> What should it say?

As we have talked about `@Null` annotations, I wonder if JDT could detect that this is always a NPE?

stephan-herrmann:
> As we have talked about @Null annotations, I wonder if JDT could detect that this is always a NPE?

Sure, methods like `String.valueOf(char[])` (and ctor `String(char[])` for that matter) just need to be properly specified as `String.valueOf(char @NonNull [])`, and then ecj will complain if that analysis is enabled.

Bonus: even though you probably don't want to wait for Oracle to make that change, you can do this yourself already today using external annotations.

laeubi:
@stephan-herrmann I more think about jdt includes these for known java functions.

stephan-herrmann:
> @stephan-herrmann I more think about jdt includes these for known java functions.

we already made a shy start in that direction, but it's a bottomless pit. Where to draw the line? Make ecj know each and every method in JDK? And also in all "well konwn" Java libraries? Are you volunteering? ;-P

No, external annotations (ideally shared, e.g., via lastnpe.org) are the unbeatable solution.

laeubi:
I think things like `String` are basic enough and given that platform was affected by this it seems valuable to save us from such pitfalls ...

iloveeclipse:
> What should it say?
> 
> 1. Warning: Your program uses overloading.

Not necessarily **that** way :-), but something like for the case below:
```
List.of(null);

ECJ:
> Type null of the last argument to method of(Object...) doesn't exactly match the vararg parameter type. 
> Cast to Object[] to confirm the non-varargs invocation, or pass individual arguments of type Object for a varargs invocation.

Javac:
> cast to Object for a varargs call
> cast to Object[] for a non-varargs call and to suppress this warning
```

PR #223, opened 2023-02-28
Provide IPath.toPath() similar to IPath.toFile()

guw:
Although it's possible to do `IPath.toFile().toPath()` this PR adds a convenience method which avoids the additional `File` object creation.

### DIFF
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
@@ -1048,26 +1048,26 @@ public class PathTest extends CoreTest {
 		//Case 1, absolute path with no trailing separator
 		IPath anyPath = new Path("/first/second/third");
 
-		assertNotNull("1.0", anyPath.toPath());
-		assertEquals("1.1", java.nio.file.Path.of("/first/second/third"), anyPath.toPath());
+		assertNotNull(anyPath.toPath());
+		assertEquals(java.nio.file.Path.of("/first/second/third"), anyPath.toPath());
 
 		// Case 2, absolute path with trailing separator
 		anyPath = new Path("/first/second/third/");
 
-		assertNotNull("2.0", anyPath.toPath());
-		assertEquals("2.1", java.nio.file.Path.of("/first/second/third/"), anyPath.toPath());
+		assertNotNull(anyPath.toPath());
+		assertEquals(java.nio.file.Path.of("/first/second/third/"), anyPath.toPath());
 
 		// Case 3, relative path with no trailing separator
 		anyPath = new Path("first/second/third");
 
-		assertNotNull("3.0", anyPath.toPath());
-		assertEquals("3.1", java.nio.file.Path.of("first/second/third"), anyPath.toPath());
+		assertNotNull(anyPath.toPath());
+		assertEquals(java.nio.file.Path.of("first/second/third"), anyPath.toPath());
 
 		// Case 4, relative path with trailing separator
 		anyPath = new Path("first/second/third/");
 
-		assertNotNull("4.0", anyPath.toPath());
-		assertEquals("4.1", java.nio.file.Path.of("first/second/third/"), anyPath.toPath());
+		assertNotNull(anyPath.toPath());
+		assertEquals(java.nio.file.Path.of("first/second/third/"), anyPath.toPath());
 
 	}
 }
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
@@ -1042,4 +1042,32 @@ public class PathTest extends CoreTest {
 		anyPath = new Path("//one/two/three/");
 		assertEquals("5.4", new Path("//"), anyPath.uptoSegment(0));
 	}
+
+	public void testToPath() {
+
+		//Case 1, absolute path with no trailing separator
+		IPath anyPath = new Path("/first/second/third");
+
+		assertNotNull("1.0", anyPath.toPath());
+		assertEquals("1.1", java.nio.file.Path.of("/first/second/third"), anyPath.toPath());
+
+		// Case 2, absolute path with trailing separator
+		anyPath = new Path("/first/second/third/");
+
+		assertNotNull("2.0", anyPath.toPath());
+		assertEquals("2.1", java.nio.file.Path.of("/first/second/third/"), anyPath.toPath());
+
+		// Case 3, relative path with no trailing separator
+		anyPath = new Path("first/second/third");
+
+		assertNotNull("3.0", anyPath.toPath());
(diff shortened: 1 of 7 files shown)


bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
```diff
@@ -1116,6 +1116,14 @@ public IPath setDevice(String value) {
 	public File toFile() {
 		return new File(toOSString());
 	}
+	
+	/* (Intentionally not included in javadoc)
+	 * @see IPath#toPath()
+	 */
+	@Override
+	public java.nio.file.Path toPath() {
+		return java.nio.file.Path.of(toOSString());
```
merks, line 1125: Why not put this implementation directly in IPath.toPath() where org.eclipse.core.runtime.IPath.toOSString() is available?

guw: Oh. I was blindly following the IPath -> Path pattern. Do we still need the implementation in Path then?

merks: A single implementation via the default method should be sufficient.  😀

laeubi: As IPath is marked as noimplement and noextend I would just put it into the (single) implementation.

HannesWell: In that case I think either way is fine and there is no great difference.

HannesWell (approved):
That is a quite handy addition, I thought about as well in the past.

iloveeclipse:
Please add trivial test for the new API.

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
```diff
@@ -1042,4 +1042,32 @@ public void testUptoSegment() {
 		anyPath = new Path("//one/two/three/");
 		assertEquals("5.4", new Path("//"), anyPath.uptoSegment(0));
 	}
+
+	public void testToPath() {
+
+		//Case 1, absolute path with no trailing separator
+		IPath anyPath = new Path("/first/second/third");
+
+		assertNotNull("1.0", anyPath.toPath());
```
HannesWell, line 1051: IMHO those marker messages are not necessary. In case the assertion fails the line information provided by the stack trace is sufficient to identify the failing line. Such generic messages are often confusing. Only verbose messages add real value, but in most cases the line-info from the stack trace is sufficient with the code available.

HannesWell (commented):
@guw can you please adjust the assertion messages and bump the version of the test plugin.

Then this is for me ready to go.

And since it is somehow the complement to this change, some of you might be interested in PR https://github.com/eclipse-equinox/equinox/pull/228.

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
guw: I share the same but I was following existing style in this file. I'd rather like to avoid introducing an inconsistency.

iloveeclipse: > I'd rather like to avoid introducing an inconsistency.

In this case I prefer to have inconsistency. 

There is no one left in the project who understands the reasoning behind such "very special" assertion messages.

laeubi: Please don't write bad tests for "consistency" and choose e meaningful message (e.g "absoloute IPath must return a non null java Path)

guw: @HannesWell What's the protocol for test bundle version update? Should it be `3.15.600.qualifier` or `3.16.0.qualifier`?

iloveeclipse: See https://wiki.eclipse.org/Version_Numbering#When_to_change_the_minor_segment

iloveeclipse: Btw, here are related links I wish every committer should have in bookmarks:

- https://wiki.eclipse.org/Version_Numbering 
- https://wiki.eclipse.org/Evolving_Java-based_APIs
- https://wiki.eclipse.org/Evolving_Java-based_APIs_2
- https://wiki.eclipse.org/Evolving_Java-based_APIs_3

laeubi: You can add them for example here:
- https://github.com/eclipse-equinox/equinox/blob/master/CONTRIBUTING.md

HannesWell: > > I'd rather like to avoid introducing an inconsistency.
> 
> In this case I prefer to have inconsistency.

Fully agree.



> You can add them for example here:
> 
>     * https://github.com/eclipse-equinox/equinox/blob/master/CONTRIBUTING.md

I think it would be better to add it to one umbrella repo, from which all others can reference so that we only have to maintain one file. Maybe in https://github.com/eclipse-platform/.github?

HannesWell:
@guw can you also have to bump the version of the following features.
- org.eclipse.equinox.core.feature
- org.eclipse.equinox.server.core

Just like the contained plugin the probably should also have a minor version bump.

HannesWell:
Looks like the `org.eclipse.equinox.core.sdk` feature is missing. But since all other modules succeeded, I think that should be the last one. :)

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
vogella: > Btw, here are related links I wish every committer should have in bookmarks:
> 
> * https://wiki.eclipse.org/Version_Numbering
> * https://wiki.eclipse.org/Evolving_Java-based_APIs
> * https://wiki.eclipse.org/Evolving_Java-based_APIs_2
> * https://wiki.eclipse.org/Evolving_Java-based_APIs_3

Is this wiki going away soon?

HannesWell:
> Looks like the `org.eclipse.equinox.core.sdk` feature is missing. But since all other modules succeeded, I think that should be the last one. :)

@guw do you want to perform that version bump so that we can squash and merge this PR?

guw:
Sorry, I am running a bit short on time right now.

HannesWell:
> Closing. Feel free to reopen if anyone is interested in picking it up.

I added your commits to https://github.com/eclipse-equinox/equinox/pull/228, where I want to improve the interoperability of IPath and Java's File/Path API in a broader scope.
Everybody interested is invited to join the discussion.

PR #228, opened 2023-03-05
Improve interoperabilitiy of IPath and Java's io.File/nio.Path

HannesWell:
This PR adds a convenient factory method to obtain an IPath from a `java.io.File` or `java.nio.file.Path`.

It is already possible to convert an IPath to a io.File and with PR #223 another method to convert to nio.Path is about to be added, but the other way around is currently missing and I usually to something like the following:
```
Path.fromOSString(file/path.toString())
```
The current implementation mimics that, but if necessary or beneficial the segments and device could be queried from the nio.path directly to avoid parsing a string.

In general I wonder if it would be better to add those methods in the IPath interface? And maybe move the other factories as well, at least those that return an IPath.

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -80,34 +80,47 @@ public final class Path implements IPath, Cloneable {
 	/** flags indicating separators (has leading, is UNC, has trailing, is for Windows) */
 	private final byte flags;
 
-	/** 
-	 * Constructs a new path from the given string path.
-	 * The string path must represent a valid file system path
-	 * on the local file system. 
-	 * The path is canonicalized and double slashes are removed
-	 * except at the beginning. (to handle UNC paths). All forward
-	 * slashes ('/') are treated as segment delimiters, and any
-	 * segment and device delimiters for the local file system are
-	 * also respected.
+	/**
+	 * Constructs a new path from the given string path. The string path must
+	 * represent a valid file system path on the local file system. The path is
+	 * canonicalized and double slashes are removed except at the beginning. (to
+	 * handle UNC paths). All forward slashes ('/') are treated as segment
+	 * delimiters, and any segment and device delimiters for the local file system
+	 * are also respected.
+	 * <p>
+	 * Instead of calling this method it is recommended to call
+	 * {@link IPath#fromOSString(String)} instead.
+	 * </p>
 	 *
-	 * @param pathString the portable string path
+	 * @param pathString the operating-system specific string path
+	 * @return the IPath representing the given OS specific string path
 	 * @see IPath#toPortableString()
+	 * @see IPath#fromOSString(String)
 	 * @since 3.1
 	 */
 	public static IPath fromOSString(String pathString) {
-		return new Path(pathString);
+		return IPath.fromOSString(pathString);
 	}
 
-	/** 
-	 * Constructs a new path from the given path string.
-	 * The path string must have been produced by a previous
-	 * call to <code>IPath.toPortableString</code>.
+	/**
+	 * Constructs a new path from the given path string. The path string must have
+	 * been produced by a previous call to <code>IPath.toPortableString</code>.
+	 * <p>
+	 * Instead of calling this method it is recommended to call
+	 * {@link IPath#fromPortableString(String)} instead.
+	 * </p>
 	 *
 	 * @param pathString the portable path string
+	 * @return the IPath representing the given portable string path
 	 * @see IPath#toPortableString()
+	 * @see IPath#fromPortableString(String)
 	 * @since 3.1
 	 */
 	public static IPath fromPortableString(String pathString) {
+		return IPath.fromPortableString(pathString);
+	}
+
+	static IPath parsePortableString(String pathString) {
 		int firstMatch = pathString.indexOf(DEVICE_SEPARATOR) + 1;
 		//no extra work required if no device characters
 		if (firstMatch <= 0)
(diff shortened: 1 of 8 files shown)


HannesWell:
This PR now also contains the changes from the abandoned PR https://github.com/eclipse-equinox/equinox/pull/223


> In general I wonder if it would be better to add those methods in the IPath interface? And maybe move the other factories as well, at least those that return an IPath.

Furthermore I now also integrated that part and added the new `fromFile/fromPath()` factory methods to the IPath interface.
What do the others think about this move? 
The most debatable part is probably the deprecation of the factories in the Path class. Are the others fine with that, do you think we should keep them forever or keep but deprecate them an remove them after an extended deprecation period?
One reason why I would like to have the factories in IPath is that, if one uses java.nio.Path the simple class-name collides with eclipse' Path class.
In general I wonder why there is an interface IPath (that clients should not implement) with only one implementation Path and both are part of the public API? IMHO that just makes thinks more complicated, but I guess it is like this for historic reasons?
Today we would probably just use the nio.Path API and provide a custom nio.FileSystem in o.e.core.resources for the workspace/platform resources.

HannesWell:
@merks @laeubi @iloveeclipse @vogella what do you think about this suggestion?

iloveeclipse:
> what do you think about this suggestion?

Which exactly? There were many questions asked, but I don't know which is supposed to be a suggestion?

HannesWell:
> > what do you think about this suggestion?
> 
> Which exactly? There were many questions asked, but I don't know which is supposed to be a suggestion? 

This change as a whole, what goes beyond #223.
Especially moving the IPath factory to the IPath Interface and deprecating the old methods in Path.
Of course also the Addition of new factory methods from File/nio.Path.

vogella:
LGTM

vogella:
I'm currently re-writting some internal code to Java 17 for which these methods would be really nice. 

@HannesWell is this one ready from your side?

HannesWell:
Glad you like the idea.
If everybody is fine with deprecating and removing the old methods after two years I would like to make the new methods completely independent of the deprecated ones so that when the derecation period is over, then can just be deleted.

And I should add some tests for the new API (probably just replace the old by the new API) and should make the new methods.

But I would first like to reach consensus on the new API and deprecation of the old methods.

vogella:
> Glad you like the idea. If everybody is fine with deprecating and removing the old methods after two years I would like to make the new methods completely independent of the deprecated ones so that when the derecation period is over, then can just be deleted.

Sounds good.
 
> And I should add some tests for the new API (probably just replace the old by the new API) and should make the new methods.

Sounds also great

HannesWell:
> > Glad you like the idea. If everybody is fine with deprecating and removing the old methods after two years I would like to make the new methods completely independent of the deprecated ones so that when the derecation period is over, then can just be deleted.
> 
> Sounds good.
> 
> > And I should add some tests for the new API (probably just replace the old by the new API) and should make the new methods.
> 
> Sounds also great

Did both now so for me this is ready, time for reviews. :)

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
```diff
@@ -90,10 +90,13 @@ public final class Path implements IPath, Cloneable {
...
 	 * also respected.
 	 *
-	 * @param pathString the portable string path
+	 * @param pathString the operating-system specific string path
+	 * @return the IPath representing the given OS specific string path
 	 * @see IPath#toPortableString()
 	 * @since 3.1
+	 * @deprecated use {@link IPath#fromOSString(String)} instead
 	 */
+	@Deprecated(since = "3.18", forRemoval = true)
```
iloveeclipse, line 99: I'm not happy about `forRemoval = true` here and below.
I see ~250 references in my SDK workspace and ~50 references in our product.

While the move from Path to IPath might be OK, but deleting this here seem to be not really justified, considering amount of existing code that uses the old API.

iloveeclipse (changes requested):
In general, sounds good.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
merks: Yes, I dread dealing with such things.  All my projects want to work on older versions of Eclipse and all my projects want to be warning free tomorrow, so I will need to create utilities to isolate such things if there really is a threat of removal.  I am not likely to be the only one with such a concern and need.

vogella:
@HannesWell I suggest to remove the forRemoval for now and only deprecate the method. This should help users to migrate to the new alternative.

mickaelistria:
> I suggest to remove the forRemoval for now and only deprecate the method. This should help users to migrate to the new alternative.

Even deprecating is IMO a bad idea as long as there are consumers of this method in Platform/SDK code base. Annotating with a `@seealso` and documenting clearly that another method can and should be used instead would be good enough for deprecation candidates, and then we should work on making the Platform/SDK code not using the deprecated method at all, and only then, we can fairly and relatively safely deprecate.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
tjwatson: I agree with others.  We should not mark this for removal at this time.

HannesWell:
I'm fine with removing the `forRemoval` for now, but would be in favor to eventually remove the deprecated methods after an extended deprecation-period. (how long should that be? Five years, ten years?). Otherwise we have another zombie like `SubProgressMonitor`.



> > I suggest to remove the forRemoval for now and only deprecate the method. This should help users to migrate to the new alternative.
> 
> Even deprecating is IMO a bad idea as long as there are consumers of this method in Platform/SDK code base. Annotating with a `@seealso` and documenting clearly that another method can and should be used instead would be good enough for deprecation candidates, and then we should work on making the Platform/SDK code not using the deprecated method at all, and only then, we can fairly and relatively safely deprecate.

In general I agree, but in this case I'm relatively confident that the migration should be simple (some tedious work, but not complicated). Therefore I would like to save the extra step and deprecate it immediately.
If, contrary to expectations, problems should arise during migration we can still un-deprecate the methods.


> My only concern is that fromPath/toPath doesn't make it clear to me it's about Java Path not equinox one but as I can't think of better I think it's fine. E.g. I thought about from[Java|JVM]Path/to[Java|JVM]Path or some similar more specific one but it looks worse to me.

Yes I had the same concern and came to the same conclusion. Another alternative would be to use `fromNIOPath()/toNIOPath()` but that also seemed strange to me. Therefore I just stuck with `fromPath/toPath()` and with `fromFile/toFile()` in mind I think that is consistent after all.

mickaelistria:
> In general I agree, but in this case I'm relatively confident that the migration should be simple (some tedious work, but not complicated). Therefore I would like to save the extra step and deprecate it immediately.

If there is no-one willing to do the tedious work before deprecating, then it's unlikely someone do it later so the deprecated methods will forever be referenced and will never be removed.
On the contrary, if the tedious work is done right now, it gives the opportunity to deprecate without introducing new warnings in Platform and give a good example of how other can mitigate the deprecation.

Removing internal references to deprecated method is not a technical requirement, it's more an organizational one: the fact that we don't want to remove the existing references right now is maybe a sign that it's not worth deprecating the method at all...

HannesWell:
> > In general I agree, but in this case I'm relatively confident that the migration should be simple (some tedious work, but not complicated). Therefore I would like to save the extra step and deprecate it immediately.
> 
> If there is no-one willing to do the tedious work before deprecating, then it's unlikely someone do it later so the deprecated methods will forever be referenced and will never be removed. On the contrary, if the tedious work is done right now, it gives the opportunity to deprecate without introducing new warnings in Platform and give a good example of how other can mitigate the deprecation.

I believe you miss understood me. I never said that I'm not willing to do the tedious work (nevertheless help would be highly appreciated). When I said I want to save extra work I was referring to saving extra work, I meant the split about introducing a new method and deprecating the old one. Of course one can not migrate before the new method is available.

In order to come to a consensus, I now went for the clean way and just added a note that it is recommended to use the IPath counterparts. Everything else can be discussed after the migration.

@iloveeclipse, @merks do you want to add your reviews?

merks:
If all the methods you'd like to deprecate in favor of the corresponding new IPath methods were actually already implemented to delegate to the IPath method, would one be able to "fix" any specific caller simply by in-lining (Alt-Shift-I) the call?  After all, if you were eventually to delete them, the implementations would have to be elsewhere.  Just asking...

HannesWell:
> If all the methods you'd like to deprecate in favor of the corresponding new IPath methods were actually already implemented to delegate to the IPath method, would one be able to "fix" any specific caller simply by in-lining (Alt-Shift-I) the call? After all, if you were eventually to delete them, the implementations would have to be elsewhere.

Good suggestion. Implemented that.

iloveeclipse (approved):
LGTM.

mickaelistria (approved):
Better methods, better behavior, no deprecation left to debate, so good for me as well ;)

vogella:
Very nice, thanks @HannesWell

HannesWell:
Created a small follow up, to also add an `EMPTY` and `ROOT` constant to IPath: 
https://github.com/eclipse-equinox/equinox/pull/259

PR #231, opened 2023-03-07
Switch build JVM to Temurin

akurtakov:
So we get latest 17.x


akurtakov:
Version bump issue to be further worked on once we get first I build.

Issue #236, opened 2023-03-21
HelpWanted | NativeCodeFinder only finds libraries from the host bundle


krooq:
Hi,

I'm trying to load a library from a fragment using the Bundle-NativeCode header.
The library fails to load, when I debug all the way through I find that the `NativeCodeFinder` seems to only return library paths from the host bundle in its `findBundleNativeCode` method.
My guess is that the `nativeCode.get(0)` sections in the method are just a typo or something that should instead refer to the moduleWire loop variable but that's just a guess.

This is all I can discern and tbh I don't really know if I'm even doing it right.  
It seems odd that this feature would be broken since a main use case for fragments is this loading of different native libs.. 
There doesn't seem to be any documentation on how to configure the Bundle-NativeCode hearder differently for a fragment compared to a regular plugin so I just assume its the same but there are only 2 SO articles kind of related to this problem and they don't seem to provide concrete solutions.

PR #238, opened 2023-03-23
Fix native code issue

tjwatson:
Fixes issue #236

### DIFF
--- a/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Core OSGi Tests
 Bundle-SymbolicName: org.eclipse.osgi.tests;singleton:=true
-Bundle-Version: 3.16.1000.qualifier
+Bundle-Version: 3.18.400.qualifier
 Bundle-Vendor: Eclipse.org
 Require-Bundle: 
  org.eclipse.core.runtime,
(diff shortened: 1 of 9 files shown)


tjwatson:
@laeubi do you understand why the `Build Equinox / build (pull_request)` build is failing to build `org.eclipse.osgi.tests` with errors like this:
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.0-SNAPSHOT:compile (default-compile) on project org.eclipse.osgi.tests: Compilation failure: Compilation failure: 
Error:  /home/runner/work/equinox/equinox/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/appadmin/FailedApp.java:[16] 
Error:  	import org.eclipse.equinox.app.IApplication;
Error:  	       ^^^^^^^^^^^^^^^^^^^^^^^
Error:  The import org.eclipse.equinox.app cannot be resolved
```
The jenkins build does not fail the same way.

tjwatson:
@krooq This PR has been merged.  There should be a nightly build done this evening with the fix.  It would be good if you could test out the fix in your environment.

krooq:
Thanks so much for jumping on this, and so quickly too!
Looks like the build is still failing.
Also I'm not really sure how to test this in my environment as it's a transitive dependency of something, it might be a little hard to for me to figure out how to get it to resolve the source version.. but I'll give it a go.

tjwatson:
> Thanks so much for jumping on this, and so quickly too! Looks like the build is still failing. Also I'm not really sure how to test this in my environment as it's a transitive dependency of something, it might be a little hard to for me to figure out how to get it to resolve the source version.. but I'll give it a go.

https://download.eclipse.org/eclipse/downloads/drops4/I20230323-1800/ has the fix if you can download and use that Eclipse SDK with your set of bundles/plugins.

If that is difficult to do I think you could just overwrite your installation of `org.eclipse.osgi` located in the plugins folder with the version from this build and then run with -clean.

PR #247, opened 2023-04-13
Define github action to build Windows launcher

Torbjorn-Svensson:
Currently, only equinox developers are allowed to launch the build jobs to produce the launcher artifacts.
As this greatly inhibits the community in helping develop these launchers, I propose to use github actions to provide the launcher artifact for at least Windows as the alternative is for the contributor to have a system available with the right version of Visual Studio and perhaps even buy a Visual Studio license just to help fix a bug.

I've been using this github action to validate the suggested change in #118 and it works.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,6 +50,40 @@ jobs:
         name: test-results
         if-no-files-found: error
         path: '**/target/*-reports/*.xml'
+  build-launcher-win64:
+    runs-on: windows-2019
+    env:
+      PROGRAM_OUTPUT: eclipse.exe
+      PROGRAM_LIBRARY: eclipse.dll
+      DEFAULT_OS: win32
+      DEFAULT_OS_ARCH: x86_64
+      DEFAULT_WS: win32
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 1
+    - name: Set up JDK 8
+      uses: actions/setup-java@v3
+      with:
+        java-version: '8'
+        distribution: 'temurin'
+    - name: Visual Studio shell
+      uses: egor-tensin/vs-shell@v2
+      with:
+        arch: x64
+    - name: Build
+      working-directory: features/org.eclipse.equinox.executable.feature/library/win32
+      run: nmake -f make_win64.mak
+      shell: pwsh
+    - name: Upload artifacts
+      uses: actions/upload-artifact@v3
+      if: success()
+      with:
+        name: Windows launcher artifacts
+        path: |
+          features/org.eclipse.equinox.executable.feature/library/win32/eclipse*.exe
+          features/org.eclipse.equinox.executable.feature/library/win32/eclipse*.dll
+        if-no-files-found: error
   tck:
     runs-on: ubuntu-latest
     steps:


vogella:
Thanks, such a build action sounds really helpful.

.github/workflows/build.yml
```diff
@@ -50,6 +50,40 @@ jobs:
...
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 1
+    - name: Set up JDK 8
+      uses: actions/setup-java@v3
+      with:
+        java-version: '8'
+        distribution: 'temurin'
+    - name: Visual Studio shell
+      uses: egor-tensin/vs-shell@v2
```
HannesWell, line 71: Would the [microsoft/setup-msbuild](https://github.com/microsoft/setup-msbuild) action work as well?
Since it is provided by Microsoft directly I would expect/hope that it is better supported.

.github/workflows/build.yml
```diff
@@ -50,6 +50,40 @@ jobs:
         name: test-results
         if-no-files-found: error
         path: '**/target/*-reports/*.xml'
+  build-launcher-win64:
+    runs-on: windows-2019
```
HannesWell, line 54: What about `windows-latest`?

HannesWell (commented):
Thanks for that, that's definitely helpful.


Wouldn't it make sense to build the launcher for other platforms as well?
Or if they have changed in the Jenkins file, just like it is done in SWT now?
https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/Jenkinsfile

.github/workflows/build.yml
Torbjorn-Svensson: This was my first github action and I googled for something that defined Visual Studio 2019 (the version referred to in the build.bat file) and it was one of the first matches.
Feel free to try out if the other action works and if so, change it.

My sole goal with this setup was to make it possible to validate #118 without installing the required tools (including buying them if needed).

.github/workflows/build.yml
Torbjorn-Svensson: Again, I found something that indicated that windows-2019 was the one including Visual Studio 2019. Maybe windows-latest contains the same version of Visual Studio or a version that would be equally good..

Torbjorn-Svensson:
> Wouldn't it make sense to build the launcher for other platforms as well? Or if they have changed in the Jenkins file, just like it is done in SWT now? https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/Jenkinsfile

Yes probably. I did not want to put too much effort into this before it was reviewed and approved as something that would be useful.
From what I can see, the link you point to only covers a subset of the targets that the launcher supports. 

The launcher build files mentions:
Linux: x86_64, ppc64le, s390x, aarch64, loongarch64
macos: x86_64, aarch64
Windows: x86_64

At the same time, if I look at [quinox.binaries](https://github.com/eclipse-equinox/equinox.binaries), then the s390x and loongarch64 Linux targets appears to be have been dropped.

Is the job that [Jenkinsfile](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/Jenkinsfile) define executed on every new push/PR that touches SWT sources? Or is this another place where there are only people from the Equinox project that can trigger the build and effectively verify that a change is doing what it's supposed to?

.github/workflows/build.yml
HannesWell: No problem :)
Then lets see what happens in https://github.com/eclipse-equinox/equinox/pull/250.

HannesWell:
> > Wouldn't it make sense to build the launcher for other platforms as well? Or if they have changed in the Jenkins file, just like it is done in SWT now? https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/Jenkinsfile
> 
> Yes probably. I did not want to put too much effort into this before it was reviewed and approved as something that would be useful. 

That's totally fine. My remark was mainly about how this could be improved further.


> From what I can see, the link you point to only covers a subset of the targets that the launcher supports.
> 
> The launcher build files mentions: 
> Linux: x86_64, ppc64le, s390x, aarch64, loongarch64 
> macos: x86_64, aarch64 
> Windows: x86_64
> 
> At the same time, if I look at [quinox.binaries](https://github.com/eclipse-equinox/equinox.binaries), then the s390x and loongarch64 Linux targets appears to be have been dropped.

If you neglect s390x and loongarch64, then the linked Jenkins pipeline to build SWT and its binaries runs on the same platforms.
That's the full list:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/6912a16abe78c905180908ea4d59218f6519c272/Jenkinsfile#L107

> 
> Is the job that [Jenkinsfile](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/Jenkinsfile) define executed on every new push/PR that touches SWT sources? Or is this another place where there are only people from the Equinox project that can trigger the build and effectively verify that a change is doing what it's supposed to?

This job runs on every push or PR, if something has changed in the native c code for the binary, but there are a lot of (ant) scripts involved to achieve that. I'm currently working to simplify that, but it is a greater undertaking.
The only thing it is not doing is to archive the build binaries. That would be trivial to change, but then the binaries should only be signed on master branch pushes.

Torbjorn-Svensson:
> This job runs on every push or PR, if something has changed in the native c code for the binary, but there are a lot of (ant) scripts involved to achieve that. I'm currently working to simplify that, but it is a greater undertaking. The only thing it is not doing is to archive the build binaries. That would be trivial to change, but then the binaries should only be signed on master branch pushes.

Then I believe that this would be the preferred route to go.
The goal I had with the github action was to simplify the testing phase as it would not require everyone to have every platform and tools for that platform in order to make a simple change like was done in #117. 

Obviously, the signing should only be done on merge, otherwise there would potentially be viruses out there with the Eclipse Foundation signature and that would be really bad! I do not know much about Jenkins pipeline (I'm not an CI guy..), but I suppose there would be some way to figure out if it's a PR build or a push and based on that, either sign the binary (and commit it to the other git repository) or simply just attach it somewhere for the user to download and verify that it works as expected.

Based on what you said above, the work on the launchers should probably be on hold until the simplification of the SWT binaries are complete in order to not do the same job twice.

Torbjorn-Svensson:
Maybe the Windows launcher could also be cross built on a Linux system to reduce the extra headache for contributors that does not have all the proper tools on a Windows computer, but still have a Windows system that they could use for testing. I know that all the Windows libraries that CDT builds are cross built, even the macos libraries are cross built.
If you were to use the CDT approach, then at least 4 platforms would be covered in a single Docker container and the contributor would just need that Docker container in order to build the launcher locally. For the non x86_64 Linux targets, I guess cross compilation could be an option, but it would require the GTK library (and other 3rd party libraries) to be available for linking. In the CDT case, we do not have any such 3rd party library, so we can easily just cross compile the library directly.

The main positive side with a single Docker container would be the reduced risk that one of the required platforms would be unavailable when needed for a build.

laeubi:
I really wonder if it makes sense to blindly submit code and then download the executable to "test" it... I think if one wants to contribute to the launcher, a local development environment is a requirement and if that's to complicated we maybe should better describe how to setup one.

PR #250, opened 2023-04-15
Use 'microsoft/setup-msbuild' in windows launcher building workflow

HannesWell:
and run on 'windows-latest'


laeubi:
Can we please move building the launcher into a separate action? It seems not very useful to mix these both, I would even suggest to skip the build if no launcher code is touched by a PR...

HannesWell:
Closing this since it seems not to work as intended.

> Can we please move building the launcher into a separate action? It seems not very useful to mix these both, I would even suggest to skip the build if no launcher code is touched by a PR...

Agree on all. But this is something for another PR.

PR #259, opened 2023-05-20
Add IPath constants EMPTY and ROOT to IPath

HannesWell:
As a follow-up to https://github.com/eclipse-equinox/equinox/pull/228, add `IPath` constants `EMPTY` and `ROOT` to `IPath` interface and recommend to use them instead of `Path.EMPTY` or `Path.ROOT`.

When starting the migration to the new factory methods from https://github.com/eclipse-equinox/equinox/pull/228 I noticed that those constants are missing. With this at least PDE can become completely free of `o.e.core.runtime.Path`.

It would be good to have that in the upcoming release too, but since we are late in the development cycle, this needs Project-Lead or PMC approval.
@vogella or @tjwatson can you please be so kind.

### DIFF
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
@@ -50,8 +50,8 @@ public class PathTest extends CoreTest {
 		assertEquals("1.1", with, without.addTrailingSeparator());
 		assertEquals("1.2", without, without.addTrailingSeparator());
 
-		assertSame("2.0", Path.ROOT, Path.ROOT.addTrailingSeparator());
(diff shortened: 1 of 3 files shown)


HannesWell:
@akurtakov, @vogella, @tjwatson can you please cast your vote?
I think it would be good to have this in for RC1.

mickaelistria:
IMO, such non-critical API addition doesn't qualify as important enough to urge it into a RC. But @tjwatson is the only project lead and so is the one who has the final say.

merks:
It's non-critical indeed, but it does complete a new feature added for 4.28, and it looks to be zero risk...

vogella:
> I'm fine with this change for RC1 as I don't see any risk in it.

Same here, go from my PMC point of view. @HannesWell feel free to merge once master is open unless another PMC member cast a negative vote

PR #262, opened 2023-06-06
use try-with-resource

jukzi:


tjwatson (commented):
I've only looked at the `org.eclipse.osgi` changes so far.  I'll try to spend some time on the others.  But would be good if others could review because I am currently low on time I can focus here.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/internal/boot/PlatformURLConnection.java
```diff
@@ -164,39 +163,31 @@ private void copyToCache() throws IOException {
...
 
 			tgtFile = new File(tgt);
-			tgtos = new FileOutputStream(tgtFile);
-
-			while (count != -1) {
-				total += count;
-				tgtos.write(buf, 0, count);
-				count = srcis.read(buf);
+			try (FileOutputStream tgtos = new FileOutputStream(tgtFile)){
+				while (count != -1) {
```
HannesWell, line 182: Can't we use Files.copy() here?
It accepts different combinations of streams and Paths.

HannesWell (commented):
I appricate this change and can have a look as well next week, when back from vacation.

jukzi:
Can equinox use BREE 17?
` The method transferTo(Writer) is undefined for the type Reader`

jukzi:
The bundle that shows an error is "org.eclipse.equinox.common".  Reader.transferTo(Writer) is defined in 9 and that bundle is already BREE 11. So it's an build error, that 8 is used to build! That can have negative consequences if removed methods are used.

iloveeclipse:
> The bundle that shows an error is "org.eclipse.equinox.common"

Are you sure?
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.0-SNAPSHOT:compile (default-compile) on project org.eclipse.osgi: Compilation failure: Compilation failure: 
[ERROR] /home/jenkins/agent/workspace/equinox_PR-262/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/log/EquinoxLogWriter.java:[375] 
[ERROR] 	fileIn.transferTo(this.writer);
[ERROR] 	       ^^^^^^^^^^
[ERROR] The method transferTo(Writer) is undefined for the type Reader
```

jukzi:
oh thanks, i used transferTo twice. EquinoxLogWriter is "org.eclipse.osgi". I'l revert for that package.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
```diff
@@ -1778,37 +1762,51 @@ private Properties findVMProfile(Generation systemGeneration) {
...
-					// consider logging
-				} finally {
-					try {
-						profileIn.close();
-					} catch (IOException e) {
-						// nothing to do
+			try (InputStream profileIn = createPropStream(systemGeneration, propJavaProfile)){
+				if (profileIn != null) {
+					Properties result = new Properties();
+					try (BufferedInputStream inStream = new BufferedInputStream(profileIn)) {
```
HannesWell, line 1788: Same as above, another buffer is probably not necessary and `result.load(profileIn);` would probably be enough and even better.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
```diff
@@ -1778,37 +1762,51 @@ private Properties findVMProfile(Generation systemGeneration) {
...
+			String javaProfile = vmProfile + PROFILE_EXT;
+			profileIn = findInSystemBundle(systemGeneration, javaProfile);
+			if (profileIn == null)
+				profileIn = getNextBestProfile(systemGeneration, JAVASE, runtimeVersion, embeddedProfileName);
+		}
+		if (profileIn == null)
+			// the profile url is still null then use the min profile the framework can use
+			profileIn = findInSystemBundle(systemGeneration, "JavaSE-1.8.profile"); //$NON-NLS-1$
+		return profileIn;
+	}
```
HannesWell, line 1779: I think this is a good opportunity to add some braces :)
```suggestion
	private InputStream createProfileStream(Generation systemGeneration, String vmProfile, String embeddedProfileName) {
		InputStream profileIn = null;
		if (vmProfile != null) {
			// look for a profile in the system bundle based on the vm profile
			profileIn = findInSystemBundle(systemGeneration, vmProfile + PROFILE_EXT);
			if (profileIn == null) {
				profileIn = getNextBestProfile(systemGeneration, JAVASE, runtimeVersion, embeddedProfileName);
			}
		}
		if (profileIn == null) {
			// the profile url is still null then use the min profile the framework can use
			profileIn = findInSystemBundle(systemGeneration, "JavaSE-1.8.profile"); //$NON-NLS-1$
		}
		return profileIn;
	}
```

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
```diff
@@ -1740,29 +1740,13 @@ private Properties findVMProfile(Generation systemGeneration) {
...
-					profileIn = getNextBestProfile(systemGeneration, JAVASE, runtimeVersion, embeddedProfileName);
-			}
-			if (profileIn == null)
-				// the profile url is still null then use the min profile the framework can use
-				profileIn = findInSystemBundle(systemGeneration, "JavaSE-1.8.profile"); //$NON-NLS-1$
-			if (profileIn != null) {
-				try {
+			try (InputStream profileIn = createProfileStream(systemGeneration, vmProfile, embeddedProfileName)) {
+				if (profileIn != null) {
 					result.load(new BufferedInputStream(profileIn));
```
HannesWell, line 1745: I have the impression that Properties.load() already uses a buffer internally, therefore a BufferedInputStream would probably only add extra copying and GC work.
So result.load(profileIn); would probably be enough and even better.

HannesWell (changes requested):
Thanks for this PR (and your patience).
In general this looks good and this fixes a number of resource-leaks.

Just a few remarks for further improvements below.

Feel free to squash all code changes into one commit and the version bumps in another one.

jukzi:
i hope i incorporated all remarks.

HannesWell (approved):
> i hope i incorporated all remarks.

Looks good now. Thank you.

PR #264, opened 2023-06-06
Avoid wrapping primitives for toString()

jukzi:
see https://rules.sonarsource.com/java/RSPEC-2131

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
@@ -2142,9 +2142,9 @@ public class Storage {
 		//
 		// This bit of code attempts to do that by using the bundle ID as an ID for the temp dir along with an incrementing ID
 		// in cases where the temp dir may already exist.
-		Long bundleID = Long.valueOf(generation.getBundleInfo().getBundleId());
+		long bundleID = generation.getBundleInfo().getBundleId();
 		for (int i = 0; i < Integer.MAX_VALUE; i++) {
-			bundleTempDir = new File(libTempDir, bundleID.toString() + "_" + Integer.valueOf(i).toString()); //$NON-NLS-1$
+			bundleTempDir = new File(libTempDir, bundleID + "_" + i); //$NON-NLS-1$
 			libTempFile = new File(bundleTempDir, libName);
 			if (bundleTempDir.exists()) {
 				if (libTempFile.exists())
--- a/bundles/org.eclipse.equinox.app/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.app/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.app; singleton:=true
-Bundle-Version: 1.6.200.qualifier
+Bundle-Version: 1.6.300.qualifier
 Bundle-Vendor: %providerName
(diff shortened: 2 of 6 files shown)


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/Storage.java
```diff
@@ -2144,7 +2144,7 @@ public String copyToTempLibrary(Generation generation, String absolutePath) {
 		// in cases where the temp dir may already exist.
 		Long bundleID = Long.valueOf(generation.getBundleInfo().getBundleId());
 		for (int i = 0; i < Integer.MAX_VALUE; i++) {
-			bundleTempDir = new File(libTempDir, bundleID.toString() + "_" + Integer.valueOf(i).toString()); //$NON-NLS-1$
+			bundleTempDir = new File(libTempDir, bundleID.toString() + "_" + Integer.toString(i)); //$NON-NLS-1$
```
tjwatson, line 2147: Not that this matters much for this code block since it is expected that we will iterate very few times before finding an available directory name (mostly just once), but this change avoids the Integer.valueOf optimization for the small Integer cache.

If this is about simplifying the code then it would be more simple to avoid any method calls and just `+ i` onto the string concatination:

`bundleID.toString() + "_" + i`

mickaelistria:
Looks like a version bump is required for org.eclipse.osgi at least.

jukzi:
> Looks like a version bump is required for org.eclipse.osgi at least.

but i bumped that already today

mickaelistria:
> but i bumped that already today

OK, last build of this patch was 4 hours ago. Was the bump merged on master in the last 4 hours (so the build missed it)? If so, please try rebasing and it should be sufficient.

HannesWell (commented):
Looks good, this just needs version bumps. At the moment
- org.eclipse.equinox.app
- org.eclipse.equinox.security.ui

But more maybe follow.

mickaelistria:
@tjwatson we'd like to merge it ASAP. Are the "Build Equinox / ..." GitHub Actions and their failures related or can we merge despite those?

laeubi:
The DS test is sometimes flapping and the TCKs are known failures (@tjwatson is there any plan to make Equinox compliant or take a look why the TCKs are not executed correctly?). 
So I think its fine to merge, nerveless I wonder why this is that urgent it needs to be merge ASAP?!? This seem not target any bug/issue...

laeubi:
By the way it seems Jenkins now successfully again reports javadoc errors, just in case someone wants to address those instead of micro-optimizations.

grafik

> Things should always be closed ASAP ;)

I can install an github action that **CLOSES** things immediately :-D

mickaelistria:
>  I can install an github action that CLOSES things immediately :D

If it can do the technical, functional, style, compatibility... review reliably first (as I include this in pre-requisites to "possible" in ASAP), that would be perfect. Can you achieve such automation within a couple of weeks? :P

laeubi:
If we can agree that test-coverage is good and an a single approving review is enough one can enable branch protection to require:

1. an approving review
2. jenkins-ci succeeds
3. no freeze period

and then enable auto-merge:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request

tjwatson:
> The DS test is sometimes flapping and the TCKs are known failures (@tjwatson is there any plan to make Equinox compliant or take a look why the TCKs are not executed correctly?).

While I would like to get the TCKs to pass, I have no time to investigate this.  Someone else would need to look at that.

> @tjwatson we'd like to merge it ASAP. Are the "Build Equinox / ..." GitHub Actions and their failures related or can we merge despite those?

Should be fine to merge.

PR #271, opened 2023-06-12
Activator.urlTrackers: final for synchronized access

jukzi:


HannesWell (approved):
Looks reasonable to me.
@tjwatson what do you think?

tjwatson (approved):
Seems fine.

PR #273, opened 2023-06-14
Fix NPE when bundle is not started

laeubi:
Currently the code checks for 'localizationTracker == null' to see if bundle is started, but that field is actually never null (as it is a final field always initilized with a value) instead the Activator instance itself can be null.

This changes the method to replace the usless check with one that checks if the activator instance is null and additionally make the method static.


akurtakov:
Change looks good to me from quick look.

Issue #274, opened 2023-06-14
Module j9stubs not found, required by org.eclipse.osgi


fipro78:
I tried to use the latest Equinox OSGi bundle 3.18.400 to check my OSGi deployment setup:

The Atomos use case is not running anymore. If I try to start the application via

```
java --add-modules=ALL-MODULE-PATH,java.net.http -p bundles -m org.apache.felix.atomos org.osgi.framework.system.packages=""
```

I get the following error:

```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module j9stubs not found, required by org.eclipse.osgi
```

If I understand correctly, the module-info.class is now generated via bnd. But it somehow now adds the requirement for j9stubs, which is not found anywhere. I can see it here https://github.com/eclipse-equinox/equinox/tree/master/bundles/org.eclipse.osgi/osgi but I don't think it is part of the resulting org.eclipse.equinox.osgi bundle and therefore org.eclipse.osgi does not work with Atomos on the module-path anymore.

rotty3000:
Try adding `j8stubs` [here](https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi/bnd.bnd#L2) with ignore=true

tjwatson:
Should be fixed in the next I-Build.

fipro78:
Just downloaded the latest I-Build and copied the org.eclipse.osgi bundle into my example. Works again as expected.

PR #275, opened 2023-06-14
Ignore j9stubs in module-info generation

fipro78:


tjwatson:
Fixes #274

@fipro78 have you been able to confirm this fixes your issue?

fipro78:
@tjwatson
Can I download the build results of the PR build somewhere? Or do I need to build it locally?

fipro78:
I tried to build locally with my changes, but I get strange errors. Maybe because I am behind a corporate proxy. The error message is:

```
Failed to resolve target definition file:/C:/Users/xxx/.m2/repository/org/eclipse/eclipse-sdk-prereqs/4.29.0-SNAPSHOT/eclipse-sdk-prereqs-4.29.0-SNAPSHOT.target: Failed to load p2 metadata repository from location https://download.eclipse.org/tools/orbit/downloads/drops/R20201118194144/repository/: Unable to read repository at https://download.eclipse.org/tools/orbit/downloads/drops/R20201118194144/repository. ConnectException: UnresolvedAddressException
```

I tried to configure a local mirror in my settings.xml like this:

```
   <mirror>
	  <id>orbit202305-local</id>
	  <mirrorOf>https://download.eclipse.org/tools/orbit/downloads/drops/R20230531010532/repository</mirrorOf>
	  <url>file:///C:/Development/update-sites/orbit-202305/</url>
	  <layout>p2</layout>
	  <mirrorOfLayouts>p2</mirrorOfLayouts>
   </mirror>
```

But the mirror doesn't seem to be used. I checked the target file that is downloaded in the local maven repository, and there I see a trailing slash in the repository location. The [Tycho docs](https://wiki.eclipse.org/Tycho/Target_Platform/Authentication_and_Mirrors) say, there should be no trailing slash.

@laeubi Do you have any idea on this? You are the Tycho expert. :)

Sorry for hijacking this topic. This is of course not related to the issue. It just brings me back to my question if it is possible to download the build results of the CI build, so I can check if my changes solved the issue, as I am currently not able to build locally.

laeubi:
`UnresolvedAddressException` looks strange, nerveless mirroring should work that way, if I remember correctly Tycho should print a message like "Using mirror for ..." if the mirror is found, here is an example with multiple mirrors defined:

https://github.com/faktorips/faktorips.base/blob/main/settings.xml

if the settings.xml is not at default location you need to specify it with `-s` option to maven.

fipro78:
The settings.xml is at the default location.

The info message before the failure is the following:

```
[INFO] Loading repository '' from mirror 'orbit202305-local' at 'file:///C:/Development/update-sites/orbit-202305/'
[INFO] ### Using TychoRepositoryTransport for remote P2 access ###
[INFO]     Cache location:         C:\Users\fad8pl\.m2\repository\.cache\tycho
[INFO]     Transport mode:         online
[INFO]     Http Transport type:    Java11Client
[INFO]     Update mode:            cache first
[INFO]     Minimum cache duration: 60 minutes
[INFO]       (you can configure this with -Dtycho.p2.transport.min-cache-minutes=<desired minimum cache duration>)
[INFO] Downloading from p2: https://download.eclipse.org/tools/orbit/downloads/drops/R20201118194144/repository/p2.index
```

So something doesn't seem to work right. Where are the eclipse-sdk-prereqs come from? Maybe I could try to edit the .target file and remove the trailing slash from the location. IIRC that solved the issue in the past in my setups.

laeubi:
> So something doesn't seem to work right. Where are the eclipse-sdk-prereqs come from? Maybe I could try to edit the .target file and remove the trailing slash from the location. IIRC that solved the issue in the past in my setups.

The sources are here:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/eclipse.platform.releng.prereqs.sdk

please don't hesitate to open a bugreport at Tycho if its just a matter of stripping the slash we should fix that (there are already some mirror tests but seem they are not covering that at the moment) or suggest changes to that file.

fipro78:
@laeubi I created https://github.com/eclipse-tycho/tycho/issues/2533 to track the issue.

@tjwatson I will try to build at home from my private machine later, so I can verify the changes from a local build.

laeubi:
@fipro78 just to make sure, you need to bump the micro version of the bundle otherwise your changes will be replaced with the baseline!

fipro78:
@laeubi 
Thanks for the pointer. You are right, after I updated locally to 3.18.600 the resulting org.eclipse.osgi bundle works again on the module path.

Should I update the version? It is already 3.18.500. So wasn't it updated already?

laeubi:
It should be updated for each change in a release, I'm not aware of any updates in this release, but I think @tjwatson can tell better here.

tjwatson:
> It should be updated for each change in a release,

We have always only updated the micro-version once each release.

> I'm not aware of any updates in this release, but I think @tjwatson can tell better here.

The `org.eclipse.osgi` project version has already been updated with 044487a5a6e4457cd2ec9795a4a123ba6b9d0eea

tjwatson:
> Thanks for the pointer. You are right, after I updated locally to 3.18.600 the resulting org.eclipse.osgi bundle works again on the module path.

I'm not sure why you needed to do that unless it was comparing to the last time you built it locally and you already had the version bump.

I merged the fix.  Thanks for confirming it fixed your problem.

Issue #278, opened 2023-06-15
!! Tests DNF since I20230614-2220 


jukzi:
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.IPath.removeTrailingSeparator()" because "path" is null
at org.eclipse.core.internal.resources.Resource.<init>(Resource.java:49)
at org.eclipse.core.internal.resources.Container.<init>(Container.java:30)
at org.eclipse.core.internal.resources.WorkspaceRoot.<init>(WorkspaceRoot.java:41)

```
locally reproducable with org.eclipse.core.tests.resources.AutomatedResourceTests on WIN that 
org.eclipse.core.runtime.Path.ROOT == null
 while
org.eclipse.core.runtime.IPath.ROOT== "/"

@HannesWell probably related to
https://github.com/eclipse-equinox/equinox/pull/277

jukzi:
I dont't see an obvious error in the commit but reverting 277 solves the issue. Probably some sort of initialization order problem.

iloveeclipse:
I've marked https://download.eclipse.org/eclipse/downloads/drops4/I20230614-2220/ as unstable, it can't be used, SDK fails immediately on startup.

```
!SESSION 2023-06-15 08:31:05.944 -----------------------------------------------
eclipse.buildId=4.29.0.I20230614-2220
java.version=17.0.6-internal
java.vendor=N/A
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -os linux -ws gtk -arch x86_64

!ENTRY org.eclipse.osgi 4 0 2023-06-15 08:31:10.614
!MESSAGE An error occurred while automatically activating bundle org.eclipse.core.resources (74).
!STACK 0
org.osgi.framework.BundleException: Exception in org.eclipse.core.resources.ResourcesPlugin.start() of bundle org.eclipse.core.resources.
	at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:839)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:767)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:1032)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.startWorker(EquinoxBundle.java:371)
[...]

PR #279, opened 2023-06-15
Fix broken init of Path constants

iloveeclipse:
As long as IPath creates instances of Path in the IPath constants init code, Path can't back reference to same IPath constants that are not initialized yet - it will see null values and everyone who uses Path constants will be broken.

Fixes https://github.com/eclipse-equinox/equinox/issues/278

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IPath.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IPath.java
@@ -61,7 +61,7 @@ public interface IPath extends Cloneable {
 	 * 
 	 * @since 3.18
 	 */
-	public static final IPath EMPTY = fromOSString(""); //$NON-NLS-1$
+	public static final IPath EMPTY = Path.Constants.empty();
 
 	/**
 	 * Constant value containing the root path with no device on the local file
@@ -69,7 +69,7 @@ public interface IPath extends Cloneable {
 	 * 
 	 * @since 3.18
 	 */
-	public static final IPath ROOT = fromOSString("/"); //$NON-NLS-1$
+	public static final IPath ROOT = Path.Constants.root();
 
 	/**
 	 * Constructs a new path from the given string path. The string path must
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
@@ -51,12 +51,38 @@ public final class Path implements IPath, Cloneable {
 	 * Carrier that ensures that the contained constants are always initialized,
 	 * regardless of if the Path class or IPath interface is initialized fist.
 	 */
-	private static class Constants {
+	static class Constants {
 		/** Constant value indicating if the current platform is Windows */
 		static final boolean RUNNING_ON_WINDOWS = java.io.File.separatorChar == '\\';
 
 		/** Constant value indicating no segments */
 		static final String[] NO_SEGMENTS = new String[0];
+
+		/**
+		 * We have cycle : Path implements IPath and IPath uses Path object instances in
+		 * interface constants.
+		 * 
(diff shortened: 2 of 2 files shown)


laeubi (approved):
Looks reasonable and saves a suspicious cast! Thanks for looking into it.

merks:
It's always surprising to find that a public static final constant in an interface can be null, but it most certainly can be if the class is not yet fully initialized.

iloveeclipse:
> It's always surprising to find that a public static final constant in an interface can be null, but it most certainly can be if the class is not yet fully initialized.

Even more surprising is that depending on **use order** the JVM is able to init both IPath and Path properly with the old code, but only if Path is **used** first!

iloveeclipse:
**This** is now really surprising:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.IPath.segment(int)" because "org.eclipse.core.runtime.IPath.ROOT" is null
	at org.eclipse.equinox.common.tests.PathTest.testFirstSegment(PathTest.java:341)
```

merks:
EMF had horrible problems like this  with circularity between XyzPackage.eINSTANCE  which uses XyzPackageImpl as the value, so I know all too well how shocking and horrible this is!   😱

iloveeclipse:
The test works fine locally in SDK (different dependencies / init!!!), so this must be again some "unexpected" init sequence.

Adding reference to Path here breaks test also locally for me:

```
diff --git a/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/BundleDefaultPreferences.java b/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/BundleDefaultPreferences.java
index 93b9243..1303bce 100644
--- a/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/BundleDefaultPreferences.java
+++ b/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/BundleDefaultPreferences.java
@@ -17,2 +17,3 @@
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.preferences.*;
@@ -46,2 +47,3 @@
 		// cache the segment count
+		System.out.println("Path.root? " + Path.ROOT);
 		IPath path = IPath.fromOSString(absolutePath());
```

HannesWell:
Thanks Andrey for looking into this and Jörg for the report. 

> > It's always surprising to find that a public static final constant in an interface can be null, but it most certainly can be if the class is not yet fully initialized.
> 
> Even more surprising is that depending on **use order** the JVM is able to init both IPath and Path properly with the old code, but only if Path is **used** first!
> 
> image
> 

This is indeed difficult in this case and I faced a similar Problem in https://github.com/eclipse-equinox/equinox/pull/259/files which is why I introduced the static `Constants` class in `Path` which broke the cycle. Unfortunately this would not help in this case. And somehow I oversaw this case in the mentioned PR.

Probably the only way to reliably avoid the cycle regardless of the initi-order is to not reuse the constants and use own objects in Path respectivly IPath.
I hope/assume that this is not a problem and over time all refernces to runtime.Path should fade out any way.

Unfortunately I cannot look into the code and test fix at my Computer before this evening.

HannesWell:
> This is indeed difficult in this case and I faced a similar Problem in https://github.com/eclipse-equinox/equinox/pull/259/files which is why I introduced the static Constants class in Path which broke the cycle. Unfortunately this would not help in this case.

Thinking about it again, it actually could work to move the constants to the Constants class, but they have to be the last fields.

You can simply test this in a simple Java application with only a Main where where you use lPath first and then Path and vice-versa.

iloveeclipse:
> it actually could work to move the constants to the Constants class

That was not enough, so I came with a slightly different approach, see last commit.

HannesWell:
> > it actually could work to move the constants to the Constants class
> 
> That was not enough, so I came with a slightly different approach, see last commit.

Thanks! This approache should be bullet-proof.

PR #280, opened 2023-06-16
Move security bundles to Java 17

akurtakov:


akurtakov:
As per https://github.com/eclipse-equinox/equinox/pull/262#issuecomment-1587310444 these are fine to move.

PR #285, opened 2023-06-28
use stronger cipher for secure storage

gireeshpunathil:
move the default from weaker MD5/DES to a stronger SHA/AES which is available in JVMs.
PBEWithMD5AndDES -> PBEWithHmacSHA512AndAES_256

### DIFF
--- a/bundles/org.eclipse.equinox.security.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.security.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Equinox security tests
 Bundle-SymbolicName: org.eclipse.equinox.security.tests;singleton:=true
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 1.3.100.qualifier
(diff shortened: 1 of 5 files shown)


gireeshpunathil:
tests did not fail when ran locally, but 2 failed here, not sure why. are those known issues?

akurtakov:
The failures look caused by the patch as I don't remember seeing these failing . 
Tests run on linux so that could be one reason why you dont see it locally if you run on other OS.

gireeshpunathil:
thanks @akurtakov - I will investigate.

gireeshpunathil:
The decrypt was throwing different error code with the new cipher.
old: `Input length must be multiple of 8 when decrypting with padded cipher`, (code:3)
new: `Wrong parameter type: IV expected`, (code: 0)

So I adjusted the test to accommodate the new scenario.

tjwatson (commented):
We need micro-version increases to the bundles.  Otherwise looks good to me.

tjwatson (approved):

tjwatson:
> @tjwatson - just pushed a change to up the bundle versions, PTAL. thanks!
Thanks, I merged it.

iloveeclipse:
FYI, this was one from the root causes of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1388.

gireeshpunathil:
thanks @iloveeclipse for catching this!

PR #294, opened 2023-07-31
Add a standard osgi repository that includes all equinox bundles

laeubi:
Currently one has to wait until the next release or use the ibuilds composite p2 repository but there is currently no way to fetch latest equinox bundles in a more lightweight way.

This uses the new tycho-repositry-plugin to build a standard OSGi repository that could be deployed e.g. as part of the mater verification build.

WDYT @akurtakov @tjwatson? The idea would be that we can deploy a standard osgi repository with each master build and possibly even have an automated release build to publish equinox bundles independent from regular release cycles, at best such repository would then even be used as contribution to simrel some time....

But first goal would be to supply something standardized to the outside world without much effort.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,7 +31,7 @@ pipeline {
 			}
 			post {
 				always {
-					archiveArtifacts artifacts: '**/*.log, **/*.jar', allowEmptyArchive: true
+					archiveArtifacts artifacts: '**/*.log, **/*.jar, **/repository/target/repository-deploy.zip', allowEmptyArchive: true
 					junit '**/target/surefire-reports/TEST-*.xml'
 					discoverGitReferenceBuild referenceJob: 'equinox/master'
 					recordIssues publishAllIssues: true, tools: [java(), mavenConsole(), javaDoc()]
--- a/pom.xml
+++ b/pom.xml
@@ -220,6 +220,7 @@
 		    <module>bundles/org.eclipse.equinox.launcher.win32.win32.x86_64</module>
 		
 		    <module>bundles/org.eclipse.osgi.tests</module>
+		    <module>repository</module>
         </modules>
     </profile>
   </profiles>
--- a/repository/pom.xml
+++ b/repository/pom.xml
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.eclipse.equinox</groupId>
+		<artifactId>parent</artifactId>
+		<version>4.29.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>repository</artifactId>
+	<packaging>repository</packaging>
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-repository-plugin</artifactId>
+				<version>${tycho.version}</version>
+				<extensions>true</extensions>
+				<executions>
+					<!-- Additionally we create another repository using the 'local' layout and place it in a deploy folder -->
+					<execution>
+						<id>local</id>
+						<phase>package</phase>
+						<goals>
+							<goal>package-repository</goal>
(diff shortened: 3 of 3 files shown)


tjwatson:
> Maven repositories are quite standard too, and the Equinox bundles are available daily from the follow Maven repository, so they're readily and quickly available:

I was about to respond similarly.  I think it would be better to publish to a maven repository for projects that want to consume the very latest I-Builds in a "standard" way other than the I-Build p2 repository.

laeubi:
@merks Funny enough each time I suggest using maven directly I get "but this requires m2e" or "people are used to http" and there are even people converting maven-p2-repositories to http ones  ;-P

But guess what, this additional creates a maven deployable artifact that can be consumed from maven and then allows to consume everything from maven, so deploy it to e.g download.eclipse.org would just be a lower barrier point and unless the maven deploy is all setup and to allow to attract some consumers.

Nerveless to work efficiently one need some kind of meta-data-repository to consume first, as maven has the drawback that it can only access **known artifact coordinates** so there is no way to know what artifact on central exports the package `org.eclipse.equinox.osgi` or provides the capability `osgi.extender; osgi.serviceloader.registrar`, this is where a repository.xml fill that gap, I have attached the result of the build here for easy investigation:

[maven-deployable-repository.xml.zip](https://github.com/eclipse-equinox/equinox/files/12217981/maven-deployable-repository.xml.zip)

merks:
Funny enough, even when artifacts are published to a Maven repository daily, tested that all Maven dependencies resolve, there are still people who look for yet another way to accomplish that same goal.

Probably we digress, but is it documented somewhere how this thing converts package imports into Maven dependencies or how it maps onto Maven group/artifact IDs?  Is that compatible with what Equinox is currently publishing to Maven?   But then I wonder, is Maven even relevant to this new goal?  

I see the zip file contains a similar/analogous representation to p2.  Is this the OBR standard?

laeubi:
> Funny enough, even when artifacts are published to a Maven repository daily, tested that all Maven dependencies resolve, there are still people who look for yet another way to accomplish that same goal.

There is a difference between something published on a graceful basis by a larger authority once a day (what would happen if another project broke the build?) on behalf of project versus the project having the power of publishing (and releasing!) its artifacts whenever needed / decided even on every build. So this is not really the same goals.

> but is it documented somewhere how this thing converts package imports into Maven dependencies or how it maps onto Maven group/artifact IDs

This has nothing to do with maven so there is no documentation about that, it simply collects all bundles of a build not more or less.

>  Is that compatible with what Equinox is currently publishing to Maven?

Equinox is not publishing anything to maven directly but if it would it would be compatible because the `deploy` and `repository` mojo uses the same underlying data.

> But then I wonder, is Maven even relevant to this new goal?

Maven is used to _build_ and can be used to _deploy_ it but in general it is not related to the topic per se that's why I mention a "traditional" Http deployment (without any maven reference then) as a first step.

> I see the zip file contains a similar/analogous representation to p2. Is this the OBR standard?

It is https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.repository.html and my long-term goal is to establish this format as a comparable alternative to the vendor specific P2 format, so there will be usable as updatesite, target location in PDE/Tycho/P2 UI ...

laeubi:
I now have modified the Jenkinsfile to also collect the (http deployable) repository ([repository-deploy.zip](https://ci.eclipse.org/equinox/job/equinox/job/PR-294/2/artifact/repository/target/repository-deploy.zip)):

https://ci.eclipse.org/equinox/job/equinox/job/PR-294/2/artifact/repository/target/

merks:
> It is https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.repository.html and my long-term goal is to establish this format as a comparable alternative to the vendor specific P2 format, so there will be usable as updatesite, target location in PDE/Tycho/P2 UI ...

It's not entirely clear why "we" need an alternative, especially for something that impacts so many downstream technologies...

laeubi:
> It's not entirely clear why "we" need an alternative, especially for something that impacts so many downstream technologies...

Because P2 already lacks support for more advanced OSGi techniques and because of its "closed model" is not able to adapt to this and other tools like to use standard documented techniques that are portable.

Beside this I don't claimed "we" need it but there are people that like to use modern and flexible techniques even if "we" lacking more and more behind and there was the same discussion long time ago when I first added support for maven-targets and today even "we" use them so it is not completely hopeless anyways ;-)

laeubi:
Support for PDE was added here:
- https://github.com/eclipse-pde/eclipse.pde/pull/692

laeubi:
Tycho support is here:
- https://github.com/eclipse-tycho/tycho/pull/2844

PR #295, opened 2023-07-31
Implement OSGi Service Loader Mediator Specification

HannesWell:
In order to use bundles that use the Java `ServiceLoader` mechanism an implementation of the `OSGi Service Loader specification` is required at runtime. The reference implementation, which is also used by the Eclipse Platform at the moment, is Apache Aries SPI Fly, which uses class-file transformation in order to intercept calls to the ServiceLoader and needs to be configured as auto-started bundle.

This PR is a simple prototype with the intend to implement the `OSGi Service Loader specification` directly in org.eclipse.osgi.
The basic idea is to intercept `ServiceLoader` calls by using a `ClassLoaderHook` and searching for the ServiceLoader in the caller classes if the chances are high that the `ClassLoader.getResource()` and `Class.forName()` is called by the ServiceLoader.

This is currently a draft, that I have implemented naively without full knowledge of the entire `OSGi Service Loader specification`, so there are probably many things to work out. Nevertheless I think it already demonstrates the basic approach and at least for supplying a logging provider to the `slf4j.api` bundle this already works.

@laeubi FYI
@tjwatson what do you think?


HannesWell:
The basic ideas could also be used to enable the usage of the `ServiceLoader` with less configuration in OSGi as part of the Core framework as discussed in https://github.com/osgi/osgi/issues/372.

laeubi:
I think this is a very creative way to solve the problem but I'm not sure if the resource/classlaoding is *always* triggered, e.g what happens if the classes are already loaded? So the best would be to add a testcase that covers all possible call variants of service loader especially those that lookup the services multiple times (also using reset())...

@rotty3000 might can help on the details with the specification itself...

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/hooks/ServiceLoaderMediatorHook.java
```diff
@@ -0,0 +1,169 @@
...
+			});
+			if (services != null) {
+				Optional<Class<?>> findFirst = services.findFirst();
+				return findFirst.orElse(null);
+			}
+		}
+		return null;
+	}
+
+	private <T> Stream<T> searchProviders(String name, ModuleClassLoader classLoader,
```
laeubi, line 124: Instead of dynamically searching this list I would use BundleTracker in the Hook that collects bundles and read their `META-INF/services` in a datastructure that can then be used to lookup current active providers, that has several advantages:

1. no need to query the wiring and iterate the bundles
2. easier to debug as there is some current state to inspect
3. instead of the `latestServiceName` we can inspect if the bundle tries to load a class that is an SPI of another bundle and then inspect the stack, requiring a certain call order seems a bit fragile if something changes in the implementation of service loader.

tjwatson:
> I think this is a very creative way to solve the problem but I'm not sure if the resource/classlaoding is _always_ triggered, e.g what happens if the classes are already loaded? So the best would be to add a testcase that covers all possible call variants of service loader especially those that lookup the services multiple times (also using reset())...

Tests will be needed regardless.  But handling cases where the class is already loaded should not be a concern if the class is loaded by a framework class loader.  This is because the hook will be in place BEFORE any class is defined from any and all bundles installed in the framework.  Say something is loaded by an OSGi connect class loader?  I am fairly certain there is no solution that can implement the service loader mediator in that case.  Only way that will be solved generally is if we can convince Java to add something to the service loader to allow us to hook into it instead of hacking around with class loaders to do this.

laeubi:
> Only way that will be solved generally is if we can convince Java to add something to the service loader to allow us to hook into it instead of hacking around with class loaders to do this.

That would be great. Do you think it would be possible to suggest such think to the OpenJDK team and describe what would be required from framework POV?

>  But handling cases where the class is already loaded should not be a concern if the class is loaded by a framework class loader.

I think you are right, the only thing I can think of is if the bundle is "too late" then the user might need to reset the serviceloader to make this visible.

tjwatson:
> I think you are right, the only thing I can think of is if the bundle is "too late" then the user might need to reset the serviceloader to make this visible.

That is a limitation of Equinox framework hooks, they can only be added when the framework is initialized.  They cannot be added dynamically later.  If using Equinox Framework Hooks then the framework must be restarted to load up the mediator implementation.

laeubi:
I mean if an impl do severcieloaderlookup and get some classes and after that I install another bundle then there is no way to tell the consumer "here is another one" ... the hook should best be installed right away from the start of the framework.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/hooks/ServiceLoaderMediatorHook.java
```diff
@@ -0,0 +1,169 @@
...
+import org.osgi.framework.Constants;
+import org.osgi.framework.wiring.BundleRequirement;
+import org.osgi.framework.wiring.BundleWire;
+import org.osgi.framework.wiring.BundleWiring;
+
+public class ServiceLoaderMediatorHook extends ClassLoaderHook {
+
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	private Supplier<Optional<Class<?>>> getCallerClassComputer() {
+		try { // The Java-9+ way
```
tjwatson, line 32: If you place this in a framework fragment then we can make it depend on a higher osgi.ee than Java 8.

HannesWell:
Thanks for all the feedback, I have already started to implemented some of the suggestions.
But I'm on vacation until mid next week and will update this PR when back and on a good new state.



> I mean if an impl do severcieloaderlookup and get some classes and after that I install another bundle then there is no way to tell the consumer "here is another one" ... the hook should best be installed right away from the start of the framework.

Yes, but I assume that ServiceLoader objects are not kept for a long time. I guess that the usual usage pattern is that one looks for as many instances of a service as desired and collects them and if later interested in instances again creates a fresh instance.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/hooks/ServiceLoaderMediatorHook.java
HannesWell: > > Instead of dynamically searching this list I would use BundleTracker in the Hook that collects bundles and read their `META-INF/services` in a datastructure that can then be used to lookup current active providers, that has several advantages:
> 
> IIRC the service loader mediator specification requires the impl to look at the wiring to allow a bundle to "use" only the providers it is wired to.

I think so as well. However having a state also allows to avoid the ThreadLocal.
In general I try to have a somewhat hybrid approach that maintains a state with all services, but when there is a query for services checks the current wires if that bundle is permitted to really obtain them.
The goal is also a bit to demonstrate how this could be done totally without (or at least with minimal) metadata as intended in https://github.com/osgi/osgi/issues/372. If that issue is once resolved all the checks for wires could simply be removed and everything possible could be supplied.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/hookregistry/HookRegistry.java
```diff
@@ -113,6 +114,7 @@ public void initialize() {
 			addClassLoaderHook(new DevClassLoadingHook(container.getConfiguration()));
 			addClassLoaderHook(new EclipseLazyStarter(container));
 			addClassLoaderHook(new WeavingHookConfigurator(container));
+			addClassLoaderHook(new ServiceLoaderMediatorHook());
```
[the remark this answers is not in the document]
HannesWell: I considered this as well, mainly for the point you mentioned below that it would allow later Java versions than 8 (9 would already be a big win).
But the big drawback of this approach is that one has to list the fragment as framework-extension in a system-property of the runtime. For example in a launch config:
`-Dosgi.framework.extensions=reference:file:${project_loc:/org.eclipse.equinox.spi}`, in a product definition this would usually be `-Dosgi.framework.extensions=reference:file:org.eclipse.equinox.spi`, assuming the spi and o.e.osgi jar are colocated.
Or is there another way to load a framework-extension with the framework's classloader so that the `hookconfigurators.properties` are found?
Because actually another goal of this was to avoid the need (if possible) for any configuration that it just works out of the box.
For the build-time some p2-instructions could be added to the fragment, but PDE does not yet have any help for that as far as I know.

laeubi:
@HannesWell any chance we can push this forward?

HannesWell:
> @HannesWell any chance we can push this forward?

Yes. I will first complete my work on the native launcher build pipeline and will then continue with this.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/hooks/ServiceLoaderMediatorHook.java
laeubi: I just wanted to mention it here for completeness, but it would also be possible to use a Multi-Release jar if we just extract this part into a Utilclass.

HannesWell:
In general this now in a working state and the point mainly missing are tests.
I've already added a first test that adds the foundation for these tests but more scenarios are missing.

That also makes me wondere: Is there a TCK for the `Service Loader Mediator` spec?

laeubi:
@HannesWell is this ready to be completed now? It looks like the TCKs are included now and it would be the perfect time having this before M1.

HannesWell:
> is this ready to be completed now? It looks like the TCKs are included now and it would be the perfect time having this before M1.

The implementation should be ready and I already tested it with a simple application of mine.
I agree we should finally complete this soon.
What's only missing are the tests. Our own tests plus the TCK. The last time I looked into it the TCK tests not only failed their assertions but also their setup was very unstable.
I'm fine to put the TCK cases aside for now (of course it would be great to have them working from the start, but I don't know how much time that would take).

But at least our own tests should work. And the last time I looked into it, I had issues isolating the 'test' services provider and consumer bundles installed into the runtime for each test case among the test cases. Often one provider was still available in the next case and therefore broke it. I was looking into ways to reliably await the full uninstallation and removal of these test bundles.

I'll look into it in the next days, but if you or anybody else want to help, that's more than welcome.

bundles/org.eclipse.equinox.spi/META-INF/services/org.eclipse.equinox.spi.internal.ServiceLoaderMediatorHook
```diff
@@ -0,0 +1 @@
+org.eclipse.equinox.spi.caller.tracing.Dummy
```
tjwatson, line 1: What is this?  I don't see that the `Dummy` class exists.

laeubi:
@HannesWell one test failure is caused by caching due to use of `ContextFinder` usage so better people would use `ServiceLoader<S> load(Class<S> service, ClassLoader loader)` to avoid stale instances - but I think thats somehow unavoidable with this approach.

The second was a problem with host/fragment tracking - it becomes a bit complex here (but the test passes) but might be simplified further if we just track all bundles and then just check if the `BundleServices` is empty or similar.

laeubi:
After adding an extra requirement the TCK should at least start again.

laeubi:
This now uncovered a little bug in Tycho as well - I installed a workaround and made the SPI test be more locally scoped what hopefully will restore the TCK execution again.

tjwatson:
I want to take a close look at the implementation of this tomorrow (Friday).  First I wanted to understand if we have a hard dependency on Java 17 from a specification perspective.   Is an implementation required to use Java 17+ APIs to implement?

laeubi:
> I want to take a close look at the implementation of this tomorrow (Friday). First I wanted to understand if we have a hard dependency on Java 17 from a specification perspective. Is an implementation required to use Java 17+ APIs to implement?

Java 9+ (due to `StackWalker`) is the minimum requirement here, Java 17 just drips in accidentally because it is the "default" and Tycho currently compiles bundles without a BREE maybe wrong using Java 17 here and the TCKs last time I checked didn't worked with Java 21.

tjwatson:
> > I want to take a close look at the implementation of this tomorrow (Friday). First I wanted to understand if we have a hard dependency on Java 17 from a specification perspective. Is an implementation required to use Java 17+ APIs to implement?
> 
> Java 9+ (due to `StackWalker`) is the minimum requirement here, Java 17 just drips in accidentally because it is the "default" and Tycho currently compiles bundles without a BREE maybe wrong using Java 17 here and the TCKs last time I checked didn't worked with Java 21.

But is that just an implementation detail or is there something in the specification that makes it impossible to implement without `StackWalker`.

laeubi:
> > > I want to take a close look at the implementation of this tomorrow (Friday). First I wanted to understand if we have a hard dependency on Java 17 from a specification perspective. Is an implementation required to use Java 17+ APIs to implement?
> > 
> > 
> > Java 9+ (due to `StackWalker`) is the minimum requirement here, Java 17 just drips in accidentally because it is the "default" and Tycho currently compiles bundles without a BREE maybe wrong using Java 17 here and the TCKs last time I checked didn't worked with Java 21.
> 
> But is that just an implementation detail or is there something in the specification that makes it impossible to implement without `StackWalker`.

As SPIFly implements it there is obviously a way - it was chosen here as a first minimal feasible implementation detail.

tjwatson:
> As SPIFly implements it there is obviously a way - it was chosen here as a first minimal feasible implementation detail.

Fair enough.  I asking from a spec perspective.  So the answer appears to be no hard dependencies on Java 9+ to implement the specification.

laeubi:
I now have pushed one last change that actually runs the official TCKs but due to https://github.com/osgi/osgi/pull/969 the official tcks currently mess up the whole test runtime so they just fail/timeout.

I'll try to catch up tomorrow with some improvements in Tycho and some more investigation, the code itself can be reviewed anyways.

laeubi:
> > As SPIFly implements it there is obviously a way - it was chosen here as a first minimal feasible implementation detail.
> 
> Fair enough. I asking from a spec perspective. So the answer appears to be no hard dependencies on Java 9+ to implement the specification.

Nope the spec does not require any specific technique - so its just our own current implementation here that needs this.

HannesWell:
> one test failure is caused by caching due to use of `ContextFinder` usage so better people would use `ServiceLoader<S> load(Class<S> service, ClassLoader loader)` to avoid stale instances - but I think thats somehow unavoidable with this approach.
> 
> The second was a problem with host/fragment tracking - it becomes a bit complex here (but the test passes) but might be simplified further if we just track all bundles and then just check if the `BundleServices` is empty or similar.

Thanks for providing these fixes. I can confirm that the proposed fixes make the tests pass and also agree that they should be revised and it should be checked if it can be integrated simpler.

I did a little bit of polishing but don't have to complete it now, but I plan to complete the polishing and resolution of all remaining TODOs by the end of next week and also to look in detail into the proposed fix of the fragment issue.
I hope that this is then finally ready.

Of course it would be great if the TCK tests pass until then too, but I would not make this a requirement for this as we also have own tests that should cover the basic functionality.
Nevertheless, thanks a lot for looking into their issues. That's really helpful.

laeubi:
For development purpose it should be possible to run the individual tests but we can disable them for the final merge until https://github.com/bndtools/bnd/pull/7421 is released (what will allow Tycho ti implement https://github.com/eclipse-tycho/tycho/issues/6334 )

bundles/org.eclipse.equinox.spi/src/org/eclipse/equinox/spi/internal/ServiceLoaderMediatorHookConfigurator.java
```diff
@@ -0,0 +1,50 @@
...
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+
+public class ServiceLoaderMediatorHookConfigurator implements HookConfigurator {
+	private static ServiceLoaderMediatorHook mediatorHook;
+
+	@Override
+	public void addHooks(HookRegistry hookRegistry) {
+		ServiceLoaderMediatorHook hook = new ServiceLoaderMediatorHook();
+		hookRegistry.addClassLoaderHook(hook);
```
tjwatson, line 29: Hooks must be added during framework initialization.  This is required for  `BundleFileWrapperFactoryHook` and `StorageHookFactory` because these hooks must be in place before we can read from the framework persistent state and create the actual `BundleFile` objects for accessing bundle content from the framework.  But I think there is no need for the same restriction for `ClassLoaderHook`.  Maybe we should allow framework extensions to register their own `ClassLoaderHook` during framework extension start.  This would reduce the burden of this framework extension because it could avoid having to have an implementation of `HookConfigurator` which must be available on the framework classpath directly before creating the `Framework` object.

We could enhance the framwork to allow framework extensions to register `ClassLoaderHook` implementations as a service when they use the system bundle `BundleContext` to register the hook, for example from the `Activator` class you have below.

bundles/org.eclipse.equinox.spi/src/org/eclipse/equinox/spi/internal/ServiceLoaderMediatorHook.java
```diff
@@ -0,0 +1,390 @@
...
+import org.osgi.framework.ServiceRegistration;
+import org.osgi.framework.namespace.HostNamespace;
+import org.osgi.framework.wiring.BundleRevision;
+import org.osgi.framework.wiring.BundleWire;
+import org.osgi.framework.wiring.BundleWiring;
+import org.osgi.service.serviceloader.ServiceLoaderNamespace;
+import org.osgi.util.tracker.BundleTracker;
+import org.osgi.util.tracker.BundleTrackerCustomizer;
+
+public class ServiceLoaderMediatorHook extends ClassLoaderHook implements BundleTrackerCustomizer<Bundle> {
```
tjwatson, line 58: Can you describe the requirements for the set thread context class loader in order to allow this hook to function properly?  I assume it must be set to something other than the default JVM platform class loader.  I assume somehow the set thread context class loader must be a bundle class loader or at least delegate to some bundle class loader in order for this approach to work.

PR #298, opened 2023-08-05
Servletbridge - Use both underscore and hypen for version separator

rougeSE:
Still using underscore as the primary separator but now there is an array of possible version separators to check instead of repeating the strings throughout the code.

### DIFF
--- a/bundles/org.eclipse.equinox.servletbridge/src/org/eclipse/equinox/servletbridge/FrameworkLauncher.java
+++ b/bundles/org.eclipse.equinox.servletbridge/src/org/eclipse/equinox/servletbridge/FrameworkLauncher.java
@@ -46,6 +46,7 @@ public class FrameworkLauncher {
 	private static final String REFERENCE_SCHEME = "reference:"; //$NON-NLS-1$
 	private static final String CONFIG_INI = "config.ini"; //$NON-NLS-1$
 	private static final String DOT_JAR = ".jar"; //$NON-NLS-1$
+	private static final String[] VERSION_SEPS = {"_", "-" }; //$NON-NLS-1$; //$NON-NLS-2$ 
 	private static final String WS_DELIM = " \t\n\r\f"; //$NON-NLS-1$
 	protected static final String FILE_SCHEME = "file:"; //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.servletbridge/src/org/eclipse/equinox/servletbridge/FrameworkLauncher.java
```diff
@@ -46,6 +46,7 @@ public class FrameworkLauncher {
 	private static final String REFERENCE_SCHEME = "reference:"; //$NON-NLS-1$
 	private static final String CONFIG_INI = "config.ini"; //$NON-NLS-1$
 	private static final String DOT_JAR = ".jar"; //$NON-NLS-1$
+	private static final String[] VERSION_SEPS = {"_", "-" }; //$NON-NLS-1$; //$NON-NLS-2$
```
tjwatson, line 49: Instead of an array here I would rather see two separately named constants like `VERSION_SEP_UNDERSCORE` and `VERION_SEP_DASH`.

bundles/org.eclipse.equinox.servletbridge/src/org/eclipse/equinox/servletbridge/FrameworkLauncher.java
```diff
@@ -896,9 +900,13 @@ protected int findMax(String[] candidates) {
 		for (int i = 0; i < candidates.length; i++) {
 			String name = candidates[i];
 			String version = ""; //$NON-NLS-1$ // Note: directory with version suffix is always > than directory without version suffix
-			int index = name.indexOf('_');
-			if (index != -1)
-				version = name.substring(index + 1);
```
tjwatson, line 901: This would then become

			int index = name.indexOf(VERSION_SEP_UNDERSCORE);
			if (index != -1) {
				index = name.indexOf(VERSION_SEP_DASH);
			}
			if (index != -1) {
				version = name.substring(index + 1);
			}

akurtakov:
@rougeSE Do you plan to amend accorting to @tjwatson 's comments?

laeubi:
Closing this for now no response from the author. @rougeSE if you still want to get this change in please rebase, resolve conflicts and adress the mentioned review comments then we can just repopen this one.

PR #300, opened 2023-08-07
Equinox Caching: Improved Exception and Shutdown Handling

xpomul:
Sadly, #233 is still not fully solved. The "truncated class file" issue occurs less often now, but it still does occur. 
In the last weeks and months, I have analyzed the issue in more depth and applied a fix locally in our codebase. With this now the issue seems to be resolved.

The issue now occurred only between shutdown and restart of the Eclipse Platform which led me to the root cause that the CacheWriter is interrupted when the Platform is shutting down, but if the Platform is fast enough, the CacheWriter might be interrupted or even terminated hard via System.exit() when the Java Process ends. 

To compensate, I have added two things:
1. I have added a join() for the CacheWriter thread with a 5s timeout. Usually it will terminate much faster than that, but you never know
2. I have pushed the Exception handling down from the main loop of the CacheWriter into the method where writing actually happens. I have also added a check that the file has been written completely. If for any reason (not matching size, interrupt encountered, exception caught) there is a risk of the cache file being corrupt, then we rather delete the cache file. This would lead to a cache miss the next time and thus, some performance penalty, as the class needs to be woven again. But at least we can be more certain that the cache file isn't corrupt.


tjwatson:
@martinlippert could you have a review on this?

martinlippert:
Quick heads-up: I am currently away from work, so can't look at this before Aug 28. If that is fine, happy to look into this then.

martinlippert (approved):
Looks good to me, thanks a lot for analyzing this in so much depth.

tjwatson:
I think this will need to wait until after we switch over to 2023-12 development because we are heading into RCs for 2023-09 release.

vogella:
Planning to merge after a green build based on @tjwatson comment

xpomul:
@vogella @tjwatson seems that this PR was forgotten? It's still not merged.
Any chances to get it into 2024-03 at least?

merks:
@xpomul 

Yes, it seems to be overlooked.  Shall I push the Update branch button to kick off new builds?

merks:
The failures are only the expected familiar ones that are unrelated.

merks:
@xpomul 

Sorry this slipped through the cracks and thank you for the contribution and the general reminder. 🥇

xpomul:
@merks Thanks, Ed :-)

PR #309, opened 2023-09-06
Version bumps for 4.30 stream

MohananRahul:
Tracked in
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1319


akurtakov:
Thanks!

PR #311, opened 2023-09-14
Ensure that the JSP support works with newer dependencies

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1251

### DIFF
--- a/bundles/org.eclipse.equinox.jsp.jasper/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.jsp.jasper/META-INF/MANIFEST.MF
@@ -4,14 +4,15 @@ Bundle-Name: %bundleName
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-SymbolicName: org.eclipse.equinox.jsp.jasper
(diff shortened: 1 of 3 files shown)


merks:
@akurtakov 

I'm not sure who else might want to review this?   The "continuous-integration/jenkins/pr-head" build passed....

akurtakov:
What is the reason to restrict jasper version to [9,10) ? It might be good to keep the current version in the range.

merks:
> What is the reason to restrict jasper version to [9,10) ? It might be good to keep the current version in the range.

I did that originally the first time I made the changes, but this time around when I did the changes, I noticed that this initializer class is not in the older bundle:

image

akurtakov:
That's good enough explanation. If/when you have the whole story with jsps working please merge this one together with the rest so we get less build issues.

merks:
These are the outstanding changes.  Remove (replace) the references to org.apache.jasper.glassfish here:

image

image

image

image

image

I think I need to commit this PR before I do the above removals...

Locally the help system works without org.apache.jasper.glassfish in the installation:

image

PR #312, opened 2023-09-15
JSP compilation depends on the org.eclipse.jdt.core.compiler package

merks:
Add an include of the org.eclipse.jdt.core.compiler.batch plugin to the org.eclipse.equinox.sdk feature.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1251

### DIFF
--- a/features/org.eclipse.equinox.sdk/feature.xml
+++ b/features/org.eclipse.equinox.sdk/feature.xml
@@ -145,4 +145,17 @@
          version="0.0.0"
          unpack="false"/>
 
+   <plugin
+         id="org.eclipse.jdt.core.compiler.batch"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.jdt.core.compiler.batch.source"
+         download-size="0"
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Ed, few questions: 
- org.eclipse.jdt.core.compiler.batch is also included in org.eclipse.jdt feature. Should it be removed there?
- why org.eclipse.equinox.sdk needs org.eclipse.jdt.core.compiler.batch? Wasn't jsp thing required in the "help" / platform?

merks:
The JSP thing is provided by 

> Ed, few questions:
> 
>     * org.eclipse.jdt.core.compiler.batch is also included in org.eclipse.jdt feature. Should it be removed there?
> 

No I think it belongs in the JDT feature.

>     * why org.eclipse.equinox.sdk needs org.eclipse.jdt.core.compiler.batch? Wasn't jsp thing required in the "help" / org.eclipse.platform?

The help system uses the JSP support from Equinox:

image

And the newer JSP compilers use JDT's batch compiler.

laeubi:
Can't we activate to include all dependencies in the SDK or do we need to explicitly manage each item separately?

PR #317, opened 2023-09-29
Port o.e.equinox.http.jetty to Jetty 12

akurtakov:
Port is straightforward except for logging as EquinoxStdErrLog had to be removed as this org.eclipse.jetty.util.log api is totally gone and has been deprecated since Jetty 10.x.
According to
https://eclipse.dev/jetty/documentation/jetty-12/operations-guide/index.html#og-server-logging there is no longer programatic way to configure logging.


akurtakov:
Tracked in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1384 .

laeubi:
@akurtakov I just wanted to note that jetty itself brings an OSGi-HTTP-Service implementation, since that service is deprecated by the spec, we might want to drop the equinox impl all together and use the jetty one where needed (I think its only the help?)

akurtakov:
> @akurtakov I just wanted to note that jetty itself brings an OSGi-HTTP-Service implementation, since that service is deprecated by the spec, we might want to drop the equinox impl all together and use the jetty one where needed (I think its only the help?)

Would you please give links to Jetty's OSGi-HTTP-Service implementation and to the deprecation in the spec for references here?

laeubi:
OSGi R7 is the last specification containing HttpService:

- https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.html

R8.1 has now removed it from the spec:

- https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.http.html

Sadly as far as I know there are no real deprecation notices (but @bjhargrave , @tjwatson  or @timothyjward might have a reference somewhere)

- https://github.com/osgi/osgi/issues/551

Also the this (failed) attempt to upgrade `HttpService` to jakarta:

- https://github.com/osgi/osgi/issues/551

The jetty-osgi bundles live here:
- https://mvnrepository.com/artifact/org.eclipse.jetty.osgi

Apache Felix also offers `HttpService` we might can use instead of maintain our own:

https://cwiki.apache.org/confluence/display/FELIX/Apache+Felix+HTTP+Service

tjwatson:
> Sadly as far as I know there are no real deprecation notices (but @bjhargrave , @tjwatson or @timothyjward might have a reference somewhere)

Because it was never deprecated.  The chapter just was not included in the 8.1 specification.  There are no plans to enhance it or provide the equivalent for Jakarta Servlet.  But Java Servlet users are free to keep using an implementation of the Http Service.  The problem Equinox has is that we have various services that are built upon the old Http Service that:

1. Need to migrate to Jakarta
2. Need to migrate to Http Service

The main one is the `org.eclipse.equinox.http.registry`.  This technically can be done.  But then the ripple effect will be large as it impacts Eclipse UA (help) but then also impacts any other projects that contribute help content with the extensions.

But I don't want to conflate updating to Jakarta with the work here that Alex has started, because it seems unrelated since Jetty 12 decouples the EE version (javax vs jakarta) so seems updating to Jetty 12 does not force Jakarta on us.

akurtakov:
Cross context dispatch is not working because of https://github.com/eclipse/jetty.project/issues/10617 but as this seems to not be priority for Jetty we should just move on.

PR #319, opened 2023-09-30
Correct description of o.e.e.compendium.sdk

Torbjorn-Svensson:
The string in the properties file is not HTML, so do not use `&quot;`.


akurtakov:
Thanks!

PR #320, opened 2023-09-30
Paths starting with "//?/" on Windows are local

Torbjorn-Svensson:
The prefix might be used to support paths longer than MAX_PATH.
https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces

### DIFF
--- a/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
+++ b/bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
@@ -318,6 +318,23 @@ public class PathTest {
 		assertEquals("3.0.win", "D:/foo/abc.txt", IPath.forWindows("/D:/foo/abc.txt").toString());
 		// fullPath = new java.io.File("D:/").toURL().getPath()
 		assertEquals("3.1.win", "D:/", IPath.forWindows("/D:/").toString());
+
+		final String verylongpath = "C:/dev/verylongpath/00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/"
+				+ "00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/root/lib/gcc/arm-none-eabi/7.3.1/include-fixed";
+		anyPath = IPath.forWindows(verylongpath);
+		assertEquals("3.2.1.win", "C:", anyPath.getDevice());
+		assertEquals("3.2.2.win", "dev", anyPath.segment(0));
+		assertEquals("3.2.3.win", verylongpath, anyPath.toString());
+
+		anyPath = IPath.forWindows("//?/" + verylongpath);
+		assertEquals("3.3.1.win", "C:", anyPath.getDevice());
+		assertEquals("3.3.2.win", "dev", anyPath.segment(0));
+		assertEquals("3.3.3.win", verylongpath, anyPath.toString());
+
+		anyPath = IPath.forWindows("\\\\?\\" + verylongpath.replace('/', '\\'));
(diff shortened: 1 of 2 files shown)


HannesWell:
This change makes the Path constructor handle `//?/`, but I wonder shouldn't the `//?/` prefix be stored somehow in order to make it available when the file-system is queried eventually?
But adding it to the device is probably not the right thing to do.

Torbjorn-Svensson:
> This change makes the Path constructor handle `//?/`, but I wonder shouldn't the `//?/` prefix be stored somehow in order to make it available when the file-system is queried eventually?
> But adding it to the device is probably not the right thing to do.

Adding it to the device is not the correct thing to do and I'm not sure if it makes sense to save it.
The main use for it, as far as I know, is to avoid the `MAX_PATH` length limitation. I suppose the JVM is already doing something along these lines internally, as my testing shows that java applications does not have problem of managing deep file trees.

Torbjorn-Svensson:
Rebased PR on 1d9eb47f09654da0ed7c205fa2cf8e664b97c08b.
Anything I need to do prior to getting this merged?

vogella:
@HannesWell any more concerns or can this be merged?

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
```diff
@@ -318,6 +318,18 @@ public void testConstructors() {
...
 		assertEquals("3.1.win", "D:/", IPath.forWindows("/D:/").toString());
+
+		final String verylongpath = "C:/dev/verylongpath/00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/"
+				+ "00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/00112233445566778899aabbccddeeff/root/lib/gcc/arm-none-eabi/7.3.1/include-fixed";
+		anyPath = IPath.forWindows(verylongpath);
+		assertEquals("3.2.1.win", "C:", anyPath.getDevice());
+		assertEquals("3.2.2.win", "dev", anyPath.segment(0));
+		assertEquals("3.2.3.win", verylongpath, anyPath.toString());
+
+		anyPath = IPath.forWindows("//?/" + verylongpath);
```
HannesWell, line 329: On Windows backslashes are used as separator by default and as the linked guide mentions `\\?\` it should be tested literally with that value (even if the Path constructor replaces `\` by `/`).
```suggestion
		anyPath = IPath.forWindows("\\\\?\\" + verylongpath);
```

HannesWell (changes requested):
Sorry for the late reply.


> > This change makes the Path constructor handle `//?/`, but I wonder shouldn't the `//?/` prefix be stored somehow in order to make it available when the file-system is queried eventually?
> > But adding it to the device is probably not the right thing to do.
> 
> Adding it to the device is not the correct thing to do and I'm not sure if it makes sense to save it. The main use for it, as far as I know, is to avoid the `MAX_PATH` length limitation. I suppose the JVM is already doing something along these lines internally, as my testing shows that java applications does not have problem of managing deep file trees.

That's right. I tested it and with the standard Java File or Path API one can handle files and directories straight forward without that prefix (unlike for example the explorer). So it does not need to be added. And with JDK22 Java will also strip  it off:
https://jdk.java.net/22/release-notes#JDK-8287843

So in general this change is fine, only the test should be slightly adjusted.

bundles/org.eclipse.equinox.common.tests/src/org/eclipse/equinox/common/tests/PathTest.java
Torbjorn-Svensson: I added a test for both variants, is this okay for you?

HannesWell (approved):
Looks all good now.
Thank you for this contribution.

PR #322, opened 2023-10-01
Force English locale for date tests

Torbjorn-Svensson:
When using `java.text.SimpleDateFormat` without a locale, the system default locale will be considered and if that is not set to an English variant, the "EEE" part of the date string will be in a non-English language. To overcome this, ensure that the English locale is used for all tests where a `java.util.Date` is stringified.


akurtakov:
Would you please resolve the conflict ? There is no need to bump the version anymore. Otherwise the patch looks good.

Torbjorn-Svensson:
> Would you please resolve the conflict ? There is no need to bump the version anymore. Otherwise the patch looks good.

Conflict resolved.
Can you merge this or do I need to wait for someone else to review it? If yes, would you mind looking at my other PRs for equinox too (they are all trivial)?

akurtakov:
Rebased it on master so verification build should work now.

akurtakov:
Thanks!

PR #324, opened 2023-10-01
Apply project settings to o.e.osgi.util

Torbjorn-Svensson:
```shell
find . -type f -print0 | \
  xargs -rd'\n' dos2unix -e
```


akurtakov:
@tjwatson As a PL I would like to get your opinion on this(and other similar) mass changes..

Torbjorn-Svensson:
My aim with all these "cleanup" PRs is to make it trivial to do code changes that only contain the intended code change in the diff after applying the automatic formatting rules in the Eclipse IDE.
Right now, if you do a minor change in one of the bundles, the entire file is rewritten and the change is "lost" in the diff.

I have never worked on TCK before. Can you please enlighten me where I can read up on how it works as I can't connect my changes in this PR to how a unit test can fail.

akurtakov:
@laeubi Can you reply to the question about TCKs?

laeubi:
Some TCKs are not successful for Equinox, but ti seems there is not much people caring about the full compatibility of edge cases. So if there are no **new** failures it should be fine:

https://github.com/eclipse-equinox/equinox#implemented-services-an-compliance

beside that, instead of reformatting all code now there are the following alternatives:

1. before working on a file, reformat it and commit that change, then perform your change and add another commit, in the Github UI it is now possible to only look at that particular diff that changed the code
2. enable "only format edited files" option to only ever change/reformat the single portion of code where a change was applied.

Torbjorn-Svensson:
> Some TCKs are not successful for Equinox, but ti seems there is not much people caring about the full compatibility of edge cases. So if there are no **new** failures it should be fine:
> 
> https://github.com/eclipse-equinox/equinox#implemented-services-an-compliance

Ok, so I assume that PRs that does not fail on any other step than the TCKs are fine.

> beside that, instead of reformatting all code now there are the following alternatives:
> 
>     1. before working on a file, reformat it and commit that change, then perform your change and add another commit, in the Github UI it is now possible to only look at that particular diff that changed the code

To me, this sounds sub-optimal as it would require more work from contributors than just doing the change once and move on with other tasks.

>     2. enable "only format edited files" option to only ever change/reformat the single portion of code where a change was applied.

I suppose you talk about "only format edited lines" and yes, that could work, but it would make the code much harder to read. Personally, I do not consider this a solution.

HannesWell:
> To me, this sounds sub-optimal as it would require more work from contributors than just doing the change once and move on with other tasks.

It is, but it is also a lot of work to review all the changes :)

And please note as well, that in some cases Equinox is embedding external sources, e.g. OSGi API (in packages `org.osgi*`) or from Apache Felix (`org.apache.felix*`)

Torbjorn-Svensson:
> > To me, this sounds sub-optimal as it would require more work from contributors than just doing the change once and move on with other tasks.
> 
> It is, but it is also a lot of work to review all the changes :)

That's why I included the command that I ran to produce the change :)

> And please note as well, that in some cases Equinox is embedding external sources, e.g. OSGi API (in packages `org.osgi*`) or from Apache Felix (`org.apache.felix*`)

Ok, so in that case, it's it desired to have them follow the coding rules defined in the bundle or that they break whenever someone tries to correct a small typo or fix a bug in them?
Depending on the reply, I can exclude them from the PR(s) if I just get a list of trees that are embedded from external sources.

tjwatson:
> My aim with all these "cleanup" PRs is to make it trivial to do code changes that only contain the intended code change in the diff after applying the automatic formatting rules in the Eclipse IDE.

In the past we had setup project specific settings format on save action -> "Format edited lines" so it does not change the whole file.  Perhaps that got lost along the way?

Torbjorn-Svensson:
> > @tjwatson As a PL I would like to get your opinion on this(and other similar) mass changes..
> 
> We should only do this for `org.eclipse` source of the project (not `org.osgi or `org.apache` source). We should make sure the project settings are set to have "Format edited Lines" only on the save action.
> 
> I'm all for converting to LF and ensuring EOF though (for `org.eclipse` source). I thought we did that along time ago.

I've updated the PRs to not include what I believe is `org.osgi` and `org.apache` sources.

Let me know if there is something else I missed in the PRs.

akurtakov (approved):

PR #350, opened 2023-10-01
Apply project settings to o.e.equinox.log.stream

Torbjorn-Svensson:
```shell
find . -type f -print0 | \
  xargs -rd'\n' dos2unix -e
```

2. Format java files:
```shell
eclipse -consolelog -nosplash -application org.eclipse.jdt.core.JavaCodeFormatter \
  -config .settings/org.eclipse.jdt.core.prefs . -data `mktemp -d`
```


bundles/org.eclipse.equinox.log.stream/src/org/eclipse/equinox/internal/log/stream/LogStreamManager.java
```diff
@@ -54,7 +57,9 @@ public void start(BundleContext bc) throws Exception {
 
 	/*
 	 * (non-Javadoc)
-	 * @see org.osgi.framework.BundleActivator#stop(org.osgi.framework.BundleContext)
+	 *
```
vogella, line 60: The (non-Javadoc) comments pre-dates the @Override so (if desired) these could also be deleted in the future. I did this for most of the non-javadoc in platform a long time ago.

Not related to this change, but just wanted to point out this opportunity.

vogella:
LGTM thanks @Torbjorn-Svensson

bundles/org.eclipse.equinox.log.stream/src/org/eclipse/equinox/internal/log/stream/LogStreamManager.java
Torbjorn-Svensson: I know, it's on my todo list, but I though this was big enough change as it is ;-)

vogella: > I know, it's on my todo list, but I though this was big enough change as it is ;-)

Cool, thanks for considering this. I still find this non-javadoc annoying... :-)

PR #371, opened 2023-10-01
Apply project settings to o.e.equinox.common

Torbjorn-Svensson:
```shell
find . -type f -print0 | \
  xargs -rd'\n' dos2unix -e
```

2. Format java files:
```shell
eclipse -consolelog -nosplash -application org.eclipse.jdt.core.JavaCodeFormatter \
  -config .settings/org.eclipse.jdt.core.prefs . -data `mktemp -d`
```


vogella:
@Torbjorn-Svensson thanks for this work, I personally always like "clean-ups".

@tjwatson @akurtakov I saw (somethere) a discussion about these format changes flying by but can't find the reference again for them. What was the decision about these changes? We have a ton of them now in the queue.

vogella:
cc @HannesWell who (IIRC) also like clean-ups.

vogella:
Ah, "re-found" it, https://github.com/eclipse-equinox/equinox/pull/324#issuecomment-1748853543 Looks like @tjwatson is OK with these clean-ups

vogella:
It is hard to verify such large automatic clean-ups in detail. But as @Torbjorn-Svensson was using the standard Eclipse formatter I will give trust to the JDT formatter to not do something breaking, especially as the build is green.

The assumption here is that if the JDT formatter does something wrong it would do the same interactive and that any (formatting) error can be adjusted by follow-up commits. 

So unless someone steps up to review these changes line by line I will merge them.

Torbjorn-Svensson:
> It is hard to verify such large automatic clean-ups in detail. But as @Torbjorn-Svensson was using the standard Eclipse formatter I will give trust to the JDT formatter to not do something breaking, especially as the build is green.
> 
> The assumption here is that if the JDT formatter does something wrong it would do the same interactive and that any (formatting) error can be adjusted by follow-up commits.
> 
> So unless someone steps up to review these changes line by line I will merge them.

If you spot anything obvious when looking though them, please let me know, otherwise I would like all the green ones to be merged first and after that rebase the rest and see if that fixed some of the other PRs that are red.
If they are still red, then I will have to do some more work in order to identify why they fail.

Btw, I provided the actual commands used to produce these PRs just for the sake that it should be possible to verify that I didn't introduce any changes that is not due to the o.e.jdt.core.JavaCodeFormatter application.

laeubi:
> Btw, I provided the actual commands used to produce these PRs just for the sake that it should be possible to verify that I didn't introduce any changes that is not due to the o.e.jdt.core.JavaCodeFormatter application.

@Torbjorn-Svensson I think it would be interesting to have a build verification that checks for fromatings in source code something similar to [spotless-maven-plugin](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md) do you have any experience if spotless support eclipse and could probably be used here?

Torbjorn-Svensson:
> > Btw, I provided the actual commands used to produce these PRs just for the sake that it should be possible to verify that I didn't introduce any changes that is not due to the o.e.jdt.core.JavaCodeFormatter application.
> 
> @Torbjorn-Svensson I think it would be interesting to have a build verification that checks for formating in source code something similar to [spotless-maven-plugin](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md) do you have any experience in using [spotless eclipse support](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#eclipse-jdt) and could probably be used here?

Sorry no, I've never tried anything in maven for this.
What I have some experience of is to do this outside the maven system, mainly how it's done in Eclipse CDT (entry point is https://github.com/eclipse-cdt/cdt/blob/main/releng/scripts/do_all_code_cleanups.sh).

PR #377, opened 2023-10-04
Do not import o.e.jetty.servlet in server.jetty feature

akurtakov:
Bundle changed id to o.e.jetty.ee8.servlet and import it not needed as it's a dependency of o.e.eqinox.http.jetty which is included in the feature and thus the servlet bundle will come as transitive dependency.


merks:
It looks like you're making really good progress!  👍

akurtakov:
Merging as the real verification will be the aggregator build.

Issue #221, opened 2023-02-24
Possible performance improvement for high frequency logging using EquinoxLogWriter


alxflam-work:
Hi,

i've been taking a look at [EquinoxLogWriter](https://github.com/eclipse-equinox/equinox/blob/a9db8e97f40f2ae514722225133ab80fff367bb5/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/log/EquinoxLogWriter.java) recently and noticed the following behavior:

`log(FrameworkLogEntry logEntry)` always creates a new `BufferedWriter` with a new `FileOutputStream` for the given log file as the `finally` block in `log` always sets the `writer` to null (`closeFile` is called). If one makes heavy usage of logging, this can be a performance issue: Say you log a large number of `FrameworkLogEntry`s, then the FOS is opened and closed for every single log invocation.

Wouldn't it be possible to improve performance by only re-creating the `BufferedWriter` when necessary: 

- once the log file changes by use of the setter: `setOutput` is already called by `setFile`, therefore the `writer` is closed and set to null
- once the log file rotates due to the max size restriction: `checkLogFileSize` already calls `setOutput` and therefore ensures the `writer` is closed and set to null if the log file changes  

As both `setFile` and `log` are synchronized, wouldn't it be possible to omit the `closeFile` call in the `log` method to keep the `writer` instance and thereby improving performance for consecutive log invocations? Sure, then there's an open FOS for the runtime of the equinox application, e.g. for the complete runtime of an eclipse RCP. But isn't that somewhat expectable for the log file? A new `dispose` method may be required which can be called to free the `writer` and any other members, this could be called e.g. by `EquinoxLogServices#stop`.

What do you think - is this something that could be improved as outlined? I may just not spot the reason why the `writer` is not reused.

Greetings,
Alex

tjwatson:
The log writing code itself predates even Equinox.  IIRC the current may be necessary for things like the PDE log viewer so that it gets notified when the content of the file changes.  At the least we need to make sure the content is flushed to disk each entry write I think.

alxflam-work:
[...]

PR #382, opened 2023-10-19
do not close file writer upon every log invocation

alxflam-work:
Hi,

this PR resolves issue #221:

- The writer is no longer closed in a finally block upon every `log` invocation
- Instead, the writer is closed inside `log` only if an `Exception` is thrown, then the writer is closed immediately. Additionally, `closeFile` is called also after the writer changed to write to std.err in the `catch` block (else that writer instance would not be closed)
- Unchanged behavior: The `writer` is still closed, aka `closeFile` is called, when the actual log file changes (e.g. explicitly or due to log file rotation caused by size limit)
- I have added a new test `EquinoxLogWriterTest` to verify that logs are `flushed` immediately but the `writer` is not closed. I tried my best with the Junit4 environment and missing mockito, please let me know if i can improve the testcase. For now, my only solution was using reflection to actually verify the invocations on the `writer`.
- `All OSGi Tests` launch config runs locally without errors (with these changes)

Greetings,
Alex

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/log/EquinoxLogWriter.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/log/EquinoxLogWriter.java
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Copyright (c) 2023 Robert Bosch GmbH - https://github.com/eclipse-equinox/equinox/issues/221
  *******************************************************************************/
 package org.eclipse.osgi.internal.log;
 
@@ -84,6 +85,9 @@ class EquinoxLogWriter implements SynchronousLogListener, LogFilter {
 
 	/** The system property used to specify command line args should be omitted from the log */
 	private static final String PROP_LOG_INCLUDE_COMMAND_LINE = "eclipse.log.include.commandline"; //$NON-NLS-1$
+	/** The system property used to specify the log file writer should be closed eagerly */
+	private static final String PROP_CLOSE_LOG_FILE_EAGERLY = "eclipse.log.closeFile.eagerClose"; //$NON-NLS-1$
+
 	/** Indicates if the console messages should be printed to the console (System.out) */
 	private boolean consoleLog = false;
 	/** Indicates if the next log message is part of a new session */
@@ -110,6 +114,15 @@ class EquinoxLogWriter implements SynchronousLogListener, LogFilter {
 	private boolean includeCommandLine = true;
 
 	private LoggerAdmin loggerAdmin = null;
+
+	/**
+	 * Controls whether the log file should be closed eagerly upon every log
+	 * invocation.
+	 * 
+	 * Can be controlled by property
+	 * {@link EquinoxLogWriter#PROP_CLOSE_LOG_FILE_EAGERLY}.
+	 */
+	private boolean closeLogFileEagerly = false;
 
 	/**
 	 * Constructs an EclipseLog which uses the specified File to log messages to
@@ -303,6 +316,8 @@ class EquinoxLogWriter implements SynchronousLogListener, LogFilter {
 			writeLog(0, logEntry);
 			writer.flush();
 		} catch (Exception e) {
+			// close log file writer upon exceptions
+			closeFile();
 			// any exceptions during logging should be caught
 			System.err.println("An exception occurred while writing to the platform log:");//$NON-NLS-1$
 			e.printStackTrace(System.err);
@@ -315,9 +330,14 @@ class EquinoxLogWriter implements SynchronousLogListener, LogFilter {
 			} catch (Exception e2) {
 				System.err.println("An exception occurred while logging to the console:");//$NON-NLS-1$
 				e2.printStackTrace(System.err);
+			} finally {
+				// ensure that the error stream writer is closed
+				closeFile();
 			}
 		} finally {
-			closeFile();
+			if (closeLogFileEagerly) {
+				closeFile();
+			}
 		}
 	}
 
@@ -679,6 +699,13 @@ class EquinoxLogWriter implements SynchronousLogListener, LogFilter {
 
 		includeCommandLine = "true".equals(environmentInfo.getConfiguration(PROP_LOG_INCLUDE_COMMAND_LINE, "true")); //$NON-NLS-1$//$NON-NLS-2$
 		applyLogLevel();
+
+		String newCloseLogFileEagerlyValue = environmentInfo.getConfiguration(PROP_CLOSE_LOG_FILE_EAGERLY);
+		if (newCloseLogFileEagerlyValue != null) {
+			if (Boolean.valueOf(newCloseLogFileEagerlyValue) == Boolean.TRUE) {
+				closeLogFileEagerly = true;
+			}
+		}
 	}
 
 	void applyLogLevel() {
--- a/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Core OSGi Tests
 Bundle-SymbolicName: org.eclipse.osgi.tests;singleton:=true
-Bundle-Version: 3.18.500.qualifier
+Bundle-Version: 3.18.600.qualifier
 Bundle-Vendor: Eclipse.org
 Require-Bundle: 
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
--- a/bundles/org.eclipse.osgi.tests/pom.xml
+++ b/bundles/org.eclipse.osgi.tests/pom.xml
@@ -19,7 +19,7 @@
 </parent>
   <groupId>org.eclipse.osgi</groupId>
   <artifactId>org.eclipse.osgi.tests</artifactId>
-  <version>3.18.500-SNAPSHOT</version>
+  <version>3.18.600-SNAPSHOT</version>
   <packaging>eclipse-test-plugin</packaging>
 
   <properties>
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/equinox/log/test/AllLogServiceTests.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/equinox/log/test/AllLogServiceTests.java
@@ -15,9 +15,10 @@ import org.junit.runners.Suite;
 
 @RunWith(Suite.class)
 @Suite.SuiteClasses({ //
-	LogServiceTest.class, //
-	LogReaderServiceTest.class, //
-	LogPermissionCollectionTest.class //
-	})
+		LogServiceTest.class, //
+		LogReaderServiceTest.class, //
+		LogPermissionCollectionTest.class, //
+		EquinoxLogWriterTest.class //
+})
 public class AllLogServiceTests {
 }
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/equinox/log/test/EquinoxLogWriterTest.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/equinox/log/test/EquinoxLogWriterTest.java
@@ -1,0 +1,182 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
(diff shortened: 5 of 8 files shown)


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/log/EquinoxLogWriter.java
```diff
@@ -315,9 +318,10 @@ private synchronized void log(FrameworkLogEntry logEntry) {
 			} catch (Exception e2) {
 				System.err.println("An exception occurred while logging to the console:");//$NON-NLS-1$
 				e2.printStackTrace(System.err);
+			} finally {
+				// ensure that the error stream writer is closed
+				closeFile();
 			}
-		} finally {
```
tjwatson, line 319: Do we ensure the writer is closed when the framework stops?
I think we need an option to preserve the old behavior also.

alxflam-work: > Do we ensure the writer is closed when the framework stops?

Actually, i thought the already existing `close` method would be called. But indeed it is not - it is called by the wrapping `FrameworkLog` but that service is manually registered, and it is never actually invoked.  Instead, i have now added calls to `close` to the `EquinoxLogWriter` instances  in the `stop` method of `EquinoxLogServices` - that works, but it would have been nicer if the `FrameworkLog` service had a lifecycle method for disposal (as with DS @Deactivate, but the service is manually registered).

> I think we need an option to preserve the old behavior also.

What kind of configuration mechanism would you suggest (VM Argument)? And what would be your expected default behavior (default = old behavior or default = new behavior)?

tjwatson: > i have now added calls to close to the EquinoxLogWriter instances in the stop method of EquinoxLogServices - that works, but it would have been nicer if the FrameworkLog service had a lifecycle method for disposal (as with DS @deactivate, but the service is manually registered).

Thanks, I think this is fine for a solution here.  The framework cannot depend on DS for service management.

alxflam-work: @tjwatson thanks, i have now added a new member `PROP_CLOSE_LOG_FILE_EAGERLY` for the new property. I've added it to `EquinoxLogWriter` instead of `EquinoxConfiguration` as all the other related log file properties are also declared in there.

tjwatson (approved):

tjwatson (commented):
One more thing.  Unfortunately this is the first change for the 2023-12 release in `org.eclipse.osgi` so we need a version bump to `Bundle-Version: 3.18.600.qualifier` in the `META-INF/MANIFEST.MF` and to `<version>3.18.600-SNAPSHOT</version>` for the `pom.xml`

Also need to update the versions in `org.eclipse.osgi.test` project.

alxflam-work:
Thanks for the hints, updated the versions, rebased and squashed commits.

tjwatson:
Thanks for the PR.  It is merged now.

iloveeclipse:
Please take a look on new 5 fails in runtime tests

- https://download.eclipse.org/eclipse/downloads/drops4/I20231024-1800/testresults/html/org.eclipse.core.tests.runtime_ep430I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

and 7 fails in p2 tests
- https://download.eclipse.org/eclipse/downloads/drops4/I20231024-1800/testresults/html/org.eclipse.equinox.p2.tests_ep430I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

I believe they are all caused by this PR.

alxflam-work:
@iloveeclipse @tjwatson the reason these tests are failing is probably cause they all try to delete the frameworks log file which is now locked as there is still an open `writer` by the `EquinoxLogWriter`. Actually, that could have already been previously cause for toggling behavior (Test tries to delete the file, another thread is concurrently invoking `log`, therefore creating the `writer` and thereby preventing the deletion).
I would suggest to go ahead like this:

All tests requiring to work with a "clean" log file for easier log file content verification should simply call `FrameworkLog#setFile` and set some temporary file. Then the test can verify the contents of this explicit test log file. In the test cleanup, the original log file can be reset on the framework log. Another option is to simply overwrite the log file contents with an empty String (which is pretty similar to deleting the file which is the current attempt in the tests), which works but does not seem as good to me (what if you want the framework log for analysis later on for a failed test but now the content is gone). What do you think?

iloveeclipse:
I honestly have no time to look into this, but it would be nice to fix the test fails soon in some way, as they prevent people to get green platform builds on PR's.

laeubi:
> All tests requiring to work with a "clean" log file for easier log file content verification should simply call `FrameworkLog#setFile` and set some temporary file.

This sounds reasonable, @alxflam-work can you provide a PR for the failing tests to do so?

alxflam-work:
> This sounds reasonable, @alxflam-work can you provide a PR for the failing tests to do so?

@laeubi i'd really like to do so, but the situation for me is the following: I only have approval from our internal OSS boards to contribute to the equinox repository. As the failing tests reside in the p2 and eclipse.platform repositories, i'll have to go through some corporate OSS process to get these repositories approved for contributions. Sadly, this can take quite some time. So in the end if i fix the failing tests this will simply take potentially many weeks due to our internal OSS process, which i guess is not acceptable. Therefore, either someone of the eclipse members provides the fix or we revert this PR and we can rework it once i have approval for p2 and eclipse.platform repository contributions? Sorry for the inconveniences caused!

What works locally for the P2 repository failing tests ArtifactMirrorApplicationTest & NewMirrorApplicationArtifactTest is this:

- remove the `delete` invocations for the log file
- in `setUp` add the following code:
`FrameworkLog log = ServiceHelper.getService(Activator.getContext(), FrameworkLog.class);
originalFrameworkLog = log.getFile();
var tmpLogFile = File.createTempFile(this.getClass().getSimpleName(), null);
tmpLogFile.deleteOnExit();
log.setFile(tmpLogFile, false);
`
- in `tearDown` add the following code:
`FrameworkLog log = ServiceHelper.getService(Activator.getContext(), FrameworkLog.class);
log.setFile(originalFrameworkLog, false);
`

The same should work also for the failing test in platform LogSerializationTest, but maybe there junit4 is used and the TemporaryFolder rule could be applied instead.

jukzi:
@tjwatson if nobody works on fixing the tests please revert, it's very frustrating that whole platform is blocked by the current state.

iloveeclipse:
> The same should work also for the failing test in platform LogSerializationTest

It doesn't work. The test deletes the log file / logs in the loop. I believe the change here broke that.
My attempt to follow your proposal is:

```
diff --git a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/runtime/LogSerializationTest.java b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/runtime/LogSerializationTest.java
index 8a9991e..8690403 100644
--- a/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/runtime/LogSerializationTest.java
+++ b/runtime/tests/org.eclipse.core.tests.runtime/src/org/eclipse/core/tests/internal/runtime/LogSerializationTest.java
@@ -15,7 +15,16 @@
 
-import java.io.*;
+import java.io.File;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.nio.file.Files;
 import java.util.Arrays;
 import org.eclipse.core.internal.runtime.RuntimeLog;
-import org.eclipse.core.runtime.*;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.MultiStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.tests.runtime.RuntimeTest;
+import org.eclipse.core.tests.runtime.RuntimeTestsPlugin;
+import org.eclipse.osgi.framework.log.FrameworkLog;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceReference;
 
@@ -32,3 +41,4 @@
 
-	protected File logFile = null;
+	protected File logFile;
+	private File originalFrameworkLog;
 
@@ -152,3 +162,3 @@
 
-	protected void doTest(String msg, IStatus[] oldStats) {
+	protected void doTest(String msg, IStatus[] oldStats) throws Exception {
 		writeLog(oldStats);
@@ -158,3 +168,3 @@
 
-	protected void doTest(String msg, IStatus status) {
+	protected void doTest(String msg, IStatus status) throws Exception {
 		doTest(msg, new IStatus[] {status});
@@ -171,5 +181,7 @@
 		//setup the log file
-		if (logFile == null) {
-			logFile = Platform.getLogFileLocation().toFile();
-		}
+		FrameworkLog log = getService(RuntimeTestsPlugin.getContext(), FrameworkLog.class);
+		originalFrameworkLog = log.getFile();
+		logFile = File.createTempFile(this.getClass().getSimpleName(), null);
+		logFile.deleteOnExit();
+		log.setFile(logFile, false);
 	}
@@ -179,6 +191,21 @@
 		super.tearDown();
-		logFile.delete();
+		FrameworkLog log = getService(RuntimeTestsPlugin.getContext(), FrameworkLog.class);
+		log.setFile(originalFrameworkLog, false);
+		Files.delete(logFile.toPath());
 	}
 
-	public void testDeepMultiStatus() {
+	public static <T> T getService(BundleContext context, Class<T> clazz) {
+		if (context == null) {
+			return null;
+		}
+		ServiceReference<T> reference = context.getServiceReference(clazz);
+		if (reference == null) {
+			return null;
+		}
+		T result = context.getService(reference);
+		context.ungetService(reference);
+		return result;
+	}
+
+	public void testDeepMultiStatus() throws Exception {
 		MultiStatus multi = new MultiStatus("id", 1, getInterestingMultiStatuses(), "ok", null);
@@ -190,3 +217,3 @@
 
-	public void testMultiMultiStatusSerialize() {
+	public void testMultiMultiStatusSerialize() throws Exception {
 		IStatus[] interesting = getInterestingMultiStatuses();
@@ -200,3 +227,3 @@
 
-	public void testMultiSerialize() {
+	public void testMultiSerialize() throws Exception {
 		IStatus[] interesting = getInterestingStatuses();
@@ -210,3 +237,3 @@
 
-	public void testMultiStatus() {
+	public void testMultiStatus() throws Exception {
 		IStatus[] interesting = getInterestingMultiStatuses();
@@ -217,3 +244,3 @@
 
-	public void testSimpleSerialize() {
+	public void testSimpleSerialize() throws Exception {
 		IStatus[] interesting = getInterestingStatuses();
@@ -224,3 +251,3 @@
 
-	protected void writeLog(IStatus status) {
+	protected void writeLog(IStatus status) throws Exception {
 		writeLog(new IStatus[] {status});
@@ -228,5 +255,5 @@
 
-	protected void writeLog(IStatus[] statuses) {
+	protected void writeLog(IStatus[] statuses) throws Exception {
 		if (logFile.exists()) {
-			logFile.delete();
+			Files.delete(logFile.toPath());
 		}
```

merks:
For what it helps, note that there is also a system property to disable the new behavior and restore the old behavior...

iloveeclipse:
> For what it helps, note that there is also a system property to disable the new behavior and restore the old behavior...

The question for me is if that new behavior is working properly.

> the reason these tests are failing is probably cause they all try to delete the frameworks log file which is now locked as there is still an open writer by the EquinoxLogWriter

If that is not working now, I would say it is a regression. The test LogSerializationTest deletes log file, logs something and expect that the content they log appears in the **same** file. This makes sense, since that is what can happen at any time in a regular environment too. If that is not working anymore, and the log file contains something different / not what the application has most recently logged, looks like a regression.

Long story short, I would propose a PR reverting the change, and let @tjwatson & @alxflam-work discuss if the new behavior works as expected, shouldn't be default or just has some implementation bug.

iloveeclipse:
Revert PR: https://github.com/eclipse-equinox/equinox/pull/388

tjwatson:
Revert was the right thing to do, although the details on if the file is closed or not are an implementation detail, not an API for the log file.  @alxflam-work you will either need to fix the tests, or more likely flip the default to be the old behavior.

alxflam-work:
@tjwatson then i'll provide a PR with the same changes, but defaulting to the old behavior. Once that is done, i can adapt the tests to no longer delete the log file but instead use a dedicated tmp log file - which should not be a semantic change for the tests. The tests should then still be green and afterwards we could flip the default behavior.

jukzi:
@alxflam-work if you propose a new PR please test the mentioned Tests before. It's not enough to assume they would work.

PR #386, opened 2023-10-23
Remove the org.eclipse.equinox.server.simple feature

merks:
This appears not to be mentioned/used anywhere.
Further it appears that the last time it was in any p2 repository was for indigo.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1361


merks:
This is the only trace I can find of it historically:

image

merks:
@akurtakov  @HeikoKlare

It appears to me that the build itself is kind of broken.  I see mention of org.eclipse.test.performance and I've seen mention of changes to that over the weekend...

It seems we're rather hosed at the moment.  😱

merks:
@akurtakov 

Given that the build itself was okay, other than the tests that failed because the tests cannot launch, I'm tempted to just merge this change so that I can move on from it.  Is that okay or is that evil and wrong?

akurtakov:
It's not nice for sure and if it was code change I would probably have voted otherwise but for this one it's best to proceed and merge IMHO.

HeikoKlare:
@merks I think you are right and the build is broken due to https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1454

The log says:
`Unresolved requirement: Import-Package: org.junit.platform.suite.api`

And the mentioned PR introduced exactly that package import. So is JUnit 5, more particularly JUnit platform, not part of the TP  used for the tests here?
I will try to have a look as soon as possible. As a quick fix, I could also change `org.eclipse.test.performance` again to use JUnit 4 instead of JUnit 5.

akurtakov:
If it's about quickfix I would rather see see  suite api added to the failing tests bundle rather than reverting test.performance change.

merks:
I agree.  Let's not revert things but rather let's move forward toward progress.

HeikoKlare:
If I see correctly, the TP used for the Maven build is `tcks.target`: https://github.com/eclipse-equinox/equinox/blob/master/releng/tcks.target

The JUnit dependencies come from [`org.osgi:org.osgi.test.cases.framework`](https://central.sonatype.com/artifact/org.osgi/org.osgi.test.cases.framework). This module includes some `junit-platform-*` modules, but not the `junit-platform-suite-api` . I did not find any `org.osgi.test.*` module so far that depends on the suite API. So should we add the `junit-platform-suite-api` one to the `tcks.target`? Looks a bit inconsistent to me to rely on `org.osg.test.*` dependencies in that target definition but to add one specific, separate JUnit dependency.

akurtakov:
@laeubi I guess you're the most knowlegable to answer ^^^.

laeubi:
If it helps adding stuff to the `tcks.target` seems suitable and as we see this error in the verification builds its easy to try out :-)

HeikoKlare:
I have to slightly correct myself: I am not sure whether the failing tests are actually executed against the `tcks.target`. That one is activated via profile to be (obviously) only used for the TCK jobs. The usual Maven build should use the TP defined by the parent, i.e., the `eclipse-sdk-prereqs` one, which contains the `junit-platform-suite-api`. So not sure whether the tests will actually be affected by changing the `tcks.target`, but as @laeubi mentioned, we could try out 🙂

HeikoKlare:
I think I found the cause: the tests use a `ConfigurationSessionTestSuite` (as part of the platform test harness), which defines the bundles to be used for test execution programmatically. The bundle `org.eclipse.performance.test` is added, but due to the dependency change of that bundle, not all transitive dependencies (i.e., `junit-platform-suite-api`) are available.

I have corrected that test suite in https://github.com/eclipse-platform/eclipse.platform/pull/758. Locally, the equinox tests run well with that change (and failed without it, like in Maven CI build).

PR #387, opened 2023-10-23
Remove jakarta.servlet-api, javax.servlet.jsp-api, javax.el-api

merks:
These bundles are included by features but are not actually needed at runtime.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1361


merks:
@akurtakov 

Of course we have the same build/test problem.

Note though that the servlet tests pass:

image

Also, it looks like the only thing that test the jasper support are the ua tests:

image

So perhaps it's okay to be evil and press forward on this change as well?

Issue #18, opened 2022-04-18
Replace embedded osgi packages by osgi-bundles from Maven-Central


HannesWell:
In M2E we would like to use `bnd.lib` 6.1 respectively 6.2 (see https://github.com/eclipse-m2e/m2e-core/pull/445) which imports newer version of the following OSGi packages than those that are embedded into `org.eclipse.osgi` respectively `org.eclipse.osgi.util`:

- `org.osgi.util.function` 1.2.0 (available 1.1.0)
- `org.osgi.util.promise` 1.2.0 (available 1.1.1)
- `org.osgi.util.tracker` 1.5.4 (available 1.5.3, which is actually sufficient because bnd.lib requires only 1.5)

Furthermore it requires the package the following package that is not embedded into any equinox bundle (at least I did not find it):
- `org.osgi.service.repository` 1.1.0

I assume it is desired to update those packages that are already available to the latest version?
Furthermore I assume to update them I have to copy the sources from the OSGi repository at the latest final release to the corresponding bundle in equinox:
https://github.com/osgi/osgi/releases/tag/r8-core-final-rerelease

How should we handle the missing `org.osgi.service.repository`?
M2E does not use the repository functionality (AFAICT) but it is required to resolve the bundle.
Should we fetch it from Maven-Central (and make it part of the M2E p2-repo and consequently contribute it to SimRel) or should it be embedded into an Equinox bundle too?

@tjwatson could you please provide some advice?

HannesWell:
An update summary about the main PRs associated with this issue:
- https://github.com/eclipse-equinox/equinox.framework/pull/41
- https://github.com/eclipse-equinox/equinox/pull/30
- https://github.com/eclipse-equinox/equinox.bundles/pull/26
- https://github.com/eclipse-equinox/equinox.bundles/pull/39

Additionally there were multiple PRs to include the OSGi bundles into those features that contain the equinox-bundles that used to embed the replaced packages (if one is interested they are all referring this issue).

The only packages that could not have been replaced yet are the packages from `org.osgi.service.http` and `org.osgi.service.http.whiteboard` because they require a capability provided `org.apache.felix.http.servlet-api`. But its inclusion is currently blocked by https://github.com/eclipse-equinox/p2/pull/64.

akurtakov:
This has caused javadoc generation failure. https://download.eclipse.org/eclipse/downloads/drops4/I20220618-1800/compilelogs/platform.doc.isv.javadoc.txt

[...]

PR #394, opened 2023-11-04
Update embedded org.osgi.util.tracker sources from 1.5.3 to 1.5.4

HannesWell:
As requested in https://github.com/eclipse-equinox/equinox/issues/18#issuecomment-1793367606 (altough this does not resolve the entire issue) update embedded org.osgi.util.tracker sources from 1.5.3 to 1.5.4.

@tjwatson I simply copied the java files from https://repo1.maven.org/maven2/org/osgi/org.osgi.util.tracker/1.5.4/org.osgi.util.tracker-1.5.4-sources.jar and replaced the existing sources with that.

But since this mainly changes license headers and SPDX-License-Identifier, I wonder if this is the right way?
Btw. to we need a CQ for this then (it is already Eclipse licensed and from another Eclipse project)?

### DIFF
--- a/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF
@@ -77,7 +77,7 @@ Export-Package: org.eclipse.core.runtime.adaptor;x-friends:="org.eclipse.core.ru
  org.osgi.service.resolver;version="1.1.1";uses:="org.osgi.resource",
  org.osgi.service.startlevel;version="1.1.1";uses:="org.osgi.framework",
  org.osgi.service.url;version="1.0.1",
- org.osgi.util.tracker;version="1.5.3";uses:="org.osgi.framework"
+ org.osgi.util.tracker;version="1.5.4";uses:="org.osgi.framework"
 Provide-Capability: osgi.service; objectClass:List<String>="org.osgi.service.log.LogReaderService,org.eclipse.equinox.log.ExtendedLogReaderService"; uses:="org.osgi.service.log",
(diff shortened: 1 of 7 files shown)


tjwatson:
We shouldn't need a CQ for this.

HannesWell:
> We shouldn't need a CQ for this.

OK good, so is the change fine in total?

merks:
FYI, in the dash logs for Obit I see an entry like this:

> maven/mavencentral/org.osgi/org.osgi.util.tracker/1.5.4, Apache-2.0, approved, clearlydefined

Torbjorn-Svensson:
Pardon a stupid question, but why do we need to duplicate the sources rather than have a dependency on the `org.osgi.util.tracker` delivery?

tjwatson:
> Pardon a stupid question, but why do we need to duplicate the sources rather than have a dependency on the `org.osgi.util.tracker` delivery?

The goal is to have an `org.eclipse.osgi` JAR that doesn't need any other JARs on the classpath to run.  That goal could be achieve any number of ways.  Historically it was most simple to just add the source to the project.  Other motivations for tracker in particular was that frameworks were free to make optimizations to the implementation and integrate it more closely to the framework's service registry implementation.  

I don't believe that we (nor felix) actually modify the tracker package anymore, but there are other classes in the OSGi Core specification that are modified (only in the implementation, otherwise we would fail signature tests from the OSGi compliance tests).  Both Felix and Equinox include the Core specification source in our projects:  https://github.com/apache/felix-dev/tree/master/framework/src/main/java/org/osgi

tjwatson (approved):

HannesWell:
Thanks Tom and Ed.

> I don't believe that we (nor felix) actually modify the tracker package anymore,

I assume that the Framework bundle cannot/musst not import packages from other bundles? So if the sources are not embedded the OSGi jar would have to be embedded.
I think that would be possible, but would probably require adjustments in the equinox Launcher since IIRC it does consider the Bundle-Classpath.
And in the end I see not much benefit from that at the moment.

PR #397, opened 2023-11-08
Encourage on rebase on forks by configuring the clone default for that

merks:
https://github.com/eclipse-jdt/eclipse.jdt/issues/79


merks:
The build is not affected by setup changes.  Cancel them.

Issue #399, opened 2023-11-10
git status / git rebase broken on master


iloveeclipse:
Regression from https://github.com/eclipse-equinox/equinox/commit/7ab85da7bbba52328168b980a616e37a82ee6f11.

See comment on https://github.com/eclipse-equinox/equinox/commit/7ab85da7bbba52328168b980a616e37a82ee6f11#r132204044

This change (and similar changes in all files in this commit) break command line git.
Checkout latest state and call `git status`

```
git status
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   bundles/org.eclipse.osgi/osgi/src/org/osgi/util/tracker/AbstractTracked.java
        modified:   bundles/org.eclipse.osgi/osgi/src/org/osgi/util/tracker/BundleTracker.java
        modified:   bundles/org.eclipse.osgi/osgi/src/org/osgi/util/tracker/BundleTrackerCustomizer.java
        modified:   bundles/org.eclipse.osgi/osgi/src/org/osgi/util/tracker/ServiceTracker.java
        modified:   bundles/org.eclipse.osgi/osgi/src/org/osgi/util/tracker/ServiceTrackerCustomizer.java
        modified:   bundles/org.eclipse.osgi/osgi/src/org/osgi/util/tracker/package-info.java
```

The problem is that git expands `$Id$` if `.gitattributes` contains `ident` attribute (and this is the case)...

This prevents me from rebasing any branch on master because git wants to undo "changes" first - but it fails to do so... https://stackoverflow.com/questions/21990758/way-to-disable-git-attribute-id-filter

iloveeclipse:
Fixed via https://github.com/eclipse-equinox/equinox/pull/400

PR #400, opened 2023-11-10
Remove ident ID value

tjwatson:


iloveeclipse:
Link to the issue: https://github.com/eclipse-equinox/equinox/issues/399

vogella:
PMC approved (in case you need that)

iloveeclipse (approved):

HannesWell:
Thanks Tom.

Indeed I used EGit and had no issue with that.
I just copied the original sources from Maven-Central.

iloveeclipse:
Unfortunately the SDK build failed, see  https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1534

iloveeclipse:
@tjwatson: Is there any reason to keep .gitattributes that cause such confusion?
Should we remove the ident sections from https://github.com/eclipse-equinox/equinox/blob/master/.gitattributes?
See git manual: https://git-scm.com/docs/gitattributes#:~:text=ident,Id%24%20upon%20check-in.

tjwatson:
> @tjwatson: Is there any reason to keep .gitattributes that cause such confusion?

We can consider removing the `ident` value from the `.gitattributes`.  It only is an issue for the osgi source we include in our repo since we don't use `$Id$` in `org.eclipse` source.  But it will mean we either checkin source that has the Id filled in or it will be blank.

iloveeclipse:
> But it will mean we either checkin source that has the Id filled in or it will be blank.

What is the problem with blank id? It is comment anyway, and I doubt the commit hash is that important for the sources shipped with a specific bundle version.

tjwatson:
> What is the problem with blank id? It is comment anyway, and I doubt the commit hash is that important for the sources shipped with a specific bundle version.

I'm not suggesting it is a problem, only that we have to decide.  In the past you could diff the equinox source with the osgi source repo and they would show no differences because the Id would get filled in correctly on checkout.  If we use a blank Id then the diff that I usually run to make sure I got the right source will show a difference.  In the end it is unfortunate that we have to work around a many year old bug in jGit.  This long running bug has simply got me use to never using eGit to commit anything, ever.

PR #403, opened 2023-11-18
Replace embedded service.http sources in org.eclipse.osgi.services by original bundles and deprecate o.e.osgi.services

HannesWell:
This replaces the remaining embedded `org.osgi.service.http*` packages in `org.eclipse.osgi.services` by the original bundles from Maven-Central and then marks `org.eclipse.osgi.services` as deprecated.
Fixes https://github.com/eclipse-equinox/equinox/issues/18.

As suggested in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239#issuecomment-1523821075:
> This is a sort of hack, but to use the existing `jakarta.servlet:jakarta.servlet-api` artifact could we also deploy a fragment to ` jakarta.servlet:jakarta.servlet-api` that adds the `osgi.contract` capability we need? If I recall correctly that was the missing piece to allow the osgi JARs to work for providing the http service APIs.

This adds a fragment `org.eclipse.osgi.service.http.servlet.contract` to supply the capability: `osgi.contract;osgi.contract=JavaServlet`, required by `org.osgi.service.http.whiteboard` to the known `javax.servlet` providing OSGi bundles from Maven-Central:
- [`javax.servlet:javax.servlet-api`](https://mvnrepository.com/artifact/javax.servlet/javax.servlet-api)
- [`jakarta.servlet:jakarta.servlet-api:[4.0.2-4.0.4]`](https://mvnrepository.com/artifact/jakarta.servlet/jakarta.servlet-api)
- [`org.eclipse.jetty.toolchain/jetty-servlet-api`](https://mvnrepository.com/artifact/org.eclipse.jetty.toolchain/jetty-servlet-api)

[`javax.servlet:servlet-api`](https://mvnrepository.com/artifact/javax.servlet/servlet-api) does not provide OSGi metadata.

One non-optimal thing with the different `javax.servlet` provides (except the ones from jetty) is that in the version 4 stream all artifacts have version 4.0 regardless of the BSN or their Maven-version change.
However after a quick/coarse check it looks like they only differences are white-space or legal text changes.

Background about OSGi contracts: https://docs.osgi.org/reference/portable-java-contracts.html

Another greater problem is that `org.osgi.service.http` has in its Manifest:
```
Import-Package:
 javax.servlet;version="[2.1,3)",
 javax.servlet.http;version="[2.1,3)"
```
In order to fix that I also added corresponding exports of `javax.servlet(.http)` in version `2.999`  to `org.eclipse.osgi.service.http.servlet.contract` that are served by the host. I compared all of the `javax.servlet` providing artifacts mentioned above and all content in  `javax.servlet` or  `javax.servlet.http` contained in [javax.servlet:servlet-api:2.2](https://mvnrepository.com/artifact/javax.servlet/servlet-api/2.2) is also provided by the other providers.


bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,17 @@
...
+ jakarta.servlet-api;bundle-version="[4.0.0,5.0.0)",
+ javax.servlet-api;bundle-version="[3.0.0,5.0.0)",
+ org.eclipse.jetty.servlet-api;bundle-version="[4.0.0,5.0.0)"
+Automatic-Module-Name: org.eclipse.osgi.service.http.servlet.contract
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Provide-Capability: osgi.contract;osgi.contract=JavaServlet
+ ;version:List<Version>="2.6,3.0,3.1,4.0"
+ ;uses:="javax.servlet,javax.servlet.http,javax.servlet.descriptor,javax.servlet.annotation"
+Export-Package: 
+ javax.servlet;version="2.999",
```
HannesWell, line 16: The real providers have
```suggestion
 javax.servlet;version="2.999";uses:="javax.servlet.annotation,javax.servlet.descriptor",
```
But the two used packages do not exist in javax.servlet 2.2

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,17 @@
...
+Bundle-SymbolicName: org.eclipse.osgi.service.http.servlet.contract;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Fragment-Host: 
+ jakarta.servlet-api;bundle-version="[4.0.0,5.0.0)",
+ javax.servlet-api;bundle-version="[3.0.0,5.0.0)",
+ org.eclipse.jetty.servlet-api;bundle-version="[4.0.0,5.0.0)"
+Automatic-Module-Name: org.eclipse.osgi.service.http.servlet.contract
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Provide-Capability: osgi.contract;osgi.contract=JavaServlet
+ ;version:List<Version>="2.6,3.0,3.1,4.0"
```
HannesWell, line 13: According to https://docs.osgi.org/reference/portable-java-contracts.html the `osgi.contract=JavaServlet` is specified with version `2.5`, `3.0`, `3.1` and `4.0`. 
But [`org.apache.felix:org.apache.felix.http.servlet-api:1.2.0`](https://mvnrepository.com/artifact/org.apache.felix/org.apache.felix.http.servlet-api/1.2.0) defines 2.6 instead of 2.5.

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,17 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JavaServlet OSGi Contract Provider for Equinox
+Bundle-SymbolicName: org.eclipse.osgi.service.http.servlet.contract;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Fragment-Host: 
+ jakarta.servlet-api;bundle-version="[4.0.0,5.0.0)",
+ javax.servlet-api;bundle-version="[3.0.0,5.0.0)",
+ org.eclipse.jetty.servlet-api;bundle-version="[4.0.0,5.0.0)"
```
HannesWell, line 9: Should we support only `javax/jakarta.servlet-api`or also jetty-servlet-api?

HannesWell:
In general to discuss the path forward we should also consider @tjwatson remarks from https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239#issuecomment-1523894295:

> > I also wonder how long will Equinox will have to support javax.servlet?
> 
> That is a complicate question.
> 
>     1. there is no http service spec that will be updated to the jakarta namespace, only the servlet whiteboard spec is updated to jakarta.  So we have to deprecate and remove the implementation of the http service implementation
> 
>     2. the javax.servlet API is in other APIs of ours like the http.registry.  So we have to deprecate some of that and replace parts that use jakarta namespace
> 
>     3. all current users of javax.servlet have to migrate their code to jakarta also
> 
> 
> These are not things we can easily do. And that is just from a deprecation breaking POV. I've not mentioned the actual work of changing our own codebase to use jakarta and who will be dedicated to doing that work.

HannesWell:
The build has problems to resolve the javax.servlet package exported by the fragment, but also in the IDE workspace (at least after a restart) the import
`Import-Package: javax.servlet;version="[2.1,3)"` is not resolved while `Import-Package: javax.servlet.http;version="[2.1,3)"` is.
I'm puzzled why this is the case.

Maybe it would just be simpler to use `org.apache.felix:org.apache.felix.http.servlet-api:1.2.0` as suggested in 
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239.

Another alternative @laeubi suggested (in a private chat), to leverage the missing metadata of [javax.servlet:servlet-api:2.5](https://mvnrepository.com/artifact/javax.servlet/servlet-api/2.5) and just generate them suitably using the new Orbit SimRel infrastructure created by Ed. The instructions could be prototyped in the sdk-target.

laeubi:
@HannesWell thanks for the efforts and detailed analysis.

I think it makes no sense to use an "official" artifact if when we need extra efforts to use it like additional fragments and alike, this defeats the purpose of it.

For me the proposal made here to reuse the felix bundled version:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239

seem much more appropriate because of the following reasons:

1. It is proven to work in OSGi and the felix-http has a very active and vibrant development and is the de factor standard
2. It is very unlikely that Equinox will ever get a comparable implementation, so as a matter of facts we should go the way we did with DS and try to migrate Equinox/Platform to the felix impl, using their servlet API package seems like a first simple and non intrusive step forward.

WDYT @akurtakov @tjwatson ?

HannesWell:
> I think it makes no sense to use an "official" artifact if when we need extra efforts to use it like additional fragments and alike, this defeats the purpose of it.
> 
> For me the proposal made here to reuse the felix bundled version:
> 
>     * [Add org.apache.felix:org.apache.felix.http.servlet-api:1.2.0 eclipse-platform/eclipse.platform.releng.aggregator#239](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239)
> 
> 
> seem much more appropriate because of the following reasons:
> 
>     1. It is proven to work in OSGi and the felix-http has a very active and vibrant development and is the de factor standard
> 
>     2. It is very unlikely that Equinox will ever get a comparable implementation, so as a matter of facts we should go the way we did with DS and try to migrate Equinox/Platform to the felix impl, using their servlet API package seems like a first simple and non intrusive step forward.
> 

Yes, with all the issues that arose on the way (the CI and WS build is still failing/has issues a.t.m.), I'm more and more inclined towards that solution as well. Especially, as you said, the two specs have probably not a big future.

HannesWell:
@tjwatson could you please share your assessment on how we should proceed here?

tjwatson:
> @tjwatson could you please share your assessment on how we should proceed here?

I think migrating to the felix http/whiteboard implementation sounds interesting.  But it has a myriad of issues with respect to extensions our implementation has on top of the spec that we either have to deprecate and eventually remove (before we can use the felix imple) or we have to figure out how to implement them on top of the felix implemetnation.

I think we should first and foremost focus on how to replace the service.http sources from the osgi.services bundle.

Perhaps we just add the capability `osgi.contract;osgi.contract=JavaServlet` to the `org.eclipse.equinox.http.servlet` bundle since it imports the necessary servlet packages so can ensure the contract is there?  That should allow the OSGi artifact for http.service to resolve.

tjwatson:
> I just wanted to note that `org.osgi.service.http` does _NOT_ require the contract at all, it is the `org.osgi.service.http.whiteboard` what is a different service and as far as I know was never provided by equinox.

We do provide an implemention of `org.osgi.service.http.whiteboard` in `org.eclipse.equinox.http.servlet`

HannesWell:
> > @tjwatson could you please share your assessment on how we should proceed here?
> 
> I think migrating to the felix http/whiteboard implementation sounds interesting. But it has a myriad of issues with respect to extensions our implementation has on top of the spec that we either have to deprecate and eventually remove (before we can use the felix imple) or we have to figure out how to implement them on top of the felix implemetnation.
>

I don't want to and cannot judge about replacing the implementation, but this PR is only about replacing the embedded http/whiteboard API sources with the OSGi bundles from bundles from Maven-Central.

But since the original http/whiteboard  OSGi bundles have some difficult dependencies (as desribed in the initial comment of this PR), the question is how to provide them.

This PR initially suggested to add a fragment to equinox that adds the required capabilities to one of the 'original' `javax.servlet` artifacts, but at the moment I'm more declined to use the `org.apache.felix.http.servlet-api:1.2` which provides the classes from these `javax.servlet` artifacts but additionally with the missing metadata.
So it would be https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239 plus this PR without the addition of the new fragment.

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,18 @@
...
+Bundle-ManifestVersion: 2
+Bundle-Name: JavaServlet OSGi Contract Provider for Equinox
+Bundle-SymbolicName: org.eclipse.osgi.service.http.servlet.contract;singleton:=true
+Automatic-Module-Name: org.eclipse.osgi.service.http.servlet.contract
+Bundle-Vendor: Eclipse.org - Equinox
+Bundle-Version: 1.0.0.qualifier
+Fragment-Host: 
+ jakarta.servlet-api;bundle-version="[4.0.0,5.0.0)",
+ javax.servlet-api;bundle-version="[3.0.0,5.0.0)",
+ org.eclipse.jetty.servlet-api;bundle-version="[4.0.0,5.0.0)"
```
tjwatson, line 11: A fragment is only allowed to specify a single host.

tjwatson:
How do we ensure what bundle is providing `javax.servlet`?  Is that a bit random at the moment?  I ask because I thought we used to use the "official" `javax.servlet:javax.servlet-api` one.  But now, in my current installation, I see `org.eclipse.jetty.servlet-api_4.0.6` is installed.  I assume this is because we let p2 just find any available provider in the available repositories for the `Import-Package` requirement for the `javax.servlet` packages.  For the moment that is pulling in the one from the jetty repository.

So if we move to use the `org.apache.felix.http.servlet-api` how do we ensure that is the ONLY one we include in the platform?  Will that further complicate others that try to include the jetty one?  It seems like it maybe more simple just to have our http.servlet implementation provide the contract or perhaps get the `org.eclipse.jetty.servlet-api` bundle to provide the contract.

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
HannesWell: Ok, somehow I assume this is possible. At least the runtime didn't complain, but looking in the spec it does not look like this is permitted.
But ok, then we would have to select one of these three, in case we continue with adding this fragment. In that case it is probably the most future-proof to choose `jakarta.servlet-api`.

laeubi: @tjwatson are you sure? I also though this but remember I have seen such "multi-host" already somewhere else (sadly I can't remember where/when) and was surprised there as well...

akurtakov:
I honestly don't know. What I have always cared for is have latest jetty and have help webapp working - whatever satisfies it is good enough.

merks:
As @akurtakov  suggests, I think we can only test that help works.  Generally I test that by launching runtime workspace and seeing in Help -> Contents works, checking which bundles are installed...

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
tjwatson: > @tjwatson are you sure? I also though this but remember I have seen such "multi-host" already somewhere else (sadly I can't remember where/when) and was surprised there as well...

Yes, I'm sure the spec `Fragment-Host` syntax only allows for a single host symbolic name to be defined:

https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.fragmenthost

```
Fragment-Host       ::= bundle-description
bundle-description  ::= symbolic-name 
                            ( ';' parameter )* // See [1.3.2](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.introduction.html#framework.general.syntax)
```

It is true a fragment may be attached to multiple host bundles, but that is only when there are multiple versions of a bundle are installed (with the same symbolic name):

https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module.fragmentbundles

> Fragments can be attached to multiple hosts with the same symbolic name but different versions. If multiple fragments with the same symbolic name match the same host, then the Framework must only select one fragment, this must be the fragment with the highest version.


> Ok, somehow I assume this is possible. At least the runtime didn't complain, but looking in the spec it does not look like this is permitted.

This is because Equinox just uses the first one listed:

https://github.com/eclipse-equinox/equinox/blob/0582356d5ab0bc359b91dc242a92ecfc767463ed/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/builders/OSGiManifestBuilderFactory.java#L501

laeubi: @tjwatson is `osgi.support.multipleHosts=true` gone already?
https://www.eclipse.org/forums/index.php?t=msg&th=276829&goto=1052923&#msg_1052923

tjwatson: Regardless of my answer (which I don't recall off the top of my head), that was only about resolve against multiple host versions (with same BSN).

laeubi: I see.. then I maybe has wrong memories here...

tjwatson:
> But just adding the contract is not sufficient since org.osgi.service.http imports the javax.servlet(.http) package in version 2 up to, but excluding version 3. org.osgi.service.http.whiteboard on the other hand imports both packages without a version range.

This implies we need a single bundle exporting two versions of the same package.  While possible, it is not truthful.  I am not sure I agree that we should move in that direction.  To that end, I really think we should repackage the deprecated `org.osgi.service.http` classes into our own bundle that imports the servlet packages at the right version we need.  That is the correct way to do this.

tjwatson:
> To that end, I really think we should repackage the deprecated `org.osgi.service.http` classes into our own bundle that imports the servlet packages at the right version we need. That is the correct way to do this.

Actually, we should import with no version and do the contract dependency like the OSGi provided `org.osgi.service.http.whiteboard` bundle does.

laeubi:
I also want to note that it seems the `org.osgi.service.http` is already pulled in I get a lot of warnings in the maven build like this:

> Problems resolving provisioning plan.: [Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet [2.1.0,3.0.0).; Unable to satisfy dependency from org.osgi.service.http 1.2.2.202109301733 to java.package; javax.servlet.http [2.1.0,3.0.0).; Unable to satisfy dependency from org.osgi.service.http.whiteboard 1.1.1.202109301733 to osgi.contract; (&(osgi.contract=JavaServlet)(version=3.1.0)).]

what is quite annoying it would be great to get a solution for this as this currently hinder us from using the Quality gate as this is reported as new warnings on maven problem parser.

HannesWell:
> > To that end, I really think we should repackage the deprecated `org.osgi.service.http` classes into our own bundle that imports the servlet packages at the right version we need. That is the correct way to do this.
> 
> Actually, we should import with no version and do the contract dependency like the OSGi provided `org.osgi.service.http.whiteboard` bundle does.

Should I just place the API sources into the Implementation bundle and export+import them or create a separate bundle?

akurtakov:
Is there any ETA on this one? It has been with us for way too long and actually prevents verifications from catching up new warnings.

HannesWell:
I'll try to complete this later tonight.
But I would like to await Tom's approval then, so it probably won't be resolved before tomorrow.

tjwatson:
> I'll try to complete this later tonight.
> But I would like to await Tom's approval then, so it probably won't be resolved before tomorrow.

I think the path forward is a new bundle with the only the old `org.osgi.service.http` package source in it which is then required by our shell of a `org.eclipse.osgi.services` bundle.  And also pull in `org.osgi.service.http.whiteboard` bundle for the whiteboard from maven central.

HannesWell:
> > I'll try to complete this later tonight.
> > But I would like to await Tom's approval then, so it probably won't be resolved before tomorrow.
> 
> I think the path forward is a new bundle with the only the old `org.osgi.service.http` package source in it which is then required by our shell of a `org.eclipse.osgi.services` bundle. And also pull in `org.osgi.service.http.whiteboard` bundle for the whiteboard from maven central.

So should the new o.o.service.http wrapper bound then also provide the osgi.contract capabiltity required by the http.whiteboard bundle?

My plan was actually to simply embed the jars of `org.osgi.service.http` and `org.osgi.service.http.whiteboard` into `org.eclipse.equinox.http.servlet`, adding a corresponding Bundle-ClassPath and export the `org.osgi.service.http*` packages (that are also imported by `o.e.equinox.http.servlet`).
When using the maven-dependencies-plugin the jars could just be copied during the build (on the CI and in the IDE) so there would be no need to check them in into git.

That should definitively work during the build and in the IDE (when m2e is installed but IIRC this is now mandated for sdk development?).
If we have a capabilities in the reactor to satisfy requirements in the TP I'm not sure if this will work (at least it didn't with the current state of this PR).

tjwatson:
> So should the new o.o.service.http wrapper bound then also provide the osgi.contract capabiltity required by the http.whiteboard bundle?

I think that is fine.

> My plan was actually to simply embed the jars of `org.osgi.service.http` and `org.osgi.service.http.whiteboard` into `org.eclipse.equinox.http.servlet`, adding a corresponding Bundle-ClassPath and export the `org.osgi.service.http*` packages (that are also imported by `o.e.equinox.http.servlet`).

That is not optimal for JAR'ed shaped bundles because then the framework has to extract out the embedded jar at runtime to load classes from it.  Probably not a big performance issue, but not something I want to promote either.  But keep in mind that we will need to do the same kind of trick we do for the `service.log` package in the `org.eclipse.osgi.services` bundle:

https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi.services/src/org/osgi/service/log/note.txt

That has an export with the directives `;mandatory:="never.ever.use";never.ever.use="true"` so that it is only there for requires of the `org.eclipse.osgi.services` bundle

> When using the maven-dependencies-plugin the jars could just be copied during the build (on the CI and in the IDE) so there would be no need to check them in into git.

That is good, but why cannot we use maven then to copy the class resources into the actual bundle root instead?

Overall it still seems less tricky to just have a new bundle `org.eclipse.equinox.http.service.apis` or something similar that has the `http.service` APIs.

HannesWell:
I just noticed that the build was failing because I didn't add the contract-fragment to the pom.xml...
Anyway, since I had the impression that it isn't the best idea we shouldn't continue with that.


> > When using the maven-dependencies-plugin the jars could just be copied during the build (on the CI and in the IDE) so there would be no need to check them in into git.
> 
> That is good, but why cannot we use maven then to copy the class resources into the actual bundle root instead?
> 
> Overall it still seems less tricky to just have a new bundle `org.eclipse.equinox.http.service.apis` or something similar that has the `http.service` APIs.

Yes we can do that. I'll just have to figure out how that behaves in the IDE and how  to make the workspace build happy we could just copy the sources when the project is checked out to let build JDT the usual way.

HannesWell:
> > When using the maven-dependencies-plugin the jars could just be copied during the build (on the CI and in the IDE) so there would be no need to check them in into git.
> 
> That is good, but why cannot we use maven then to copy the class resources into the actual bundle root instead?
> 
> Overall it still seems less tricky to just have a new bundle `org.eclipse.equinox.http.service.apis` or something similar that has the `http.service` APIs.

Have done that now.
To only not so nice part is that at the moment I didn't manage to convince tycho via the build.properties to add the content of the classes folder (in which the compiled classes from the original o.o.service.http jar are copied) at the root of the built bundle. So I added a `Bundle-Classpath: classes/` fro now.

Another open point is the question if we need a license file in the new plugin and if yes, which license since all the contained code is apache licensed.

> But keep in mind that we will need to do the same kind of trick we do for the `service.log` package in the `org.eclipse.osgi.services` bundle:
> 
> https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi.services/src/org/osgi/service/log/note.txt
> 
> That has an export with the directives `;mandatory:="never.ever.use";never.ever.use="true"` so that it is only there for requires of the `org.eclipse.osgi.services` bundle

Why is that in this case? For `org.osg.service.log` this was only done because that package is also contained in `org.eclipse.osgi` but that (framework-) bundle is not re-exported by `o.e.services`. That's not the case for the `o.o.service.http*` packages.

laeubi:
I just wanted to note that the "never ever use" construct actually has had made problem in the past because bnd is just using that and happily adds the directive to manifest and I wonder why not a simple import/export can be used here.

Beside that I'm not quite sure adding a **new** bundle to support the legacy `org.osgi.service.http` is a good idea also making some "tricks" to embeds the original sources/classes sounds a bit odd given these will likely never change and the main idea of **reusing** things is/was that we do not wrap things again and again and end up with the "same thing" multiple times.

If that the way to go, I see two options that might fit better here:

1. We can leverage "orbit 2.0" to rebundlle anything as @merks is already doing for some of the stuff there then at least we don't need to host things at equinox and do special tricks like maven copy and alike.
2. We can move the sources of `org.osgi.service.http` to the bundle that actually implements the http-service and have it import/export the packages there, this is actually one of the rare cases (impl want to carry its API) where import/export with substitution packages is useful. It then can also export any missing contracts for the `org.osgi.service.http.whiteboard`
 
For the `o.e.services` this might not be optimal **but** I strongly believe that no one should have ever had "require-bundle" that bundle anyways, so if that's still done and only for legacy reasons then it would be fine for me to pull in the http-impl bundle for those as they likely will need it anyways as they run inside eclipse *or* that might finally convince them to finally migrate.

tjwatson:
> Why is that in this case? For `org.osg.service.log` this was only done because that package is also contained in `org.eclipse.osgi` but that (framework-) bundle is not re-exported by `o.e.services`. That's not the case for the `o.o.service.http*` packages.

Because you were suggesting to have `o.e.osgi.services` have a hard require-bundle (re-export) on the Equinox http service implemntation.  We would want to avoid that, but we need to do the trick of making an export that is not possible to be imported easily so that our own import is forced to get it from something else.  This ensures that the bundle contract of `o.e.osgi.services` remains the same.

tjwatson:
> I just wanted to note that the "never ever use" construct actually has had made problem in the past because bnd is just using that and happily adds the directive to manifest and I wonder why not a simple import/export can be used here.

Seems like a bnd problem, not ours.  I explained why it is needed above.

> Beside that I'm not quite sure adding a **new** bundle to support the legacy `org.osgi.service.http` is a good idea also making some "tricks" to embeds the original sources/classes sounds a bit odd given these will likely never change and the main idea of **reusing** things is/was that we do not wrap things again and again and end up with the "same thing" multiple times.

I would rather not do anything to embed the original classes into our bundles, just seems like a complication we can avoid altogether.  I think we should put the java source into a bundle and just do what we need to package them ourselves.  I think it would be best to have a separate bundle for that vs. putting that directly into the implementation.

HannesWell:
> > Why is that in this case? For `org.osg.service.log` this was only done because that package is also contained in `org.eclipse.osgi` but that (framework-) bundle is not re-exported by `o.e.services`. That's not the case for the `o.o.service.http*` packages.
> 
> Because you were suggesting to have `o.e.osgi.services` have a hard require-bundle (re-export) on the Equinox http service implemntation. We would want to avoid that, but we need to do the trick of making an export that is not possible to be imported easily so that our own import is forced to get it from something else. This ensures that the bundle contract of `o.e.osgi.services` remains the same.

OK, yes makes sense. I was just confused because for `o.o.service.log` we did it for another reason (because the package is also contained in o.e.osgi). But if we have a an API-only bundle that shouldn't be necessary, should it?

> > Beside that I'm not quite sure adding a **new** bundle to support the legacy `org.osgi.service.http` is a good idea also making some "tricks" to embeds the original sources/classes sounds a bit odd given these will likely never change and the main idea of **reusing** things is/was that we do not wrap things again and again and end up with the "same thing" multiple times.
> 
> I would rather not do anything to embed the original classes into our bundles, just seems like a complication we can avoid altogether. I think we should put the java source into a bundle and just do what we need to package them ourselves. I think it would be best to have a separate bundle for that vs. putting that directly into the implementation.

That's fine. I did it because (I had the impression) you asked to use the original class files, but yes just recompiling them makes it simpler. I have now changed the pom.xml to copy the sources of `o.o.service.http` to the `src` folder. From a build-setup perspective it would be even simple if the sources are just checked in into the SCM. We could then remove the pom.xml but would have the sources effectively just moved form o.e.osgi.services to the new bundle.

In general there are multiple different possible alternatives how to solve this, but since I would like to complete this task at some time I think it is fine to go this way.

I also added a about.html based on the one from `o.e.osgi.services`, but I think it has to be overhauled to reflect that the OSGi Alliance is now the OSGi Working-Group at the EF. If anybody knows the exact formalities (e.g. does the OSGi WG have an address?) or has a proposal for the about.html I would be happy about that, otherwise I will look it up.

Besides the `about.html` this is ready for me.

HannesWell:
> If that the way to go, I see two options that might fit better here:
> 
>     1. We can leverage "orbit 2.0" to rebundlle anything as @merks is already doing for some of the stuff there then at least we don't need to host things at equinox and do special tricks like maven copy and alike.

In general this sounds like a not to bad alternative as well, but I would just re-use the BSN `org.eclipse.equinox.http.service.api` to reflect which project is 'responsible' for that re-packaging.
But I wonder how this would be done in detail? Looking at https://github.com/eclipse-orbit I see only Orbit-legacy as a place for that and it would be basically the same as if we add the api-bundle here. And if that's the case, I would keep it in equinox.

merks:
What is the goal?  To rebundle what is already an OSGi bundle? 

That's a bit of an involved process because m2e does not support forced replacing an existing MANIFEST.MF.  

So when that is needed, as in the cases of the bogus axis OSGi manifest, we must strip the MANIFEST.MF, publish the stripped to jar repo.eclipse.org, and then consume it from there as we normally consume things directly from Maven central:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/maven-deploy/MavenAxis.jenkinsfile

https://github.com/eclipse-orbit/orbit-simrel/blob/a059cb654a8310e311d76dbe3bc346c131d3606b/maven-bnd/tp/MavenBND.target#L641-L656

laeubi:
> I would rather not do anything to embed the original classes into our bundles, just seems like a complication we can avoid altogether. 

I don't see that this complicates anything, but inventing a new bundle will (again) bind us to something we in some point of time will be hard to get rid of...

> I think we should put the java source into a bundle and just do **what we need** to package them ourselves. 

Especially as I have highlighted the important part here, we are only doing it for our needs (== the http service impl) it won't be anything useful for others and actually we would need the import/export then at this place to make it possible to substitute the package e.g. if one want to use it with the felix impl in one framework.

@merks the problem is (if I have not completely lost track here) that the officals require some kind of newer API/contracts than we currently want to support, but still not sure why 

- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/239

is not an option, adding a new API(!) bundle that replicates already existing repackaged variants seems completely wrong to me, especially if we want to support some kind of legacy service.

laeubi (changes requested):
I think this has become to cumbersome to handle in one go with to mayn confusing implications.

I therefore created
- https://github.com/eclipse-equinox/equinox/issues/434

as a place for discussion to at a first step **only** replace the `org.osgi.service.http` package here and then after that investigate / do the same for the `org.osgi.service.http.whiteboard`.

tjwatson:
There is much discussion about how to use various things to reconsume the "official" `org.osgi.service.http` bundle which is deprecated by OSGi.  I still firmly believe the most simple thing is to build that java source in our own bundle project and publish our own bundle for this one deprecated package.  I see no advantage to using Orbit to do that.  Either way I believe the final bundle BSN has to be different from the OSGi WG provided one.  Otherwise all sorts of confusion will arise.

> I don't see that this complicates anything, but inventing a new bundle will (again) bind us to something we in some point of time will be hard to get rid of...

No matter what we cannot use the `org.osgi.service.http` bundle from the OSGi WG in its current form because it binds us to a version of javax.servlet which is too old.  There is little chance OSGi WG will publish an updated version of this that will make it work.  No matter what we do it will be something we are bound to.  Weather that is repackaging in Orbit or rebuilding from source ourselves into our own bundle.

HannesWell:
> I still firmly believe the most simple thing is to build that java source in our own bundle project and publish our own bundle for this one deprecated package.  I see no advantage to using Orbit to do that.  Either way I believe the final bundle BSN has to be different from the OSGi WG provided one.  Otherwise all sorts of confusion will arise.

I also tend to prefer that solution since there is no ideal way at the moment.
And I would like to avoid having a third provider of the servlet API which would be the consequence of option 1 and 3 of #434.

No matter how we do it, a new artifact will be necessary, either provided by Equinox or Orbit (or Felix servlet API), unless the 'offical' OSGi bundles are updated.

> 
> > I don't see that this complicates anything, but inventing a new bundle will (again) bind us to something we in some point of time will be hard to get rid of...
> 
> No matter what we cannot use the `org.osgi.service.http` bundle from the OSGi WG in its current form because it binds us to a version of javax.servlet which is too old.  There is little chance OSGi WG will publish an updated version of this that will make it work.  No matter what we do it will be something we are bound to.  Weather that is repackaging in Orbit or rebuilding from source ourselves into our own bundle.

Can you estimate the effort and possible timeframe to publish a micro version update with only a wider import packe version range?

But even if o.o.service.http is updated there is still the problem of the missing JavaServlet contact required by the http.whiteboard. Would it make sense to remove that contract and just relay on imported packages?

tjwatson:
> Can you estimate the effort and possible timeframe to publish a micro version update with only a wider import packe version range?

The effort is probably small.  Although the OSGi WG build and publishing machinery isn't equipped to doing a release of a single component.  So it would have to be some kind of one-off process.

But it isn't about the amount of effort.  Instead it is about providing a release to something that is deprecated and removed from the latest release of the specification:  https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.servlet.html

And the source isn't even in the `main` branch of OSGi anymore.
It is in the past `r8-cmpn-final`:  https://github.com/osgi/osgi/tree/r8-cmpn-final/org.osgi.service.http
But it is not in the `main` branch: https://github.com/osgi/osgi/tree/main/org.osgi.service.http

HannesWell:
Thanks Tom for the elaboration. It sounds doable but is probably the solution with the biggest effort.

Now that we have discussed (I think exhaustively) various approaches to solve this problem, I would like to come to a decision in order to finally complete this.

Personally I would be fine with repacking `o.o.service.http` in equinox either in a new api bundle or with the API source embedded into the equinox.servlet implementation.
I would also be fine to create a new fragment to jakarta.servlet (version 4 which provides `javax.servlet`) to add exports and the contract. With all the concerns to repack things that's currently my favourite because no bundle is changed or repacked and a fragment cannot be required (only included in Features) so I would hope that it can be removed when not needed anymore.
I'm not in favor of repacking an old javax.servlet API bundles in orbit or o.o.service.http or adding felix servlet API (a third provider of the servlet API is IMHO too much).

But in the end I have no strong opinion about any solution and would mainly like to reach a conclusion.

@tjwatson and @laeubi could you please list the possible solutions which would be acceptable for you and your preferred one.
I think Tom has already stated that his preference is to repack o.o.service.http in a dedicated API bundle.

Maybe the intersection of acceptable solutions is not empty, but if it is I think Tom has to decide as project lead.

laeubi:
Actually I opened the new issue so we can have a more focused discussion there instead of adding to this one where it already becomes confusing, and where I also summarized some of the the (alternative) options.

> Either way I believe the final bundle BSN has to be different from the OSGi WG provided one. Otherwise all sorts of confusion will arise.

It still will contain the same packages but with a slightly different use constraints, therefore my suggestion to do it like its done for felix that the impl simply carry the api packages and reexport them instead of inventing "just another" provider that only differs in name.

> form because it binds us to a version of javax.servlet which is too old

Can you explain what means "too" old so where do we need Servlet 3+ features and why it can't be solved by a fragment as suggested here

- https://github.com/eclipse-equinox/equinox/issues/434

> But it isn't about the amount of effort. Instead it is about providing a release to something that is deprecated and removed from the latest release of the specification

Actually one don't need a build, sources or any special infrastructure. You can simply download the jar+pom, the sources and javadoc, then modify the manifest to whatever is desired and upload the binaries again like described here:

https://maven.apache.org/guides/mini/guide-3rd-party-jars-remote.html

only think to consider is that one needs to generate a new signature for the modified jar.

> I would also be fine to create a new fragment to jakarta.servlet

As described [here](https://github.com/eclipse-equinox/equinox/issues/434) I would better add a fragment to the `org.osgi.http` service itself to modify its headers to what we desire.

> a third provider of the servlet API is IMHO too much

I can't see how it matter how much servlet API provider we have as these have no dependencies and play quite well in parallel.

> could you please list the possible solutions which would be acceptable for you and your preferred one.

I already did it here:
- https://github.com/eclipse-equinox/equinox/issues/434

tjwatson:
> Can you explain what means "too" old so where do we need Servlet 3+ features and why it can't be solved by a fragment as suggested here

Servlet < 4 is too old and that is what the "official" old deprecated OSGi WG published bundle requires with its import range..  The supported Jetty version we use is Servlet 4.  We must use servlet 4 with the supported jetty version.

We cannot  "solve with a fragment" here.  A fragment can only add new requirements to a host, it cannot replace requirements in that host.

> @tjwatson and @laeubi could you please list the possible solutions which would be acceptable for you and your preferred one.
I think Tom has already stated that his preference is to repack o.o.service.http in a dedicated API bundle.

For the record, my vote is to import the java source into our own bundle project to provide the `org.osgi.service.http` package and have the `org.eclipse.osgi.services` bundle require/re-export that bundle.  I see no advantage to placing any build dependency on a deprecated bundle from the OSGi WG which is guaranteed to never get a version update.  We will have no way to service anything out of the package classes if we do that.

laeubi:
> We cannot "solve with a fragment" here. A fragment can only add new requirements to a host, it cannot replace requirements in that host.

I'm not completely sure, [the spec says](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.fragmentbundles):

> Append the import definitions for the Fragment bundle that do not conflict with an import definition of the host to the import definitions of the host bundle. 

a conflict is described as

> A host and a fragment conflict when they cannot resolve to provide a consistent class space. 

So enhancing the ragen should usually not create a conflict unless someone strictly requires the lower version and I think I have seen such usage somewhere already, but have not tested it specifically for this case.

tjwatson:
> So enhancing the ragen should usually not create a conflict unless someone strictly requires the lower version and I think I have seen such usage somewhere already, but have not tested it specifically for this case.

From `org.eclipse.jetty.ee8.servlet`
```
Import-Package: javax.servlet;version="[4.0.0,5)", javax.servlet.http;version="[4.0.0,5)"
```

From the deprecated OSGi WG bundle `org.osgi.service.http`
```
Import-Package: javax.servlet;version="[2.1,3)",javax.servlet.http;version="[2.1,3)"
```

There is no intersection between `[2.1,3)` and `[4.0.0,5)`.  The import from the old deprecated  `org.osgi.service.http` bundle even tries to use semantic versioning for the package import of javax.servlet.  The OSGi WG gave up trying to force Java/Jakarta EE to follow that.  For example, we maintained that the servlet 3.0 and 3.1 specifications really were compatible with Serlvet 2 so should have been some package version 2.x.  But the industry didn't agree to do semantic package versions for the javax.servlet packages.  So that range `[2.1,3)` is even more bogus.

Adding a fragment to the old deprecated `org.osgi.service.http` cannot help.  Say we add

```
 Import-Package: javax.servlet;version="[2.1,5)",javax.servlet.http;version="[2.1,5)" with a fragment.
```

That simply adds a new requirement to the host, but the original requirement with range `[2.1,3)` must still be satisfied only now it must be satisfied to a capability that not only satisfies the original range but also the new range of `[2.1,5)` such that both are satisfied.  That is only possible with a package version <3.

We need to stop trying to think of ways to use an old deprecated bundle binary from another project.  We still need the API, we should maintain the API ourselves since the OSGi WG has given up on it.  Once we stop using the API altogether then we can remove it.

tjwatson:
> I have checked that actually no one in platform is using the API except equinox.http/extensions so what about then put it into `org.eclipse.equinox.http.servlet` and export it with a new minor version e.g. 1.3.0 for the packages? The it would be clear that it is "our" API, for consumers it should not make a difference and the impl can have 1.3 as lower bound to always bind to our API?

I'm losing track of what problem that solves.  `1.3.0` doesn't make sense to me because we did not add any new API to the package.  My vote remains the same.

laeubi:
> I'm losing track of what problem that solves. 1.3.0 doesn't make sense to me because we did not add any new API

The [spec says](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.html)

> Http Service implementations must support **at least version 2.1** of the Java Servlet API. 

if we now have a package that do not support Java Servlet API 2.1 anymore but requires 4+ we effectively have a new API / service contract as you have explained above.

tjwatson:
> if we now have a package that do not support Java Servlet API 2.1 anymore but requires 4+ we effectively have a new API / service contract as you have explained above there is no intersect between "old" contract and "actual".

I never implied that.  Updating a dependency of the bundle that happens to export some API package does not change the contract of that API package.

HannesWell:
> Adding a fragment to the old deprecated `org.osgi.service.http` cannot help. Say we add
> 
> ```
>  Import-Package: javax.servlet;version="[2.1,5)",javax.servlet.http;version="[2.1,5)" with a fragment.
> ```
> 
> That simply adds a new requirement to the host, but the original requirement with range `[2.1,3)` must still be satisfied only now it must be satisfied to a capability that not only satisfies the original range but also the new range of `[2.1,5)` such that both are satisfied. That is only possible with a package version <3.

At least my initial suggestion in this PR was not to add a fragment to `org.osgi.service.http` but to `jakarta.servlet-api` (version 4 which provides the `javax.servlet` packages) in order to extend the provided capabilities of the latter bundle in order to satisfy the difficult requirements of `o.o.service.http` and `o.o.service.http.whiteboard`:
```
Bundle-SymbolicName: org.eclipse.osgi.service.http.servlet.contract;singleton:=true
Fragment-Host: jakarta.servlet-api;bundle-version="[4.0.0,5.0.0)"
Provide-Capability: osgi.contract;osgi.contract=JavaServlet
 ;version:List<Version>="2.6,3.0,3.1,4.0"
 ;uses:="javax.servlet,javax.servlet.http,javax.servlet.descriptor,javax.servlet.annotation"
Export-Package: 
 javax.servlet;version="2.999",
 javax.servlet.http;version="2.999";uses:="javax.servlet"
```
This would have the advantage that we can use only 'official' jars for real content plus that it is only pulled in if something uses the `o.o.service.http` jar. So if one day all usages in the platform are gone, that fragment can be removed immediately since one cannot reference it (at least in bundles). To be sure I just immediately mark the fragment as deprecated to tell users that one should not rely on its export of javax.servlet version 2.

I know it would be hacky, but IMHO to most elegant way to solve this misery.
@tjwatson, @laeubi would you be fine with that?

I also looked into embedding the `o.o.service.http` api sources into `org.eclipse.equinox.http.servlet` and exporting it from that, but that would have the disadvantage that `org.eclipse.equinox.http.registry` would effectively require `equinox.http.registry` and at least from the features they are included in I have the impression that until now they are not necessarily used together.
So if we decide to embedd the sources in equinox I agree with Tom that it should be a dedicated api-bundle.
But if we plan to remove all dependencies on `o.o.service.http` in the future we can also immediately mark the api-bundle as deprecated to reflect that the spec is 'deprecated' too.

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
HannesWell: Changed the list to have the official version. Since o.o.service.http.whiteboard only uses 3.1 all the others doesn't matter for our use-case anyway.

bundles/org.eclipse.osgi.service.http.servlet.contract/META-INF/MANIFEST.MF
HannesWell: Applied this now.
Additionally used packages should not harm consumers of the servlet version 2 package since they can't use it.

HannesWell (commented):
Fixed some of the old remarks and also marked the fragment already as deprecated so to make clear that nobody should relay on the servlet packages version 2 being exported or the `JavaServlet` contract provided by it.
If one day we got rid of o.o.service.http we should be able to remove it immediately (the contract can then be provided by some other plugin, e.g. the one that implements the http.whiteboard spec).

merks:
Quite a few things use the to-be-deprecated bundle:

image

The Orbit Jetty build would fail like this without that bundle:
```
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.orbit.maven.jetty.site raw:1.0.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.0.0-SNAPSHOT
[ERROR]   Missing requirement: org.eclipse.jetty.ee10.osgi.alpn.fragment 12.0.4 requires 'osgi.bundle; org.eclipse.osgi 0.0.0' but it could not be found
[ERROR]   Cannot satisfy dependency: org.eclipse.orbit.maven.jetty.all.feature.group 12.0.4 depends on: org.eclipse.equinox.p2.iu; org.eclipse.jetty.ee10.osgi.alpn.fragment [12.0.4,12.0.4]
[ERROR]   Cannot satisfy dependency: org.eclipse.orbit.maven.jetty.site raw:1.0.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.0.0-SNAPSHOT depends on: org.eclipse.equinox.p2.iu; org.eclipse.orbit.maven.jetty.all.feature.group 0.0.0
```
The Orbit OSGi would fail like this:
```
[ERROR]   Software being installed: org.eclipse.orbit.maven.osgi.site raw:1.0.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.0.0-SNAPSHOT
[ERROR]   Missing requirement: org.eclipse.core.runtime 3.30.0.v20231102-0719 requires 'osgi.bundle; org.eclipse.osgi [3.17.0,4.0.0)' but it could not be found
[ERROR]   Missing requirement: org.eclipse.equinox.common 3.18.200.v20231106-1826 requires 'osgi.bundle; org.eclipse.osgi [3.17.200,4.0.0)' but it could not be found
[ERROR]   Cannot satisfy dependency: org.bndtools.templating 7.0.0.202310060912 depends on: java.package; org.eclipse.core.runtime 0.0.0
[ERROR]   Cannot satisfy dependency: org.eclipse.orbit.maven.osgi.all.feature.group 4.31.0.v20231214-0900 depends on: org.eclipse.equinox.p2.iu; org.bndtools.templating [7.0.0.202310060912,7.0.0.202310060912]
[ERROR]   Cannot satisfy dependency: org.eclipse.orbit.maven.osgi.site raw:1.0.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.0.0-SNAPSHOT depends on: org.eclipse.equinox.p2.iu; org.eclipse.orbit.maven.osgi.all.feature.group 0.0.0
```
I guess it will be quite some significant period of time before there are no requirements on the bundle anymore....

HannesWell:
Thanks Ed for the analysis

The requirements on the exported packages will automatically vanish respectively will be redirected, once this PR is completed and the http packages are not contained in this Plugin/Bundle anymore.
The requirements on this the `o.e.osgi.service` bundle should be replaced by Imports of the required packages. Then the requirements are also automatically 're-directed' to the official osgi jars, too.

Correspondingly the inclusion in features should just be removed. I would only keep it in at least one equinox feature so that it is still included in the sdk repo and in the TP of consumers that still use it.

I'll reach out to the affected projects ones this is completed.

> I guess it will be quite some significant period of time before there are no requirements on the bundle anymore....

I assumed that we have the usual two years deprecation period?

HannesWell:
> > We need to stop trying to think of ways to use an old deprecated bundle binary from another project. We still need the API, we should maintain the API ourselves since the OSGi WG has given up on it. Once we stop using the API altogether then we can remove it.
> 
> I've not seen any arguments against this approach that involves no hacking, n repackaging, and no confusing multi-version exports.

OK, I just wanted to make sure that this suggestion is not lost because I had the impression that it was confused with a fragment for `o.o.service.http` instead of `jakarta/javax.servlet`.
Then lets move the `org.osgi.service.http` classes into a new `org.eclipse.equinox.http.service.api` bundle. When all elements that reference something from `org.osgi.service.http` we can just deprecate and eventually remove the bundle with them.

I have update this PR to move the classes into a new `org.eclipse.equinox.http.service.api` bundle. At the moment the project is configured to copy the sources in the bundle using the maven-depencency-plugin on the fly during CI and workspace builds.
If you prefer to have the sources checked into git, I can change that as well.
Are you then fine with this PR as it is?

tjwatson:
> Are you then fine with this PR as it is?

How does that work with the projects imported into the IDE?  Does PDE and M2E handle this and get everything compiled correctly in the workspace?

tjwatson:
> Yes. Of course m2e has to be installed. But IIRC it is already part of the Oomph development setup for the SDK.

Then I am fine with this approach, but if there are any issues we should simply place the `.java` source into the project and be done with it.

merks:
I'm such a non-fan to not checking in code, but that's just my opinion.  Please be 100% sure that the Oomph setup actually works out-of-the box from a freshly cloned clone.

tjwatson:
> I'm such a non-fan to not checking in code, but that's just my opinion.

Not just yours, that is mine also.  It always seems more simple to just have the source checked into the source repo vs. trying to pull it out of some pre-existing artifact.

tjwatson:
> trying to pull it out of some pre-existing artifact.

One advantage I can think of is that it would stop others from opening PRs trying to "correct" the formatting of the source one way or another :)

HannesWell:
> > Yes. Of course m2e has to be installed. But IIRC it is already part of the Oomph development setup for the SDK.
> 
> Then I am fine with this approach,

Great. I'm glad we can finally complete this. Thank you all for the discussion and your patience.


> I'm such a non-fan to not checking in code, but that's just my opinion. Please be 100% sure that the Oomph setup actually works out-of-the box from a freshly cloned clone.

I verified that and for me it worked well to import this and fetch the sources just with the Oomph setup. I'm never 100% sure about anything but I'm very confident it works if m2e is installed. :)
But because I have the impression that, contrary to my previous assumption m2e is not installed with the [SDK setup](https://raw.githubusercontent.com/eclipse-platform/eclipse.platform.releng.aggregator/master/oomph/PlatformSDKConfiguration.setup) and have now kept the sources checked-in in the SCM.



> > trying to pull it out of some pre-existing artifact.
> 
> One advantage I can think of is that it would stop others from opening PRs trying to "correct" the formatting of the source one way or another :)

That's right. And it is also simpler to upgrade to a new version because only the version number in the pom (and the manifest in Manifest-First plugins) needs to be updated. For this abandoned spec that's not so interesting, but for example for the embedded sources in `org.eclipse.osgi` that might be of interest.
If you are interested and we agree to make m2e mandatory for SDK development I can provide a PR to implement that.

tjwatson:
> If you are interested and we agree to make m2e mandatory for SDK development I can provide a PR to implement that.

I would rather not.  There are cases where we modify the internals of the `org.osgi` source in that project.  I believe it is more simple to keep the source as is.  Besides that, being used as the implementation during the development of the specification there are times where we need the source in our repo before OSGi has even published a source artifact at all.

PR #404, opened 2023-11-20
Fix unnecessary ecj problem markers

jukzi:
`At least one of the problems in category ... is not analysed due to a compiler option being ignored`

https://github.com/eclipse-platform/.github/discussions/169


jukzi:
there are only unrelated test errors

HannesWell (approved):
Thank you.

PR #406, opened 2023-11-25
Move equinox.common/registry/app/preferences and dependents to Java-17

HannesWell:
The migrated plugins are all projects in equinox that are currently on Java-11.

@tjwatson I hope none of these plugins is one of the ones you want to keep at Java-1.8.
Could you please list the projects that should stay at JavaSE-1.8?
I only recall org.eclipse.osgi for sure, but I think not all of those that are currently at 1.8 need to stay there and could be migrated to Java-17 as well?


HannesWell:
> These are the ones I need to stay at Java 8:

Thanks for the list. I left those bundles untouched plus the following ones
- org.eclipse.equinox.supplement
Considered this effectivly as part of org.eclipse.osgi
- org.eclipse.osgi.compatibility.state
Not developed anymore
- org.eclipse.osgi.util and org.eclipse.osgi.services
Only provides metadata already respectively will (hopefully) only provide metadata soon
- org.eclipse.equinox.launcher
Assuming that it is used to launch a most basic osgi application I assumed it should stay at 1.8 too

All other bundles, including all tests now require Java-17. For the tests I think a lower java requirement is not necessary since the tests run in the CI with Java-17 anyways.

akurtakov:
manual rebase is needed

PR #407, opened 2023-11-25
Delete unused/unpublished o.e.equinox.slf4j.stub

HannesWell:
The Plugin `org.eclipse.equinox.slf4j.stub` is not published anymore (at least not in https://download.eclipse.org/eclipse/updates/4.29/R-4.29-202309031000/plugins/) and in general I don't know any use of it these day.
Setting up logging with slf4j nowadays should work completely without it.

Or does anybody know a reason to keep this?


laeubi:
I would love to see https://github.com/osgi/slf4j-osgi used somehow to connect to the Equinox/Eclipse logging, now that we can consume maven bundles it should be perfectly fine...

HannesWell:
> I would love to see https://github.com/osgi/slf4j-osgi used somehow to connect to the Equinox/Eclipse logging, now that we can consume maven bundles it should be perfectly fine...

Have you tried using it with a recent eclipse? After a quick look I would assume that it should work out of the box as slf4j-backend so that all slf4j log-events end up as osgi(-log) events.
Probably that bundle has to be activated explicitly in the product configuration.

tjwatson:
IIRC this was necessary for:

org.eclipse.equinox.console.jaas.fragment
org.eclipse.equinox.console.ssh
org.eclipse.equinox.console.ssh.tests

But if we now have the APIs there all the time then this stub can likely be deleted.

akurtakov:
Ignoring always failing checks.

PR #408, opened 2023-11-26
Add update release workflow

laeubi:


laeubi:
Equinox requires the native launchers to be present for this to work and therefore the version bump does not work, but maybe they PR is created anyways with how far the workflow got it to work... lets see.

laeubi:
Workflow has failed, one enhancement might be to still create the PR:

https://github.com/eclipse-equinox/equinox/actions/runs/6993787119

PR #413, opened 2023-11-29
Update tycho-build to 4.0.4

akurtakov:


akurtakov:
@laeubi what's the status with the failing TCKs? Is anyone working on fixing them?

laeubi:
Not sure if anyone in Eclipse WG / OSGi WG has demands of Equinox passing all official TCKs.... at least we must take care that we do not break **more** TCKs...

laeubi:
e.g. [Device Access Specification](https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.device.html) has failing test, no idea if anyone is actually using this, maybe we can drop it from equinox completely @tjwatson ?

laeubi:
You can see the results from master here:

https://github.com/eclipse-equinox/equinox#implemented-services-an-compliance

the you can compare with for example this:
https://github.com/eclipse-equinox/equinox/actions/runs/7030130135/job/19129880065

akurtakov:
I have doubts that anyone will do the extra effort (me included).

laeubi:
> I have doubts that anyone will do the extra effort (me included).

As said it all depends on if we need Equinox to be something that implements something like OSGi or we need it to be an  OSGi Specification compliant implementation ...

akurtakov:
My question here is from releng POV - as long as people get used to see the red cross they stop caring about verification builds at all.

PR #416, opened 2023-12-03
Add data URL handler to equinox common

laeubi:
Currently I'm migrating code from bndtools -> pde and bndtools make use of the [data uri sceme](https://en.wikipedia.org/wiki/Data_URI_scheme) in several places.

As this protocol is universal, has no dependencies and is unlikely to evolve in the future  it seems suitable to be made available to a broader audience and therefore is more suitable to be hosted at equinox than with PDE project.

CQ is here
- https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/11781


laeubi:
CQ was approved now!

PR #419, opened 2023-12-05
fix 'access is emulated by a synthetic accessor'

jukzi:


jukzi:
there are only unrelated TCK fails

tjwatson (approved):

PR #420, opened 2023-12-05
EclipsePreferences: intern() keys and values

jukzi:
removes all duplicate Strings from
JavaModelManager.PerProjectInfo.options

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1692

### DIFF
--- a/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/EclipsePreferences.java
+++ b/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/EclipsePreferences.java
@@ -625,6 +625,11 @@ public class EclipsePreferences implements IEclipsePreferences, IScope {
 	 * null if no value existed.
 	 */
 	protected String internalPut(String key, String newValue) {
+		// intern the Strings to save memory
+		// shareStrings() comes too late,
+		// when plugins may have already copied preferences
+		newValue = newValue.intern();
+		key = key.intern();
 		synchronized (childAndPropertyLock) {
 			// illegal state if this node has been removed
 			checkRemoved();


HannesWell:
This is similar to https://github.com/eclipse-equinox/equinox/pull/33, isn't it?

vogella:
> This is similar to #33, isn't it?

@jukzi did you check if the performance of intern is better in modern JVMs?

merks:
It’s the circle of life that every complex topic comes around again as if it were touched for the very first time. 😱

jukzi:
> This is similar to #33, isn't it?

related, yes. i could not verify #33 solves jdt's problem, because that PR has conflicts with master

jukzi:
> @jukzi did you check if the performance of intern is better in modern JVMs?

no. i guess it's not, but it doesn't matter for this "few" preferences. This API is not in a hotspot. And the following "synchronize" will be in the same order of magnitude as intern()

HannesWell:
> It’s the circle of life that every complex topic comes around again as if it were touched for the very first time. 😱

Luckily we have a discussion we can point too 😅 


> > This is similar to #33, isn't it?
> 
> related, yes. i could not verify #33 solves jdt's problem, because that PR has conflicts with master

Resolved the conflicts and re-based it, but be aware of https://github.com/eclipse-equinox/equinox/pull/33#issuecomment-1509723507. So at the moment it is not exactly the same, but used to be it and could be changed to be it.

vogella:
https://github.com/eclipse-equinox/equinox/pull/33 also does some clean-up in the code base but seems to still use its own string pool. So my impression is that the combination of this and 33 would be the best one.

vogella:
> Resolved the conflicts and re-based it, but be aware of [#33 (comment)](https://github.com/eclipse-equinox/equinox/pull/33#issuecomment-1509723507). So at the moment it is not exactly the same, but used to be it and could be changed to be it.

Please do so, if we discover later issues with the internal method of the JVM we still have add the custom pool.

PR #422, opened 2023-12-08
Path: use java.util.Arrays convenience

jukzi:
instead of System.arraycopy

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
@@ -359,8 +359,7 @@ public final class Path implements IPath, Cloneable {
 		if (isRoot() || isEmpty() || hasTrailingSeparator())
 			return this;
 		int len = segments.length;
-		String[] newSegments = new String[len];
-		System.arraycopy(segments, 0, newSegments, 0, len - 1);
+		String[] newSegments = Arrays.copyOf(segments, len);
 		newSegments[len - 1] = segments[len - 1] + '.' + extension;
 		return new Path(device, newSegments, flags);
 	}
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
```diff
@@ -510,8 +507,7 @@ private static String[] collapseParentReferences(boolean isAbsolute, String[] se
 		if (stackPointer == segmentCount)
 			return segments;
 		// build the new segment array backwards by popping the stack
-		String[] newSegments = new String[stackPointer];
-		System.arraycopy(stack, 0, newSegments, 0, stackPointer);
+		String[] newSegments = Arrays.copyOf(stack, stackPointer);
 		return newSegments;
```
HannesWell, line 511:
```suggestion
		return Arrays.copyOf(stack, stackPointer);
```

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
```diff
@@ -1226,8 +1220,7 @@ public int segmentCount() {
 	 */
 	@Override
 	public String[] segments() {
-		String[] segmentCopy = new String[segments.length];
-		System.arraycopy(segments, 0, segmentCopy, 0, segments.length);
+		String[] segmentCopy = Arrays.copyOf(segments, segments.length);
 		return segmentCopy;
```
HannesWell, line 1224:
```suggestion
		return Arrays.copyOf(segments, segments.length);
```

HannesWell (commented):
Looks good, thank you.
Only two calls can be further inlined.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
jukzi: done

HannesWell (approved):
Looks good.
Thank you.

PR #423, opened 2023-12-09
Miscellaneous minor clean-ups in preference(-scope) implementations

HannesWell:
- Make AbstractScope.hashCode() more efficient by also considering the location, which is also checked in equals().
- replace Collections.synchronizedMap(new HashMap<>()) by ConcurrentHashMap
- Use Status factories
- Leverage default implementation of ISafeRunnable.handleException()


HannesWell:
Is anybody interested in reviewing this?
The changes are mostly local and ignoring white-space changes makes it much simpler.

laeubi:
I think if build is fine and no new API is introduced you don't need a second review and we trust in you here :-)

HannesWell:
That's fine. :)
Just didn't want to wait for nothing to happen.

The build passed so lets merge this.

PR #425, opened 2023-12-09
Replace deprecated PluginVersionIdentifier by org.osgi.framework.Version

HannesWell:
With this and https://github.com/eclipse-platform/eclipse.platform.ui/pull/1392 `PluginVersionIdentifier` is only referenced in its own tests.
Since it is deprecated since 2015 (8 years from now) do you think it is time to deprecate it for removal and kick of the removal process?

Does anybody want to review this? If yes, it is easier to do it, when ignoring white-space.


laeubi:
> Since it is deprecated since 2015 (8 years from now) do you think it is time to deprecate it for removal

do it :+1:

HannesWell:
> > Since it is deprecated since 2015 (8 years from now) do you think it is time to deprecate it for removal
> 
> do it 👍

Done in https://github.com/eclipse-equinox/equinox/pull/427.

PR #427, opened 2023-12-10
Deprecate PluginVersionIdentifier for removal and start removal process

HannesWell:
Now that all references to `PluginVersionIdentifier` are removed in the Eclipse SDK (as far as I can tell) and `PluginVersionIdentifier` has been deprecated since 2015, I suggest to now deprecate it for removal so that we can remove it in two years.

@tjwatson if you and everyone else is fine with it I will also add a corresponding entry to the porting guide.


vogella:
Lgtm

HannesWell:
Deprecation notice added with
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1678

vogella:
Porting guide is optional, if you use the for removal in the annotation, that is sufficient

PR #429, opened 2023-12-11
IProgressMonitor: submonitor(int, ProgressOption...) API #392

jukzi:
and deprecate IProgressMonitorWithBlocking for removal

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
@@ -254,6 +254,43 @@ public interface IProgressMonitor {
 	}
 
 	/**
+	 * Creates a new IProgressMonitor that will consume the given number of ticks
+	 * from this IProgressMonitor.
+	 * 
+	 * This is meant as an easy to use replacement for deprecated
+	 * {@code org.eclipse.core.runtime.SubProgressMonitor.SubProgressMonitor(IProgressMonitor, int, int)}.
+	 * Unlike {link {@link #slice(int)} the returned IProgressMonitor is meant for
+	 * use in the current thread and normally propagate calls to the parent (depends
+	 * on the used options).
+	 * 
+	 * @param totalWork number of ticks to consume from the receiver
+	 * @parma options optional ProgressOption
+	 * @return new IProgressMonitor that may be used in place of a new
+	 *         IProgressMonitor
+	 * @since 3.19
+	 */
+	public default IProgressMonitor submonitor(int totalWork, ProgressOption... options) {
+		return SubMonitor.create(this, totalWork, options);
+	}
+
+	/**
+	 * Options for {@link #submonitor(int, ProgressOption...)}
+	 * 
+	 * @since 3.19
+	 */
+	public static enum ProgressOption {
+		/**
+		 * Calls to <code>subTask</code> should not have any effect. This is equivalent
+		 * to {@link SubMonitor#SUPPRESS_SUBTASK}
+		 */
+		SUPPRESS_SUBTASK,
+		/**
+		 * The main task label should be prepended to the subtask label.
+		 */
+		PREPEND_MAIN_LABEL,
+	}
+
+	/**
 	 * Calls {@link #done()} on the given monitor if is non-null. If the given
 	 * monitor is null, this is a no-op.
 	 * 
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
@@ -206,6 +206,7 @@ import org.eclipse.core.internal.runtime.TracingOptions;
  * 
  * @since org.eclipse.equinox.common 3.3
  */
+@SuppressWarnings({ "removal", "deprecation" })
 public final class SubMonitor implements IProgressMonitorWithBlocking {
 
 	/**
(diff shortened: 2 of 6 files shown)


bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
```diff
@@ -253,6 +253,37 @@ public default IProgressMonitor slice(int work) {
...
+	 * Creates a new IProgressMonitor that will consume the given number of ticks
+	 * from this IProgressMonitor.
+	 * 
+	 * @param totalWork number of ticks to consume from the receiver
+	 * @parma options optional ProgressOption
+	 * @return new IProgressMonitor that may be used in place of a new
+	 *         IProgressMonitor
+	 * @since 3.19
+	 */
+	public default IProgressMonitor create(int totalWork, ProgressOption... options) {
```
laeubi, line 266: Create sounds strange here, can we maybe name this simply 
```suggestion
	public default IProgressMonitor submonitor(int totalWork, ProgressOption... options) {
```

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
```diff
@@ -784,6 +785,38 @@ public SubMonitor newChild(int totalWork) {
...
+		}
+		if (options.length == 1 && options[0] == ProgressOption.SUPPRESS_SUBTASK) {
+			return newChild(totalWork, SUPPRESS_BEGINTASK | SUPPRESS_SUBTASK);
+		}
+		// use default implementation for ProgressOption.PREPEND_MAIN_LABEL:
+		return create(this, totalWork, options);
+	}
+
+	static IProgressMonitor create(IProgressMonitor mon, int totalWork, ProgressOption[] options) {
+		if (options.length == 0) {
```
laeubi, line 801: As a very first step one should check here if the passed in IProgressMonitor is already an instance of submonitor and then transform this into a split.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
```diff
@@ -253,6 +253,37 @@ public default IProgressMonitor slice(int work) {
 		return new SlicedProgressMonitor(this, work);
 	}
 
+	/**
+	 * Creates a new IProgressMonitor that will consume the given number of ticks
+	 * from this IProgressMonitor.
```
laeubi, line 258: I think we should add some note here about what the difference to slice (e.g. usually can not be used in MT environments) and if it is safe to call begin task on the new instance.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
```diff
@@ -253,6 +253,37 @@ public default IProgressMonitor slice(int work) {
...
+	public default IProgressMonitor create(int totalWork, ProgressOption... options) {
+		return SubMonitor.create(this, totalWork, options);
+	}
+
+	/**
+	 * Options for {@link #create(int, ProgressOption...)}
+	 * 
+	 * @since 3.19
+	 */
+	public static enum ProgressOption {
```
laeubi, line 275: what about the other ones here?

grafik

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
```diff
@@ -784,6 +785,38 @@ public SubMonitor newChild(int totalWork) {
...
+	@Override
+	public IProgressMonitor create(int totalWork, ProgressOption... options) {
+		if (options.length == 0) {
+			return newChild(totalWork);
+		}
+		if (options.length == 1 && options[0] == ProgressOption.SUPPRESS_SUBTASK) {
+			return newChild(totalWork, SUPPRESS_BEGINTASK | SUPPRESS_SUBTASK);
+		}
+		// use default implementation for ProgressOption.PREPEND_MAIN_LABEL:
+		return create(this, totalWork, options);
```
laeubi, line 797: Can we please just implement `PREPEND_MAIN_LABEL_TO_SUBTASK` in SubMonitor? It should not be to hard to implement and I think such change should prove we do not need to rely on deprecated API and all can be implemented by SubMonitor itself.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
jukzi: ok

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
jukzi: thats not needed: In the case of Submonitor the overload org.eclipse.core.runtime.SubMonitor.submonitor(int, ProgressOption...) has already checked that.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
jukzi: i will add something

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
jukzi: I don't know any usecase who needs those. If anybody needs some of those the API is extendable for new Options.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
laeubi: Okay its a bit confusing can this method maybe moved to "Sub**Progress**Monitor" or even just add this impl to the default of `IProgressMonitor` to make it more clear that this is actually nothing that belongs to `SubMonitor`?

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
jukzi: If anybody wants to add such an optimization that could be done in future commits. I see no value in making the implementation more complicated.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
jukzi: > default of `IProgressMonitor` to make it more clear that this is actually nothing that belongs to `SubMonitor`?

I started with that, but its a bit confusing to call the default implementation fallback from SubMonitor since its super interface is deprecated. As soon as IProgressMonitorWithBlocking is removed that would be the better way.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/IProgressMonitor.java
laeubi: That's not an argument, these are options of `SubMonitor` so they should be supported that people can call this method as a replacement maybe and `SubMonitor` can return best matching child instead of using competing APIs e.g. `SUPPRESS_ISCANCELED` is used in `org.eclipse.jdt.core.refactoring.CompilationUnitChange` , `SUPPRESS_ALL_LABELS` (what includes `SUPPRESS_SETTASKNAME`) is used in `org.eclipse.equinox.p2.operations.RemediationOperation` ... other projects might use others as well.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SubMonitor.java
laeubi: > I started with that, but its a bit confusing to call the default implementation fallback from SubMonitor since its super interface is deprecated. 

As mentioned before, `SubMonitor` should never need to call any default implementation it should fully implement the API itself so we can prove the new API is actually implementable (without rely on old deprecated API).

PR #430, opened 2023-12-12
SlicedProgressMonitor: fix javadoc

jukzi:


jukzi:
there are only unrelated TCK fails, please review/submit

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SlicedProgressMonitor.java
```diff
@@ -14,8 +14,9 @@
 package org.eclipse.core.runtime;
 
 /**
- * default implementation of a {@link SlicedProgressMonitor} that synchronizes
- * on the given monitor to report work
+ * default implementation for
+ * {@link org.eclipse.core.runtime.IProgressMonitor#slice(int)} that
```
laeubi, line 18: It is a public class that can be used in other context as well so I don't think its correct to say it is for THAT implementation of this specific method...

jukzi: at the moment the javadoc says SlicedProgressMonitor implments SlicedProgressMonitor - which is a nobrainer

laeubi: Well in fact it says

> default implementation of a SlicedProgressMonitor that ...

you can remove the link if you want and maybe reword it as

> implementation of a ProgressMonitor that handles sliced work by ...

... or similar, but it can be used by anyone or any implementation and even can be extended an is not only for `IProgressMonitor#slice(int)` default impl.

PR #431, opened 2023-12-12
Path: use private getSegments()

jukzi:
this is a preparational refactoring for a possible different datastructure for the segments


jukzi:
There are only known TCK fails. Mind to reviev/submit any commiter, please?

PR #432, opened 2023-12-12
Path: intern new Strings

jukzi:
When a filename is read from filesystem the time to intern() is not significant compared to the IO time but can save memory in the workspace

in case when i got an OOME it could have saved ~15MB
image

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java
@@ -184,7 +184,7 @@ public final class Path implements IPath, Cloneable {
 		String devicePart = null;
 		int pathLength = pathString.length();
 		if (firstMatch == pathLength || pathString.charAt(firstMatch) != DEVICE_SEPARATOR) {
-			devicePart = pathString.substring(0, firstMatch);
+			devicePart = pathString.substring(0, firstMatch).intern();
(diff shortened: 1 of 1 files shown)


merks:
I imagine that here, like with EMF's URIs, massive duplication of strings are involved!   

Not that it's a huge concern, but are interned strings actually garage collectable in modern VMs?  (Sorry if that was answered elsewhere already.)

jukzi:
> are interned strings actually garage collectable in modern VMs

https://github.com/eclipse-equinox/equinox/pull/33#issuecomment-1845910299
image

jukzi:
There are only known TCK fails. Mind to reviev/submit any commiter, please?

iloveeclipse:
Jörg, have you measured build time differences? We trade time vs memory here, and Path is heavily used during build.

jukzi:
> Jörg, have you measured build time differences? We trade time vs memory here, and Path is heavily used during build.

good that you asked: i checked and stumbled on a very ineffective org.eclipse.core.internal.watson.ElementTreeIterator.requestPath() which uses String concatenation just to split them afterwards again, which used intern():
image

jukzi:
Let's drop this: it also interns ephemeral Paths
https://github.com/eclipse-platform/eclipse.platform/commit/4e0f4a4911cf90fbd6a31261c34f23ae82e5b445 should be of much more help and we can investigate newer dumps for more potential - if needed at all.

PR #437, opened 2023-12-15
Deprecate (and mark for removal) HttpContextExtensionService

laeubi:
The HttpContextExtensionService is effectively unused in platform and uses the abandoned HttpService it is API.

We therefore should deprecate and remove that so the org.eclipse.equinox.http.registry bundle becomes free from any API and only contains internal implementation what would make us free to change how it works,

### DIFF
--- a/bundles/org.eclipse.equinox.http.registry/src/org/eclipse/equinox/http/registry/HttpContextExtensionService.java
+++ b/bundles/org.eclipse.equinox.http.registry/src/org/eclipse/equinox/http/registry/HttpContextExtensionService.java
@@ -27,6 +27,7 @@ import org.osgi.service.http.HttpService;
  * @noimplement This interface is not intended to be implemented by clients.
  * @noextend This interface is not intended to be extended by clients.
  */
+@Deprecated(forRemoval = true)
 public interface HttpContextExtensionService {
 	/**
 	 * returns the HttpContext associated with the HttpService reference and http


vogella:
+1

tjwatson:
Should check for usage from RAP @mknauer

Note this would also have to remove the extension point:

```
     <extension-point id="httpcontexts" name="%httpcontextsName" schema="schema/httpcontexts.exsd"/>
```

I find it odd that we deprecate this one extension point when all other extension points in `org.eclipse.equinox.http.registry` are also dependent on the old http service.

laeubi:
> I find it odd that we deprecate this one extension point when all other extension points in org.eclipse.equinox.http.registry are also dependent on the old http service.

As long as the extension point do not depend on `org.osgi.http`  package I don't think they depend on http service as the impl can choose any technique (including whiteboard http service)

HannesWell:
In general I fully support deprecating and eventually removing unused elements especially when they cause problems like `o.o.service.http`.
Nevertheless in this case I'm not familiar with the class to deprecate, but from looking at it I wonder why the only method of the interface `HttpContextExtensionService.getHttpContext()` is never called (at least Eclipse does not give me a caller)?

> > I find it odd that we deprecate this one extension point when all other extension points in org.eclipse.equinox.http.registry are also dependent on the old http service.
> 
> As long as the extension point do not depend on `org.osgi.http` package I don't think they depend on http service as the impl can choose any technique (including whiteboard http service)

The extension point contains
https://github.com/eclipse-equinox/equinox/blob/a72b8ecfe2b501d811fc77ca918965bce233a8b8/bundles/org.eclipse.equinox.http.registry/schema/httpcontexts.exsd#L69
so I would say it depends on `org.osgi.service.http`.

And to give a broader dependencies overview, in the SimRel only RAP and WST depend on equinox.http.registry:

grafik

laeubi:
>  but from looking at it I wonder why the only method of the interface HttpContextExtensionService.getHttpContext() is never called (at least Eclipse does not give me a caller)?

That's what I written above:

> The `HttpContextExtensionService` is effectively **unused** in platform

HannesWell:
> > but from looking at it I wonder why the only method of the interface HttpContextExtensionService.getHttpContext() is never called (at least Eclipse does not give me a caller)?
> 
> That's what I written above:
> 
> > The `HttpContextExtensionService` is effectively **unused** in platform

OK, but before deprecating an interface usually all references to it in our code-base should be removed.
@laeubi can you please do the corresponding changes/migrations in this PR so that when the interface is eventually removed nothing else has to be changed.

PR #438, opened 2023-12-15
Deprecate (and mark for removal) ExtendedHttpService

laeubi:
The ExtendedHttpService is effectively unused in platform and uses the abandoned HttpService it is API.

We therefore should deprecate and remove that to get rid of the dependency here a more modern replacement for the purpose of this extended form (registering filters) is today the [Http Whiteboard Specification](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.http.whiteboard.html)

### DIFF
--- a/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/ExtendedHttpService.java
+++ b/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/ExtendedHttpService.java
@@ -25,6 +25,7 @@ import org.osgi.service.http.*;
  * @noimplement This interface is not intended to be implemented by clients.
  */
 @ProviderType
+@Deprecated(forRemoval = true)
 public interface ExtendedHttpService extends HttpService {
 
 	/**
--- a/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/HttpServiceMultipartServlet.java
+++ b/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/HttpServiceMultipartServlet.java
@@ -20,6 +20,7 @@ import javax.servlet.http.HttpServlet;
  * @deprecated No longer required.
  * @since 1.3
  */
+@Deprecated(forRemoval = true)
 public class HttpServiceMultipartServlet extends HttpServlet {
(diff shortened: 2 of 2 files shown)


vogella:
+1

tjwatson:
> The ExtendedHttpService is effectively unused in platform and uses the abandoned HttpService it is API.

See `org.eclipse.equinox.http.registry.internal.HttpRegistryManager.registerFilter(FilterContribution)`

laeubi:
> org.eclipse.equinox.http.registry.**internal**.HttpRegistryManager

I highlighted the important part ;-)

tjwatson:
> > org.eclipse.equinox.http.registry.**internal**.HttpRegistryManager
> 
> I highlighted the important part ;-)

That is not helpful.  You asserted that `ExtendedHttpService` is not used by the project.  I pointed out that it is used by `HttpRegistryManager` which IS the internal implementation behind the public API extension point `<extension-point id="filters" name="%filtersName" schema="schema/filters.exsd"/>` from the `org.eclipse.equinox.http.registry` bundle.

laeubi:
The point is that we do not need a **public** API to implement **HttpRegistryManager** it could be a private one, so if no one except internal implementations is calling that API, and if you look at the implementation it does not even use that API at all, it looks if the service and checks for a method that has a special signature, so as long as the EquinoxHttp service has a public method it will still work, it does not need it to be mentioned in a public API interface.

laeubi:
@tjwatson @vogella how should we proceed here, my goal would be to make `ExtendedHttpService` an internal implementation detail of Equinox, so no "external" consumers exits anymore (e.g. we can move it into an internal package). WDYT?

laeubi:
I now started migration of the HelpCenter to using the Whiteboardservices, any help is welcome, it is always the same pattern and mostly one needs to find out where the servlet it used to test is is working properly:

- https://github.com/eclipse-platform/eclipse.platform/pull/2023
- https://github.com/eclipse-platform/eclipse.platform/pull/2014

laeubi:
> What about https://github.com/eclipse-platform/eclipse.platform/blob/eb38d0fdee43e96d061efa13903753ecfa206ed7/ua/org.eclipse.help.webapp/plugin.xml#L110 ?

That's a bit special but also doable I'll look into this hopefully later on, for now it would be good to convert the easy ones first :-)

PR #439, opened 2023-12-15
Convert to lambda in o.e.osgi

akurtakov:


akurtakov:
Yet another iteration on https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1659

HannesWell:
In the past (pre GitHub time) I did similar clean-ups in a broader scope but it had to be reverted because it led to a worse performance on Java 8 which was not acceptable for @tjwatson.

I think a single place doesn't hurt much, but if done on more places it could become relevant again. :/

tjwatson:
These look fine.  They are low traffic calls so I don't expect it to impact performance much.

For the record, it seemed that the JIT compiler could not optimize the lambda reference nearly as much as the anonymous class.  I don't remember the specific details but was told by the Java team that they do expect the JIT to get better in that regards at some point, but I do not think it has gotten up to the optimizations of anonymous class compilation speeds.

PR #441, opened 2023-12-15
Fix the error that CoreExcepton is indirectly required

merks:
The org.eclipse.core.tests.harness.PerformanceTestRunner is not optional.


merks:
I see this problem in my workspace before this change:

image

PR #443, opened 2023-12-16
Remove unused dependencies to o.o.service.http and javax.servlet

HannesWell:
The PDE dependencies analysis says these dependencies are not used and I also could not find any textual reference to the packages removed (e.g. for an reflective class-loading).
If the build succeeds we should remove the dependencies, especially to problematic packages like `org.osgi.service.http`.


merks:
There is a non-zero chances that the help will stop working.   But I can't seem to fetch the changes:

image

I guess the test failures are a problem too...

HannesWell:
> There is a non-zero chances that the help will stop working. But I can't seem to fetch the changes:

Did you select the right remote, i.e. the one that points to this repo and not your fork when fetching?

HannesWell:
> I guess the test failures are a problem too...

I investigate them, luckily they are reproducible in the IDE.

merks:
The Help -> Help Contents comes up like this in a debug launch workspace:

image

Lots of logged exception.

The JSP support is rather fragile.

merks:
Fetching again fails like this now:

image

HannesWell:
The removal of the optional dependencies of `org.eclipse.equinox.http.servlet` to `javax.servlet.descriptor` was what let many test cases fail.
Lets see if this already fixes all.

merks:
FYI, the BundleProxyClassLoader is partly what makes this stuff fragile.  When migrating to newer dependencies, I had to hack this class...

HannesWell:
@merks thanks for your help in testing the help.
Can you check if adding back the optional dependency `javax.servlet.descriptor` to `org.eclipse.equinox.jsp.jasper` fixes the problem?

laeubi:
> Can you check if adding back the optional dependency `javax.servlet.descriptor` to `org.eclipse.equinox.jsp.jasper`

Obviously it is not *that* optional at all, can we make it a real dependency?

HannesWell:
I think I understood the stack-trace and hope/am confident that adding back `javax.servlet.descriptor` should fix it.

> > Can you check if adding back the optional dependency `javax.servlet.descriptor` to `org.eclipse.equinox.jsp.jasper`
> 
> Obviously it is not _that_ optional at all, can we make it a real dependency?

With what has been noticed here yes, but before I just described what the Manifest said. :)
But I agree it seems to be wrong.

HannesWell:
All tests pass now and if I start an Eclipse from my workspace and open the help via `Help -> Help Contents` everything seems to work without errors.
@merks can you confirm this?



> > Can you check if adding back the optional dependency `javax.servlet.descriptor` to `org.eclipse.equinox.jsp.jasper`
> 
> Obviously it is not _that_ optional at all, can we make it a real dependency?

Made the dependency non-optional now.

merks:
Yes, that works for me.  👍

PR #449, opened 2023-12-18
Unify Configuration- and InstancePreferences into common super-class

HannesWell:
This avoids code duplication, which would be increased when similar 'singleton' scopes are implemented like in https://github.com/eclipse-equinox/equinox/pull/446.


HannesWell:
Is anybody interested in reviewing this?

merks:
It looks generally fine.

HannesWell:
Thank you Ed. I just added a few more braces and made fields final where possible.

Issue #191, opened 2023-01-20
org.eclipse.equinox.device seem to fail some TCK tests


laeubi:
Currently the org.eclipse.equinox.device seem to fail some TCK test see test file that can be opened in Eclipse:
[TEST-org.osgi.test.cases.device-8.1.0.202212101352.xml.zip](https://github.com/eclipse-equinox/equinox/files/10464046/TEST-org.osgi.test.cases.device-8.1.0.202212101352.xml.zip)

laeubi:
@tjwatson are you aware of any users of those service that might be interested to complete the implementation to make it comply to the TCK? Or should we probably deprecate / remove that implementation?

tjwatson:
> @tjwatson are you aware of any users of those service that might be interested to complete the implementation to make it comply to the TCK? Or should we probably deprecate / remove that implementation?

I am not aware of anyone that uses this.  I vote to deprecate. There have been no real fixes to this bundle in many years.  Anyone using it can continue to use the existing version already released.

PR #451, opened 2023-12-20
Deprecate the device service implementation

laeubi:
Fix https://github.com/eclipse-equinox/equinox/issues/191


tjwatson (approved):

tjwatson:
Is it expected that the Device Access TCK would still show up in the PR checks, but with `No tests found`?

laeubi:
> Is it expected that the Device Access TCK would still show up in the PR checks, but with `No tests found`?

It might be that it only vanishes when merged to master, will check that...

HannesWell:
Just created the N&N entry for this in combination with #18 as requested in https://github.com/eclipse-equinox/equinox/issues/18#issuecomment-1863900159.
- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/124

PR #454, opened 2023-12-21
Migrate debugoptions, hooks and resources in o.e.osgi.tests to JUnit 4

HeikoKlare:
This migrates all tests in `org.eclipse.osgi.tests.debugoptions`, `org.eclipse.osgi.tests.hooks.framework` and
`org.eclipse.osgi.tests.resource` to JUnit 4

* Add JUnit 4 annotations `@Test`, `@Before`,  @After`
* Remove inheritance of JUnit 3 TestCase and CoreTest

### DIFF
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/debugoptions/DebugOptionsTestCase.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/debugoptions/DebugOptionsTestCase.java
@@ -14,6 +14,14 @@
 
 package org.eclipse.osgi.tests.debugoptions;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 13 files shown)


bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/resource/BasicTest.java
```diff
@@ -457,12 +462,12 @@ private void assertNotRequirementMatches(Requirement requirement, Capability cap
 		assertFalse("Requirement matches capability", ((BundleRequirement) requirement).matches((BundleCapability) capability));
 	}
 
-	private void assertNotNull(Capability capability) {
-		Assert.assertNotNull("Null capability", capability);
+	private void assertNotNullCapabilities(Capability capability) {
+		assertNotNull("Null capability", capability);
 	}
```
HannesWell, line 467: IMHO it would also be suitable to just inline `assertNotNull(capability)`.
At least from my perspective these comments are often not that helpful because in most cases one has to check the code anyways.

HannesWell (commented):
Thank you for this too.

Looks good, just the addition of two unnecessary empty java-doc tags should be removed.

bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/resource/BasicTest.java
HeikoKlare: Fully agreed. I have simply removed the custom `assertNotNull` methods in 752ffd9.

bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/debugoptions/DebugOptionsTestCase.java
```diff
@@ -770,8 +814,11 @@ public void testTraceFile03() {
 
 	/**
 	 * test DebugTrace.traceEntry(option)
-	*/
-	public void testTraceFile04() {
+	 * 
+	 * @throws IOException
```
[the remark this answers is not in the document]
HeikoKlare: Sorry for adding these unnecessary tags and not properly cleaning up before submitting the PR. Maybe we could think about disabling this automatic javadoc changes for test bundles. It's also configured that way for some test bundles in platform.
FIxed in 752ffd9.

HannesWell (approved):
Looks good, thank you.

PR #455, opened 2023-12-21
Migrate resolver tests in o.e.osgi.tests to JUnit 4

HeikoKlare:
This migrates all tests in `org.eclipse.osgi.tests.resolver` and `org.eclipse.osgi.tests.services.resolver` to JUnit 4

* Add JUnit 4 annotations `@Test`, `@Before`, `@After`
* Remove inheritance of JUnit 3 `TestCase` and `CoreTest`

### DIFF
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/services/datalocation/BasicLocationTests.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/services/datalocation/BasicLocationTests.java
@@ -13,6 +13,14 @@
  *******************************************************************************/
 package org.eclipse.osgi.tests.services.datalocation;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeTrue;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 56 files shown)


bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/services/datalocation/BasicLocationTests.java
```diff
@@ -86,10 +96,17 @@ private void checkLocation(Location location, boolean leading, boolean trailing,
...
-		if (windows)
+		if (OS.isWindows())
 			assertTrue(url.toExternalForm() + " should " + (leading ? "" : "not") + " have a leading slash", url.getFile().startsWith("/") == leading);
 	}
 
+	private void fail(String message, Throwable exception) {
+		AssertionFailedError error = new AssertionFailedError(message);
+		error.initCause(exception);
+		throw error;
+	}
```
HeikoKlare, line 107: fyi: this is a reduced version of the method that was inherited from `CoreTest` before. I kept it instead of replacing all the `fail` calls as the exceptions in those cases are part of the functionality-under-test, so they should make the test fail rather than error on occurrence (see https://github.com/eclipse-equinox/equinox/pull/453#discussion_r1434420881).

bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/services/datalocation/BasicLocationTests.java
```diff
@@ -142,19 +159,21 @@ public void testCreateLocation03() {
...
+	@Test
 	public void testCreateLocation04() {
 		Location configLocation = configLocationTracker.getService();
 		File testLocationFile = OSGiTestsActivator.getContext().getDataFile("testLocations/testCreateLocation04");
 		Location testLocation = configLocation.createLocation(null, null, true);
 		try {
 			testLocation.set(testLocationFile.toURL(), true);
 			testLocation.release();
-			fail("Should not be able to lock read-only location");
+			Assert.fail("Should not be able to lock read-only location");
```
HeikoKlare, line 170: Could not be replaced with an `assertThrows` as this test actually fails silently, see #460

HeikoKlare:
> This pull request skips 1 test.
> ```
> org.eclipse.osgi.tests.services.datalocation.BasicLocationTests ‑ testUNC
> ```

Skipped test is because the conditional return was replaced with a JUnit 4 `assume` statement.

akurtakov (approved):

HannesWell:
Thank you Alex for taking over and of course to Heiko.

PR #456, opened 2023-12-21
Make session tests in o.e.osgi.tests independent from OSGiTest/CoreTest

HeikoKlare:
Several session test classes in `org.eclipse.osgi.tests` are subclasses of either `CoreTest` or the `OSGiTest` subclass of `CoreTest`. This change make the test classes directly inherit from JUnit 3's `TestCase` to become independent of the `CoreTest` class. The `TestCase` inheritance is necessary until session tests do not rely on JUnit 3 anymore.

* Remove inheritance of `CoreTest` and replace `OSGiTest` inheritance with static imports of inherited helper methods

### DIFF
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/appadmin/ApplicationAdminTest.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/appadmin/ApplicationAdminTest.java
@@ -13,15 +13,20 @@
  *******************************************************************************/
 package org.eclipse.osgi.tests.appadmin;
 
+import static org.eclipse.osgi.tests.OSGiTest.PI_OSGI_TESTS;
+import static org.eclipse.osgi.tests.OSGiTest.addRequiredOSGiTestsBundles;
+import static org.eclipse.osgi.tests.OSGiTestsActivator.getContext;
+
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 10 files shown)


HannesWell:
> This change make the test classes directly inherit from JUnit 3's `TestCase` to become independent of the `CoreTest` class. The `TestCase` inheritance is necessary until session tests do not rely on JUnit 3 anymore.

From your comment in the other PR I assume the migration to Junit-4 is too much work respectively too different to a JUnit-5 pendant so that they should be migrated to Junit-5 directly?

HeikoKlare:
> From your comment in the other PR I assume the migration to Junit-4 is too much work respectively too different to a JUnit-5 pendant so that they should be migrated to Junit-5 directly?

That's my expectation, yes. I will try to figure that out in an attempt to migrate the session test framework. First thing to do for the migration of session tests is the migration of the framework (`SessionTest`, `SessionTestSuite` and all its derivatives), as they assume a specific (JUnit-3-conforming) structure of the session test classes. So the migrating of the classes in Equinox affected by this PR to either JUnit 4 or 5 first requires an update of that framework in the Platform. This PR only removes the unnecessary inheritance hierarchy between `TestCase` and the actual test class to (1) allow the removal of `CoreTest` in Platform and (2) prepare for a later JUnit migration in which complex inheritance hierarchies are a problem anyway.

akurtakov (approved):

PR #457, opened 2023-12-21
Migrate several test classes in o.e.osgi.tests to JUnit 4

HeikoKlare:
This migrates all tests to JUnit 4 that are in packages containing only a single test class or only a single test not yet migrated to JUnit 4.

* Add JUnit 4 annotations `@Test`, `@Before`, `@After`
* Remove inheritance of JUnit 3 `TestCase` and `CoreTest`
* Adapt performance tests to JUnit-4-conforming API


HannesWell (approved):
Looks good and number of tests executed remained the same.

Thank you.

PR #458, opened 2023-12-21
Remove abstract OSGiTest class and dependency to CoreTest 

HeikoKlare:
This removes all dependencies to the to-be-removed JUnit 3-specific CoreTest class. To this end, the OSGiTest class, which only contains utility methods, is removed and the remaining utility fields and methods moved to the existing OSGiTestActivator class.


akurtakov:
@HeikoKlare You should be able to continue here as all other PRs are merged in

HeikoKlare:
Thanks, @akurtakov! I've update this PR accordingly.

akurtakov (approved):

PR #462, opened 2023-12-23
Remove now unnecessary calls to String.intern() in PreferencesService

HannesWell:
The preference keys on which `String.intern()` is called are already interned when they are put into the EclipsePreferences object since https://github.com/eclipse-equinox/equinox/pull/33.

Not calling intern again avoids the overhead of it.

### DIFF
--- a/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
+++ b/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
@@ -132,13 +132,8 @@ public class PreferencesService implements IPreferencesService {
 				List<String> propsToRemove = new ArrayList<>(Arrays.asList(globalNode.keys()));
 				if (keys.length > 0) {
 					for (String key : keys) {
-						// preferences that are not in the applied node
-						// will be removed
+						// preferences that are not in the applied node will be removed
 						propsToRemove.remove(key);
-						// intern strings we import because some people
-						// in their property change listeners use identity
-						// instead of equals. See bug 20193 and 20534.
-						key = key.intern();
 						String value = node.get(key, null);
 						if (value != null) {
 							if (EclipsePreferences.DEBUG_PREFERENCE_SET) {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
```diff
@@ -149,8 +144,10 @@ public IStatus applyPreferences(IExportedPreferences preferences) throws CoreExc
 						}
 					}
 				}
+				if (!propsToRemove.isEmpty() && !(globalNode instanceof EclipsePreferences)) {
+					propsToRemove.replaceAll(String::intern);
+				}
```
HannesWell, line 149: The elements in the keys array above are obtained from an `ExportedPreferences`, which is a subclass of `EclipsePreference` and therefore has its keys already interned (unless somebody extends this internl class and changes the method to store elements, but those that do that should take care of interning themself).

On the other hand `globalNode` is just a `IEclipsePreferences` node from the tree and because `Clients may implement this interface`, it can be an arbitrary implementation. Therefore interning  `propsToRemove` can only be skipped if the node is of a `EclipsePreferences` subclass.

bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
```diff
@@ -149,8 +144,12 @@ public IStatus applyPreferences(IExportedPreferences preferences) throws CoreExc
 						}
 					}
 				}
+				if (!propsToRemove.isEmpty() && !(globalNode instanceof EclipsePreferences)) {
+					// intern strings we import because some people in their property change
+					// listeners use identity instead of equals. See bug 20193 and 20534.
+					propsToRemove.replaceAll(String::intern);
+				}
```
HannesWell, line 151: The elements in the keys array above are obtained from an `ExportedPreferences`, which is a subclass of `EclipsePreference` and therefore has its keys already interned (unless somebody extends this internl class and changes the method to store elements, but those that do that should take care of interning themself).

On the other hand `globalNode` is just a `IEclipsePreferences` node from the tree and because `Clients may implement this interface`, it can be an arbitrary implementation. Therefore interning  `propsToRemove` can only be skipped if the node is of a `EclipsePreferences` subclass.

In general I would call the listeners that rely on string identity comparison badly implemented.

PR #463, opened 2023-12-23
Clean-up and simplify Preferences (mainly add braces)

HannesWell:


bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
```diff
@@ -868,17 +899,18 @@ private boolean scopeMatches(String scope, IEclipsePreferences tree) {
...
 	public void setDefaultLookupOrder(String qualifier, String key, String[] order) {
 		String registryKey = getRegistryKey(qualifier, key);
-		if (order == null)
+		if (order == null) {
 			DEFAULTS_REGISTRY.remove(registryKey);
-		else {
-			LookupOrder obj = new LookupOrder(order);
-			DEFAULTS_REGISTRY.put(registryKey, obj);
+		} else {
+			Arrays.asList(order).forEach(Objects::requireNonNull);
```
HeikoKlare, line 905: @HannesWell I think this introduced a regression, because the method now throws a `NullPointerException` rather than an `IllegalArgumentException` in case any of the `order` entries is null. This is incompatible with the method's documentation and makes the Platform test case `org.eclipse.core.tests.internal.preferences.PreferencesServiceTest.testLookupOrder()` fail.

I have documented the regression in #464.

Issue #464, opened 2024-01-04
PreferenceService#setDefaultLookupOrder() throws wrong exception


HeikoKlare:
The method `org.eclipse.core.internal.preferences.PreferencesService.setDefaultLookupOrder(String, String, String[])` throws an invalid exception since #463.
The method's documentation specifies that an `IllegalArgumentException` is thrown if any of the order entries is `null`:
https://github.com/eclipse-equinox/equinox/blob/d52b39e8c8407a6b8e9c554df917bbb30407bb3f/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/runtime/preferences/IPreferencesService.java#L676-L683

but since the changes of #463 it throws a `NullPointerException` in that case:
https://github.com/eclipse-equinox/equinox/blob/d52b39e8c8407a6b8e9c554df917bbb30407bb3f/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java#L904

The modified behavior makes the Platform test case `org.eclipse.core.tests.internal.preferences.PreferencesServiceTest.testLookupOrder()` fail, see, e.g.: https://github.com/eclipse-platform/eclipse.platform/actions/runs/7406826268/job/20152527917

From my point of view, the old behavior should be restored (as the only alternative would be to change the API (documentation)).

PR #465, opened 2024-01-04
Proper exception in PreferencesService#setDefaultLookupOrder() #464

HeikoKlare:
`PreferencesService#setDefaultLookupOrder(String, String, String[])` has recently been changed to now throwing a `NullPointerException` instead of an `IllegalArgumentException` in case of any of its order arguments being `null`. This is incompatible with the API documentation in `IPreferencesService`.
This change restores the existing behavior of throwing an `IllegalArgumentException` in the mentioned case.

Fixes https://github.com/eclipse-equinox/equinox/issues/464

### DIFF
--- a/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
+++ b/bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
@@ -901,7 +901,9 @@ public class PreferencesService implements IPreferencesService {
 		if (order == null) {
 			DEFAULTS_REGISTRY.remove(registryKey);
 		} else {
-			Arrays.asList(order).forEach(Objects::requireNonNull);
+			if (Arrays.stream(order).anyMatch(Objects::isNull)) {
+				throw new IllegalArgumentException();
+			}
 			DEFAULTS_REGISTRY.put(registryKey, order);
 		}
 	}


bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
```diff
@@ -901,7 +902,9 @@ public void setDefaultLookupOrder(String qualifier, String key, String[] order)
 		if (order == null) {
 			DEFAULTS_REGISTRY.remove(registryKey);
 		} else {
-			Arrays.asList(order).forEach(Objects::requireNonNull);
+			if (Stream.of(order).anyMatch(Objects::isNull)) {
```
HannesWell, line 905: What do you think about using `Arrays.stream()` directly which is called by `Stream.of()`?
At least for me it makes it more clear, that it is a stream of elements of an array and not a stream with a single element order (although this would not be possible with the var-args argument).
```suggestion
			if (Arrays.stream(order).anyMatch(Objects::isNull)) {
```

HannesWell (approved):
Thank you @HeikoKlare for fixing this regression. The change looks good and reasonable.
I just have one tiny nit-pick.

bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
HeikoKlare: Sure, totally makes sense. I was not aware of that method in `Arrays` but always used the `Stream` operation. Thanks for the hint!

iloveeclipse:
Thanks Heiko!

bundles/org.eclipse.equinox.preferences/src/org/eclipse/core/internal/preferences/PreferencesService.java
HannesWell: Great. Thanks for the update!

PR #467, opened 2024-01-04
Use submodule to include the equinox.binaries

laeubi:
Currently the user needs to checkout two repositories side-by-side and specify the location of it at the commandline to fully build equinox.

This includes the other repository as a submodule now to make it easier for users/buildscripts as they can see the content of both repositories as one linear directory structure.


iloveeclipse:
Can that have negative effect on platform aggregator / SDK build (including all the scripts/jobs) - they assume  all "other" repositories to be "usual" git repos, and may fail to perform proper git commands to checkout right state?

laeubi:
> Please verify that releng.aggregator repo works with this change. I assume some scripts will need adjustments for submodules.

Will do once it is working here, as the aggregator build already uses submodules a lot I don expect anything specifically here but who knows :-)

laeubi:
Build is now working, so next would be to test it with the aggregator build....

laeubi:
Testbuild is here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1701

HannesWell:
Did you verify that building the natives still works?
I haven't looked into the build config yet, so cannot tell if including the binaries repo as sub-module is really relevant, but we should be sure.

laeubi:
The binaries should never be build here, the binary repos is only used to copy the files into the feature.

iloveeclipse:
Shouldn't we use git-lfs instead? That would make everything even easier.

laeubi:
> Shouldn't we use git-lfs instead? That would make everything even easier.

I won't mind [but last time git-lfs makes "everything easier"](https://github.com/eclipse-platform/eclipse.platform.swt/discussions/497) it took > 1 year to do so.

iloveeclipse:
This is irrelevant, discussion took so long because there were no one to work on, not because git-lfs.
We are talking about decisions that have long term effects.
From my point of view same criteria apply here as for SWT.
So instead of changing to git modules and still have multiple repos, we could simplify that and have one repo.

laeubi:
> This is irrelevant, discussion took so long because there were no one to work on, not because git-lfs.

So can you work on merging both repos using git-lfs? I have only limited knowledge about equinox binary repos and can only tell this one wants to copy a file that is there to execute the maven build, while for SWT there was a lot more involved, on the other hand the binaries here are only build on request...

So for me any solution would sufficient as long as it is performed.

HannesWell:
This is already discussed in https://github.com/eclipse-equinox/equinox.binaries/issues/1.

I think it would be good to continue there.

I'll try to have a look at making the binaries compile on demand automatically like SWT in the equinox pipeline (although this is somewhat orthogonal).

laeubi:
> I'll try to have a look at making the binaries compile on demand automatically like SWT in the equinox pipeline (although this is somewhat orthogonal).

Before butting effort in it please note that the goal is not to compile the binaries as part of the build, they are changing very infrequently (sometimes not for years) and therefore seems a bit waste of CI resources to compile them with each change.

PR #475, opened 2024-01-12
Version bump of launcher fragments

akurtakov:
This rendered the build unstable - fail with comparator errors. It comes due to natives rebuild for
https://github.com/eclipse-equinox/equinox/pull/474 .

### DIFF
--- a/bundles/org.eclipse.equinox.launcher.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.launcher.cocoa.macosx.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.equinox.launcher.cocoa.macosx.aarch64;singleton:=true
-Bundle-Version: 1.2.800.qualifier
+Bundle-Version: 1.2.900.qualifier
 Fragment-Host: org.eclipse.equinox.launcher;bundle-version="[1.5.0,1.7.0)"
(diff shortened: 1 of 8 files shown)


akurtakov:
Also reported at https://download.eclipse.org/eclipse/downloads/drops4/I20240112-0200/buildlogs/reporeports/reports/versionChecks.html .
Once this one is in it would be best validated via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1711 which failed with this error.
I believe it's high time to change the I-build to fail on missing version bump immediately instead of having to wait for hours to figure smth like that - see the PR failed at 20 mins due to that error.
@merks @iloveeclipse @laeubi wdyt about ^?

iloveeclipse:
> I believe it's high time to change the I-build to fail on missing version bump immediately instead of having to wait for hours to figure smth like that - see the PR failed at 20 mins due to that error.

Agree. But better if we would not need to detect missing version bumps in fragments by catching this at equinox already. So why don't we fail in equinox?

laeubi:
Rebuilding launchers is a separate job (that *might* can fail not sure) so you won't see any errors unless you build **equinox** after rebuilding **launchers**, in that case it would probably be already reported...

akurtakov:
> > I believe it's high time to change the I-build to fail on missing version bump immediately instead of having to wait for hours to figure smth like that - see the PR failed at 20 mins due to that error.
> 
> Agree. But better if we would not need to detect missing version bumps in fragments by catching this at equinox already. So why don't we fail in equinox?

Because the build is done by bot which commits in two repos and if the binaries repo (which contains only dll/so files) gets new content we can't afford to not update the buildable repo. One it was in the maser build of equinox failed https://ci.eclipse.org/equinox/job/equinox/job/master/294/ .

If/when equinox.binaries repo is gone maybe it could be made to open a PR rather than push directly so it's detected.
There is also the question whether equinox verification job (and all others actually) should send mails to respective dev lists on master build failures as this would give us one more change to catch such issue earlier.

PR #478, opened 2024-01-12
Replace <tt> by <code> tags in javadoc

laeubi:
See 
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1533

### DIFF
--- a/bundles/org.eclipse.equinox.concurrent/src/org/eclipse/equinox/concurrent/future/IFuture.java
+++ b/bundles/org.eclipse.equinox.concurrent/src/org/eclipse/equinox/concurrent/future/IFuture.java
@@ -58,8 +58,8 @@ public interface IFuture<ResultType> {
 	/**
 	 * Cancel the operation
 	 * 
-	 * @return <tt>false</tt> if the operation could not be canceled, typically
-	 *         because it has already completed normally; <tt>true</tt> otherwise
+	 * @return <code>false</code> if the operation could not be canceled, typically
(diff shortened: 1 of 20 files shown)


bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Logger.java
```diff
@@ -22,7 +22,7 @@
 
 /**
  * <p>
- * This class mimics the standard OSGi <tt>LogService</tt> interface. An
+ * This class mimics the standard OSGi <code>LogService</code> interface. An
```
tjwatson, line 25: I would prefer not to modify felix code for our coding standards.

laeubi: `<tt>` tag is deprecated by html5 instead either

- [kbd](https://wiki.selfhtml.org/wiki/HTML/Elemente/kbd)
- [var](https://wiki.selfhtml.org/wiki/HTML/Elemente/var)
- [code](https://wiki.selfhtml.org/wiki/HTML/Elemente/code)
- [samp](https://wiki.selfhtml.org/wiki/HTML/Elemente/samp)

must be used to this is not really a coding standard question and javadoc complains about it, I can remove it from this change as we do not generate javadoc for it but it probably should be addressed at felix then.

tjwatson (approved):

PR #484, opened 2024-01-19
get rid of deprecated Class.newInstance()

jukzi:


bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/bundles/ClassLoadingBundleTests.java
```diff
@@ -2149,7 +2158,8 @@ public void testDefaultLocalUninstall() throws Exception {
 
 	@Test
 	public void testBug490902() throws BundleException, InterruptedException, ClassNotFoundException,
-			InstantiationException, IllegalAccessException {
+			InstantiationException, IllegalAccessException, IllegalArgumentException, InvocationTargetException,
+			NoSuchMethodException, SecurityException {
```
HannesWell, line 2162: For test cases that throw that many different type of exception, I would just specify `throws Exception`.
If an exception is throw its always an error.

HannesWell (approved):
Looks good as well, thank you.
The declared exceptions could IMHO be reduced to just `Exception`, but that's not critical for this.

PR #485, opened 2024-01-22
remove Exceptions not actually thrown

jukzi:
eliminates problem markers and dead code


jukzi:
only unrelated TCK fails...

jukzi:
jenkins says fixes 19
image

laeubi (approved):
It looks sane but I would give final word @tjwatson

vogella:
> can any committer review/submit, please?

I can nominate you as committer, if you want. Clean-up work is usually simpler if you have write access.

jukzi:
> I can nominate you as committer, if you want. Clean-up work is usually simpler if you have write access.

ok, please do so. But i won't do much more then cleanups. Regarding bugs i want to focus on other projects.

PR #493, opened 2024-01-30
Small Resolver refactoring for PDE API tooling stability

iloveeclipse:
- changed Maps/Sets to "Linked" versions where iteration was done
- changed resolveBundles() to resolve fragments only if asked to - this avoids the problem that fragments are resolved before host bundles
- added some debug logging for unresolved fragment hosts

See https://github.com/eclipse-pde/eclipse.pde/issues/1073

Note: this change is technically **not required** to fix https://github.com/eclipse-pde/eclipse.pde/issues/1073, it only makes random things to be less random and should simplify our life for debugging other bugs in PDE.

### DIFF
--- a/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/module/ResolverImpl.java
+++ b/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/module/ResolverImpl.java
@@ -26,6 +26,8 @@ import java.util.Dictionary;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


tjwatson (commented):
I'm not sure I understand why system bundle fragments should not be resolved when resolving the system bundle.  Is it that other fragments were somehow getting set to resolved also in the first call to `resolveBundles0`?

iloveeclipse:
> I'm not sure I understand why system bundle fragments should not be resolved when resolving the system bundle. Is it that other fragments were somehow getting set to resolved also in the first call to `resolveBundles0`?

It is not system bundle fragments that were resolving in the first call to resolve system bundle, but **all** fragments.

PR #494, opened 2024-02-01
pom.xml: remove code.ignoredWarnings

jukzi:
to get same warnings on jenkins as in workspace


akurtakov:
TCK failure is unrelated.

PR #498, opened 2024-02-04
Handle return value 24 the same was as 23.

merks:
https://github.com/eclipse-equinox/p2/issues/166

### DIFF
--- a/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
+++ b/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
@@ -1435,10 +1435,10 @@ public class Main {
 			// In case something weird happens, just dump stack - logging is not available at this point
 			t.printStackTrace();
 		} finally {
-			// If the return code is 23, that means that Equinox requested a restart.
-			// In order to distinguish the request for a restart, do a System.exit(23)
+			// If the return code is 23 or 24, that means that Equinox requested a restart.
+			// In order to distinguish the request for a restart, do a System.exit(result)
 			// no matter of 'osgi.noShutdown' runtime property value.
-			if (!Boolean.getBoolean(PROP_NOSHUTDOWN) || result == 23)
+			if (!Boolean.getBoolean(PROP_NOSHUTDOWN) || result == 23 || result == 24)
 				// make sure we always terminate the VM
 				System.exit(result);
 		}


merks:
@umairsair

I'm not sure how to test if this is needed but it wold seem consistent with the exiting handling for 23.

bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1438,7 +1438,7 @@ public static void main(String[] args) {
 			// If the return code is 23, that means that Equinox requested a restart.
 			// In order to distinguish the request for a restart, do a System.exit(23)
 			// no matter of 'osgi.noShutdown' runtime property value.
-			if (!Boolean.getBoolean(PROP_NOSHUTDOWN) || result == 23)
+			if (!Boolean.getBoolean(PROP_NOSHUTDOWN) || result == 23 || result == 24)
```
iloveeclipse, line 1441: Ed, the comment has to be updated

merks: You think?  😱 😖

iloveeclipse: Well, comment says something about 23, but then we have 23 and 24, with no idea why 24 is needed

merks: I was only joking.   If you read the linked issue you will know why 24 is needed.  In any case, note how the comment is actually wrong and or simply misleading already to suggest that it does System.exit(23) when the code actually does System.exit(result).  I'll fix it.

laeubi:
I just wanted to link this one here regarding exit codes it sometimes comes to a surprise:
- https://github.com/eclipse-equinox/equinox/discussions/378

merks:
This code has not been touched in a very long time so I doubt there is any one around who can tell us what's right or wrong.  I think consistently handing 24 as we do 23 is most likely the correct and consistent thing to do and it's what I've done elsewhere to fix actual problems, e.g.,

https://github.com/eclipse-pde/eclipse.pde/pull/1121

HannesWell:
I wonder if there should be some global constants for that magic numbers?

laeubi:
> I wonder if there should be some global constants for that magic numbers?

You mean like https://github.com/eclipse-equinox/equinox/blob/850923a5273e11487e5a0b2b505bfdef637d7d2d/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/app/IApplication.java#L48 :-)

HannesWell:
> > I wonder if there should be some global constants for that magic numbers?
> 
> You mean like

For example, yes. :)
I think it would be good to use that where possible, i.e. platform and PDE. Here it is probably not accessible.

merks:
I cancel this because https://github.com/eclipse-platform/eclipse.platform.ui/pull/1622 avoids surfacing that exit code.

Issue #499, opened 2024-02-05
"Exception saving profile preferences"


jukzi:
i got this two log entries after another:

eclipse.buildId=4.31.0.I20240131-1800
java.version=21.0.2

```
org.eclipse.equinox.p2.touchpoint.natives
Warning
Mon Feb 05 08:55:58 CET 2024
Could not remove temporary backup directory (it is safe to manually delete it and its contents): C:\Users\jkubitz\AppData\Local\Temp\C__Users_jkubitz_platform-2024-01-10_eclipse_1be71e04-0019-45a0-9663-13b0971d78eb

```

```
Exception saving profile preferences

org.osgi.service.prefs.BackingStoreException: Exception saving preferences to: C:/Users/jkubitz/platform-2024-01-10/eclipse/p2/org.eclipse.equinox.p2.engine/profileRegistry/C__Users_jkubitz_platform-2024-01-10_eclipse.profile/.data/.settings/org.eclipse.equinox.p2.artifact.repository.prefs.
	at org.eclipse.core.internal.preferences.EclipsePreferences.write(EclipsePreferences.java:304)
	at org.eclipse.core.internal.preferences.EclipsePreferences.save(EclipsePreferences.java:998)
	at org.eclipse.equinox.internal.p2.engine.ProfilePreferences.doSave(ProfilePreferences.java:141)
	at org.eclipse.equinox.internal.p2.engine.ProfilePreferences$SaveJob.run(ProfilePreferences.java:54)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
Caused by: java.nio.file.NoSuchFileException: C:\Users\jkubitz\platform-2024-01-10\eclipse\p2\org.eclipse.equinox.p2.engine\profileRegistry\C__Users_jkubitz_platform-2024-01-10_eclipse.profile\.data\.settings\org.eclipse.equinox.p2.artifact.repository.prefs
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:85)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileCopy.move(WindowsFileCopy.java:403)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.move(WindowsFileSystemProvider.java:291)
[...]

PR #500, opened 2024-02-05
fix "Exception saving profile preferences" #499

jukzi:
synchronize static + retry Files.move due to JDK bug

https://github.com/eclipse-equinox/equinox/issues/499


jukzi:
> But @jukzi please add a reference to the mentioned bug so that later it can be decided if the workaround is still necessary.

done

iloveeclipse:
Let's merge that one.

PR #513, opened 2024-02-13
CacheWriter: fix Non-externalized string warning

jukzi:
Non-externalized string literal; it should be followed by //$NON-NLS-<n>$


jukzi:
image

PR #514, opened 2024-02-13
set ut8 encoding

jukzi:
to fix encoding warning
"Project '...' has no explicit encoding set"


merks:
I wonder why it thinks so many tests are removed?

jukzi:
> I wonder why it thinks so many tests are removed?

my best guess: it did not execute the tests after detecting missing version bump

PR #516, opened 2024-02-13
replace deprecated X509Certificate methods

jukzi:
The method getSubjectDN() from the type X509Certificate is deprecated The method getIssuerDN() from the type X509Certificate is deprecated


merks:
Thank for all the cleanups!

PR #517, opened 2024-02-13
CloseableURLClassLoader: replace deprecated getPackage()

jukzi:
The method getPackage(String) from the type ClassLoader is deprecated since version 9


tjwatson (approved):
LGTM

PR #533, opened 2024-02-16
Add support for Coordinator to Equinox Configuration Admin

laeubi:
Chapter 11 of the Configuration Admin Service Specification mandates the support of the Coordinator service but currently Equinox fails the TCK tests in that area.

This adds support for participation in the coordinator for the Equinox Configuration Admin.

### DIFF
--- a/bundles/org.eclipse.equinox.cm/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.cm/META-INF/MANIFEST.MF
@@ -8,8 +8,9 @@ Bundle-Version: 1.6.0.qualifier
 Bundle-Activator: org.eclipse.equinox.internal.cm.Activator
 Import-Package: org.osgi.framework;version="1.7.0",
  org.osgi.service.cm;version="[1.6,1.7)",
+ org.osgi.service.coordinator;version="[1.0.0,2.0.0]",
+ org.osgi.service.event;version="1.0";resolution:=optional,
  org.osgi.service.log;version="1.3.0",
- org.osgi.service.event;version="1.0"; resolution:=optional,
  org.osgi.util.tracker;version="1.3.1"
 Bundle-RequiredExecutionEnvironment: JavaSE-17
 Provide-Capability: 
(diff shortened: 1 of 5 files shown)


laeubi:
I could get it now down to one failing test in the TCK but this is tricky if not fully specified, sadly [the chapter in the Spec is rather short](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.cm.html#service.cm-coordinatorsupport):

> The [Coordinator Service Specification](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.coordinator.html) defines a mechanism for multiple parties to collaborate on a common task without a priori knowledge of who will collaborate in that task. The Configuration Admin service must participate in such scenarios to coordinate with provisioning or configuration tasks.
> 
> If configurations are created, updated or deleted and an implicit coordination exists, the Configuration Admin service must delay notifications until the coordination terminates. However the configuration changes must be persisted immediately. Updating a Managed Service or Managed Service Factory and informing asynchronous listeners is delayed until the coordination terminates, regardless of whether the coordination fails or terminates regularly. Registered synchronous listeners will be informed immediately when the change happens regardless of a coordination.

The failing test works that way:

1. registers a `ManagedService` and receives an update with `properties = null`
2. The starts a coordination update it and delete it afterwards
3. the is asserts that update was only called one

The assert fails for equinox because two calls to update with `properties = null` are recorded, but from the spec text it is unclear that this is wrong as it only says the update must be delayed, but not that it must detect that the service actually is already contain the state it is now updated to.

Especially the chapter [104.7.4 Updating a Configuration ](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.cm.html#service.cm-updating.configuration) mentions that a ConfigAdmin might blindly update a `ManagedService` if no special action is taken here, so I think this is maybe even covered by the spec.

iloveeclipse:
@laeubi : not sure you've missed that we are in RC1 phase, and all commits require a review from other committer before merge?

merks:
The email suggests the freeze starts the 21st

https://www.eclipse.org/lists/equinox-dev/msg09560.html

I haven’t seen PRs with freeze failures yet either. E.g., all the aggregator PRs did not indicate the is a freeze yet. 

Perhaps there is some confusion or miscommunication?

laeubi:
> @laeubi : not sure you've missed that we are in RC1 phase, and all commits require a review from other committer before merge?

Thtas possible its still a bit confusing sometimes but as the freeze check does not complains I thought that it is fine calendar says RC1 stabilization start February 20 6am

iloveeclipse:
I think it is obvious, that all changes merged after M3 are going for RC1?
So independently when the RC1 is planned all changes **after M3** require extra approval, review etc.

PR #536, opened 2024-02-21
[memory] intern ConfigurationElement.propertiesAndValue

jukzi:
o.e.core.internal.registry.ConfigurationElement.propertiesAndValue is likely to contain duplicate Strings like "true"

### DIFF
--- a/bundles/org.eclipse.equinox.registry/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.registry/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.registry;singleton:=true
-Bundle-Version: 3.12.0.qualifier
+Bundle-Version: 3.12.100.qualifier
 Bundle-Localization: plugin
 Export-Package: org.eclipse.core.internal.adapter;x-internal:=true,
  org.eclipse.core.internal.registry;x-friends:="org.eclipse.core.runtime",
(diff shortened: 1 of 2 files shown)


jukzi:
I would like to get rid of this 4MB overhead:
image

bundles/org.eclipse.equinox.registry/src/org/eclipse/core/internal/registry/TableReader.java
```diff
@@ -299,7 +299,7 @@ private String[] readPropertiesAndValue(DataInputStream inputStream) throws IOEx
 			return RegistryObjectManager.EMPTY_STRING_ARRAY;
 		String[] properties = new String[numberOfProperties];
 		for (int i = 0; i < numberOfProperties; i++) {
-			properties[i] = readStringOrNull(inputStream);
+			properties[i] = readStringOrNull(inputStream).intern();
```
laeubi, line 302: The method name suggest it could be null, maybe better intern in the method itself?

jukzi: thanks for the hint - in which method?

laeubi: > thanks for the hint - in which method?

`readStringOrNull`

merks: I'm not sure if there are two such methods:

image

laeubi: > I'm not sure if there are two such methods:
> 
> image

Well the one that this TabelReader calls then...

PR #537, opened 2024-02-22
Fix ModuleCapability/ModuleRequirement hashcode/equals according to API

laeubi:
The API of Capability/Requirement currently defines the contract to be: This Capability/Requirement is equal to another Capability/Requirement if they have the same namespace, directives and attributes and are declared by the same resource.

But Equinox currently implements hashCode/equals in terms of object identity.

A similar problem might be in `ModuleRevision` where it says two resources are equal if:

> This Resource is equal to another Resource if both have the same content and come from the same location. Location may be defined as the bundle location if the resource is an installed bundle or the repository location if the resource is in a repository.

I find this a bit unclear what "content" means, in the context of a repository it might be the hashsum but for a Bundle(Revision) I think this is hard to answer.

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleCapability.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleCapability.java
@@ -16,8 +16,10 @@ package org.eclipse.osgi.container;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import org.osgi.framework.namespace.NativeNamespace;
 import org.osgi.framework.wiring.BundleCapability;
+import org.osgi.resource.Capability;
 
 /**
  * An implementation of {@link BundleCapability}.
@@ -30,6 +32,7 @@ public final class ModuleCapability implements BundleCapability {
 	private final Map<String, Object> attributes;
 	private final Map<String, Object> transientAttrs;
 	private final ModuleRevision revision;
+	private transient int hashCode;
 
 	ModuleCapability(String namespace, Map<String, String> directives, Map<String, Object> attributes,
 			ModuleRevision revision) {
@@ -92,5 +95,27 @@ public final class ModuleCapability implements BundleCapability {
 	@Override
 	public String toString() {
 		return namespace + ModuleContainer.toString(attributes, false) + ModuleContainer.toString(directives, true);
+	}
+
+	@Override
+	public int hashCode() {
+		if (hashCode != 0) {
+			return hashCode;
+		}
+		return hashCode = Objects.hash(namespace, directives, attributes, revision);
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj) {
+			return true;
+		}
+		if (obj instanceof Capability) {
+			Capability other = (Capability) obj;
+			return Objects.equals(namespace, other.getNamespace()) && Objects.equals(directives, other.getDirectives())
+					&& Objects.equals(attributes, other.getAttributes())
+					&& Objects.equals(revision, other.getResource());
+		}
+		return false;
 	}
 }
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleRequirement.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleRequirement.java
@@ -15,6 +15,7 @@ package org.eclipse.osgi.container;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import org.eclipse.osgi.internal.container.Capabilities;
 import org.eclipse.osgi.internal.framework.FilterImpl;
 import org.osgi.framework.InvalidSyntaxException;
@@ -24,6 +25,7 @@ import org.osgi.framework.namespace.PackageNamespace;
 import org.osgi.framework.wiring.BundleCapability;
 import org.osgi.framework.wiring.BundleRequirement;
 import org.osgi.resource.Namespace;
+import org.osgi.resource.Requirement;
 
 /**
  * An implementation of {@link BundleRequirement}. This requirement implements
@@ -38,6 +40,7 @@ public class ModuleRequirement implements BundleRequirement {
 	private final Map<String, String> directives;
 	private final Map<String, Object> attributes;
 	private final ModuleRevision revision;
+	private transient int hashCode;
 
 	ModuleRequirement(String namespace, Map<String, String> directives, Map<String, ?> attributes,
 			ModuleRevision revision) {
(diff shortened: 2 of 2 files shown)


laeubi:
@tjwatson how can we proceed here? I wonder what would be the advantage of `Object.equals` over the specified behavior, e.g. if Equinox require it on some places then still Object Identity might better be used explicitly (and documented) on these places?

tjwatson:
> @tjwatson how can we proceed here?

I don't think we should proceed here.  In the framework each `BundleRevision` is represented by a distinct installation generation of a `Bundle`.  There are cases where the same `Bundle` may be installed multiple times if using https://github.com/osgi/osgi/blob/11d440e336ac2c22bade54105ed4ef2c9aab79d8/org.osgi.framework/src/org/osgi/framework/hooks/resolver/ResolverHook.java#L178 to filter out singleton collisions.  It is also possible when doing bundle updates which leave behind "pending" removals for a particular `BundleRevision`, but then the same bundle is installed in a different location.

Equinox regions does this to create isolated regions.  So for the case where the "same" bundle JAR (Resource) is installed multiple times (into different "regions"), the framework HAS to treat each `BundleRevision` instance as identity unique.  Since each instance of `BundleRevision` is identity unique then that implies each Capability and Requirement from a revision is also identity unique.  This is by design and should not be changed.

laeubi:
But "same bundle" then means **same** bsn and same version is this really something to concern? Otherwise Req/Cap can never be the same (as bundle version would be different). Also as mentioned the Framework internally can still treat them differently than the default equals/hashcode...

tjwatson:
> But "same bundle" then means **same** bsn and same version is this really something to concern?

Not sure what you mean by concern.  Yes same exact JAR is installed multiple times at a different bundle location.  For this the framework has to treat each `BundleRevision` as different even if they are backed by the same exact bundle JAR.

> Otherwise Req/Cap can never be the same (as bundle version would be different).

Yes, this is what I am asserting.  Each `BundleRevision` is identity unique so each cap/req is identity unique by design.

> Also as mentioned the Framework internally can still treat them differently than the default equals/hashcode...

I see no advantage to that complication when each `BundleRevision` must be identity unique, just use the default equals/hashcode to get you that behavior.

laeubi:
But if the underlying BundleRevision == Resource is different then the Req/Cap wont be equal as well so maybe we are talking about different things.

So I think we are safe here, I just wanted to reiterate on what the spec says:

> This Resource is equal to another Resource if both have **the same content and come from the same location**.

So if bundle location is different two resources are not equal.

Now for what is done here (Req/Cap) it says

> Capability/Requirement is equal to another Capability/Requirement if they have the same namespace, directives and attributes and are declared by the same resource.

So that should be safe as well from what you described, as each Rev is a different Resource.

tjwatson:
> So that should be safe as well from what you described, as each Rev is a different Resource.

Okay I think we are on the same page then?  We agree that the BundleRevision instances managed by the framework must be use identity equals/hashcode which implies that req/cap also can use identity equals/hash.  To do extra calculations based on attributes/directives would be a waste because the backing resource (BundleRevision) is always identity unique.

laeubi:
> which implies that req/cap also can use identity equals/hash. 

The spec says something different and I don't see why Req/Cap *must* use identity equals/hash here.

>  To do extra calculations based on attributes/directives would be a waste because the backing resource (BundleRevision) is always identity unique.

A common short-cut for this is to do identity comparison first, so I don't see we loose much here but just assume I do wrap a Requirement (e.g. bnd does that because Req/Cap and Resource are all immutable objects) and then compare it to the original Requirement. For clarity we now name them `BndRequirement` and `EquinoxRequirement`, what now happens is

```
BndRequirement.equals(EquinoxRequirement) = true (because is follows OSGi Spec)
EquinoxRequirement.equals(BndRequirement) = false (because it **NOT** follwo OSGi Spec and use identity)
```

even though (from the perspective of the spec) must return equal in both cases.

tjwatson:
> A common short-cut for this is to do identity comparison first

I don't see how that works.  Identity comparison is false and you fall back to the other non-identity comparison just makes it always use non-identity comparison which will have it return true when it should be false.

tjwatson:
> Why should it return true when false is expected?

I've lost track on what you are proposing or what problem you are trying to fix.  Perhaps an update to the PR would help?

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleCapability.java
```diff
@@ -93,4 +96,26 @@ public ModuleRevision getResource() {
...
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj) {
+			return true;
+		}
+		if (obj instanceof Capability) {
+			Capability other = (Capability) obj;
+			return Objects.equals(namespace, other.getNamespace()) && Objects.equals(directives, other.getDirectives())
+					&& Objects.equals(attributes, other.getAttributes())
+					&& Objects.equals(revision, other.getResource());
```
tjwatson, line 117: because the `revision` here is always a `BundleRevision` the `Objects.equals` call is always going to return false here unless the `obj` is somehow another `Capability` from the same `BundleRevision`.

tjwatson: But in that case it would have to be a case where a `BundleRevision has two capabilities with the exact same attributes and directives.

laeubi: >  unless the obj is somehow another Capability from the same BundleRevision.

This is only true when you assume that `obj` can only ever come from the running equinox system. But assume for example the class `aQute.bnd.osgi.resource.CapabilityImpl` is used and I construct it from a ModuleCapability:

```
ModuleCapability moduleCapability = ....;
aQute.bnd.osgi.resource.CapabilityImpl wrapper = new CapabilityImpl(moduleCapability.getNamespace(),
		moduleCapability.getResource(), moduleCapability.getDirectives(), moduleCapability.getAttributes());
System.out.println(moduleCapability.equals(wrapper)); // --> must be true
System.out.println(wrapper.equals(moduleCapability)); // --> must be true
```

then I have two `Capability` objects that should per specification compare to true when calling equals and produce the same hashcode. I hope this makes it more clear what this PR is about

Thinking more about it for performance reason one might compare the resource first, but that's more an implementation decision.

tjwatson: I see, but seems strange that they wrap the capability but not the resource.

I definitely would want the resource equality checked before checking the attrs/directives.

laeubi: > I definitely would want the resource equality checked before checking the attrs/directives.

Done!

PR #469, opened 2024-01-09
Additional testcase that show some unexpected behavior of resolver

laeubi:
# this is not to meant to be merged in its current form

The added testcase demonstrates some unexpected behavior of resolver we have already seen in the past eclipse simrel releases and I see regular with custom installations, also m2e suffers from long running resolve operations when starting test from the IDE.

To summarize the both things that hurt the most here are boil down to the following facts:

## Undesired substitutions

-  A bundle `V1.0` has a substitution package `A` and imports it with consumer range (e.g. `[1,2)`) and export it with a specific version `1.0`
- Now there is also a bundle `V1.5` that exports the package `A` with version `1.5` exporting the package with version `1.5`
- Now there is a consumer `C` importing `A` with range `[1,2)`
- Now there is a provider `P` importing  `A` with range `[1,1.5)`

Under some circumstance now it happens that the package of bundle `V1.0` get substituted with these from `V1.5`  resulting in `P` not being resolved because the import package can not be resolved anymore.

**What is expected**: the resolver does not substitute the package for `V1.0` because that will make it impossible for `P` to resolve.

FYI @tjwatson in the bundle test set I can provide to you this is the `org.eclipse.lsp4j.jsonrpc` bundle that gets packages substituted

## Undesired binding to a higher package version causing use-constraint violations

- There is a bundle `A` importing a package with a consumer range (e.g. `[1,2)`)
- There is a bundle `B` importing a package with a narrow range (e.g. `[1,1.5)`)
- The package is available in two versions (e.g. 1.0 + 1.8)
- Bundle `B` requires `A`.

Under some circumstances it now happens that `A` is bound to a higher version (`1.8`) and `B` is failed to resolve because there is a use constraint violation because `B` is exposed to the same package in different versions, the bundle `A` is actually not used by any other bundle at all.

**What is expected**:  the resolver does not bind `A` to a higher version if a consumer of `A` restrict it to a lower version already as it makes any consumer of `A` exposed to use-constraint violations.

@tjwatson in the bundle test set I can provide to you this is the `org.eclipse.sprotty.server` bundle that gets packages `com.google.gson;version="[2.8,3)"` bound to a higher gson version than those bundles that require `org.eclipse.sprotty.server` 

## What we have done in the past

[...]

Issue #448, opened 2023-12-18
Add an option to instruct equinox to clean the resolver state


laeubi:
Currently if one performs an install of new software it can happen that:

1. it takes very very long to start but after that it works
2. it takes very very long and one gets very obscure resolve errors
3. sometimes even eclipse crash completely (e.g. a bundle with native code is updated)

many of the cases (except there are really incompatible installs) can be solved by starting the framework with `-clean` option but this does not only clean the resolver state but nay persisted state and is a manual action to perform. Instead P2 should be able to instruct the framework to remove clear the current resolved state and start from fresh.

The least intrusive way seem to have a marker file that is stored at `configuration/org.eclipse.osgi` e.g. `.refreshstate`, whenever equinox find that file it should delete it and throw away its previously cached state.

@tjwatson do you think this is suitable, what would be the best place to start with such implementation?

HannesWell:
This reminds me of https://github.com/eclipse-pde/eclipse.pde/pull/775#issuecomment-1740866323.

tjwatson:
> This reminds me of [eclipse-pde/eclipse.pde#775 (comment)](https://github.com/eclipse-pde/eclipse.pde/pull/775#issuecomment-1740866323).

Right, with commit 907ec00a829e2fe03f156089775c70ddac834d6e I see no need for this mechanism.  P2 can already force a re-resolve of all bundles and should with that commit.  Any issues not fixed by that approach will not be fixed by this new flag to tell the framework itself to re-resolve all installed bundles on start.

laeubi:
> Any issues not fixed by that approach will not be fixed by this new flag to tell the framework itself to re-resolve all installed bundles on start.

I'm not sure this assertion is true. e.g. this only happens when simple configurator has started and then refresh everything but the "bad" things happen before. There is still a big difference between an empty `configuration/org.eclipse.osgi` directory (`-clean`) and starting with an initialized one, so I really want this flag to behave as if `configuration/org.eclipse.osgi` is an empty directory right from the start, so it should not _re-resolve_ but start from fresh as if there is no resolved state at all but for example retain bundle data areas and bundle ids.

tjwatson:
> so I really want this flag to behave as if `configuration/org.eclipse.osgi` is an empty directory right from the start, so it should not _re-resolve_ but start from fresh as if there is no resolved state at all but for example retain bundle data areas and bundle ids

[...]

PR #538, opened 2024-02-22
Use namespace, directives and attributes to compare equal of Requirement

laeubi:
Currently the resolver is ultra slow in some scenarios and even never finds a valid solution, just adding more and more permutations to check.

The reason is that one check that should actually *reduce* the number of problematic permutations fails because it uses Requirement#equals what is incapable of compare requirements across Resource boundaries, that means even if the requirement describes the same restrictions it is not considered equal and therefore selected as a possible permutation to remove this, then the validation check detects this is unresolvable adding possible new permutations and so on.

This now performs an equals check instead on namespace, directives and attributes (what can be disabled by a system property) what makes the resolver in such case from running forever to just an blink of an eye.

@tjwatson this _dramatically_ reduces resolving time and leads to a resolved state where previously confusing use constraint violation and missing packages are reported that are actually resolvable. For my example described here:

- https://github.com/eclipse-equinox/equinox/pull/469
- https://github.com/eclipse-equinox/equinox/issues/448

it goes from minutes (until it hits the timeout) to milliseconds, I therefore suggest to include this in RC1, I added a system property as last resort if this would break something badly, so please review and give +1 as PL, for the record I would also give +1 as PL but would want your vote and review as binding here!

FYI @tivervac

### DIFF
--- a/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
+++ b/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
@@ -18,23 +18,49 @@
  */
 package org.apache.felix.resolver;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.apache.felix.resolver.ResolverImpl.PermutationType;
 import org.apache.felix.resolver.ResolverImpl.ResolveSession;
 import org.apache.felix.resolver.reason.ReasonException;
-import org.apache.felix.resolver.util.*;
+import org.apache.felix.resolver.util.CandidateSelector;
+import org.apache.felix.resolver.util.CopyOnWriteSet;
+import org.apache.felix.resolver.util.OpenHashMap;
+import org.apache.felix.resolver.util.OpenHashMapList;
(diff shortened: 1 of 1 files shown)


laeubi:
In my example there is the following requirements from two bundles that are using substitution packages:

```
osgi.wiring.package; filter:="(osgi.wiring.package=org.eclipse.lsp4j.jsonrpc)"
osgi.wiring.package; filter:="(osgi.wiring.package=org.eclipse.lsp4j.jsonrpc)"
```

their toString looks equals but if you look into the object you will see that one requirement is from

`osgi.identity; osgi.identity="org.eclipse.lsp4j.jsonrpc"; type="osgi.bundle"; version:Version="0.19.0.v20221118-0359"`

and the other from

`osgi.identity; osgi.identity="org.eclipse.lsp4j.jsonrpc"; type="osgi.bundle"; version:Version="0.21.1.v20230829-0012"`

so using equals compare them as **false** resulting in (the drop of lines is where the resolver gives up after one minute):

grafik

after apply they fix this results in the following:

grafik

stbischof:
@timothyjward fyi

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
```diff
@@ -1260,6 +1287,25 @@ public boolean canRemoveCandidate(Requirement req)
         return false;
     }
 
+	private boolean isMatchingRequirement(Requirement requirement, Requirement other) {
+		if (USE_REQUIREMENTS_EQUALS) {
```
stbischof, line 1291: you may want to document this fallback somewhere

tjwatson:
> so using equals compare them as **false** resulting in (the drop of lines is where the resolver gives up after one minute):

But they are two different requirements because they are from two different bundles.

laeubi:
Convert to draft as it does not seem to solve the issue in all cases...

laeubi:
@tjwatson I agree that this seems not the correct way to address the issue, I'll closing this for now and try to came up with an improved solution. This has most probably be detected in the inital collection of candidates so the resolver can't run into a conflict.

PR #540, opened 2024-02-26
Expose the permutation data through the felix resolver logger

laeubi:
Currently it is not really possible to write a test that asserts something about the complexity of a problem that the resolver has performed. This has the risk of undetected performance regressions but also means performance improvements can not be easily investigated. Also in general code (like for example console commands) it could be useful to gather some statistics.

This now adds two new methods to the logger, one that is performing a log each time when a permutation is added or processed to either log this by the caller or perform any statistics on them.

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleResolver.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleResolver.java
@@ -38,6 +38,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.felix.resolver.Logger;
+import org.apache.felix.resolver.PermutationType;
 import org.apache.felix.resolver.ResolutionError;
 import org.apache.felix.resolver.ResolverImpl;
 import org.eclipse.osgi.container.ModuleRequirement.DynamicModuleRequirement;
@@ -89,11 +90,13 @@ final class ModuleResolver {
 	private static final String OPTION_USES = OPTION_RESOLVER + "/uses"; //$NON-NLS-1$
 	private static final String OPTION_WIRING = OPTION_RESOLVER + "/wiring"; //$NON-NLS-1$
 	private static final String OPTION_REPORT = OPTION_RESOLVER + "/report"; //$NON-NLS-1$
+	private static final String OPTION_PERMUTATION = OPTION_RESOLVER + "/permutation"; //$NON-NLS-1$
 
 	boolean DEBUG_ROOTS = false;
 	boolean DEBUG_PROVIDERS = false;
 	boolean DEBUG_HOOKS = false;
 	boolean DEBUG_USES = false;
+	boolean DEBUG_PERMUTATIONS = false;
 	boolean DEBUG_WIRING = false;
 	boolean DEBUG_REPORT = false;
 
@@ -115,6 +118,7 @@ final class ModuleResolver {
 		DEBUG_USES = debugAll || options.getBooleanOption(OPTION_USES, false);
 		DEBUG_WIRING = debugAll || options.getBooleanOption(OPTION_WIRING, false);
 		DEBUG_REPORT = debugAll || options.getBooleanOption(OPTION_REPORT, false);
+		DEBUG_PERMUTATIONS = debugAll || options.getBooleanOption(OPTION_PERMUTATION, false);
 	}
 
 	static final Collection<String> NON_PAYLOAD_CAPABILITIES = Arrays.asList(IdentityNamespace.IDENTITY_NAMESPACE);
@@ -470,6 +474,11 @@ final class ModuleResolver {
 
 		class ResolveLogger extends Logger {
 			private Map<Resource, ResolutionException> errors = null;
+			public int totalPerm;
+			public int processedPerm;
+			public int usesPerm;
+			public int importPerm;
+			public int subPerm;
 
 			public ResolveLogger() {
 				super(DEBUG_USES ? Logger.LOG_DEBUG : 0);
@@ -508,6 +517,28 @@ final class ModuleResolver {
 			protected void doLog(int level, String msg, Throwable throwable) {
 				Debug.println("RESOLVER: " + msg + SEPARATOR //$NON-NLS-1$
 						+ (throwable != null ? (TAB + TAB + throwable.getMessage()) : "")); //$NON-NLS-1$
+			}
+
+			@Override
+			public void logPermutationAdded(PermutationType type) {
+				totalPerm++;
+				switch (type) {
+				case USES:
+					usesPerm++;
+					break;
+				case IMPORT:
+					importPerm++;
+					break;
+				case SUBSTITUTE:
+					subPerm++;
+				}
+			}
+
+			PermutationType[] permutationTypes = PermutationType.values();
+
+			@Override
+			public void logProcessPermutation(PermutationType type) {
+				processedPerm++;
 			}
 
 		}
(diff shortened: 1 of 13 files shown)


bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/report/resolution/ResolutionReport.java
```diff
@@ -193,4 +193,24 @@ public interface Listener {
 	 * @return a resolution report message.
 	 */
 	String getResolutionReportMessage(Resource resource);
+
```
tjwatson, line 196: Why do we need an inner interface for this.  `ResolutionReport` should be considered a `ProviderType` and we can add new methods to it.  If there are risks of multiple impls then the new methods can use default implementations that return -1 like the javadoc says.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
```diff
@@ -45,7 +77,7 @@ public class ResolverImpl implements Resolver
 
     private final Executor m_executor;
 
-    enum PermutationType {
+    public enum PermutationType {
```
tjwatson, line 80: It is odd to have the Logger takes enums of this inner type on the ResolverImpl.  Should the enum be moved to Logger?

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleResolver.java
```diff
@@ -470,6 +477,8 @@ class ResolveProcess extends ResolveContext implements Comparator<Capability>, E
 
 		class ResolveLogger extends Logger {
 			private Map<Resource, ResolutionException> errors = null;
+			public int totalPerm;
+			public int processedPerm;
```
tjwatson, line 481: These could be stored in `org.eclipse.osgi.container.ModuleResolutionReport.Builder` and incremented in there so that all the data is stored there for the report.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleResolver.java
```diff
@@ -962,7 +989,20 @@ ModuleResolutionReport resolve() {
 					if (DEBUG_WIRING) {
 						printWirings(result);
 					}
-					report = reportBuilder.build(result, re);
+					int total = logger.totalPerm;
```
tjwatson, line 992: I would rather have new methods on the report builder to increment the values and then the `build` simply sets the final values into the report directly.  I don't see the need for this extra interface here.

bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/report/resolution/ResolutionReport.java
laeubi: It should be but it is not marked as those and the default is sadly consumer type :-\

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
laeubi: I would rather extract it into an own class then.

bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/report/resolution/ResolutionReport.java
tjwatson: I would mark it it as provider type then.  Regardless, we can add default impl to interface types that are consumer type.  But it is wrong to offer this as a consumer type anyway.

laeubi: okay I can do this still then we probably need to either silence or enhance the API check (would be a good opportunity I think) to understand `@Provider` / `@Consumer` type.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
laeubi: Done.

bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/report/resolution/ResolutionReport.java
laeubi: done.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleResolver.java
laeubi: done.

laeubi:
@tjwatson the checks are now green (except the usual TCK one) are you fine with this change?

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
```diff
@@ -91,24 +85,40 @@ static class ResolveSession implements Runnable
         private final ConcurrentMap<String, List<String>> m_usesCache = new ConcurrentHashMap<String, List<String>>();
         private ResolutionError m_currentError;
         volatile private CancellationException m_isCancelled = null;
+        private Logger logger;
+        private ToIntFunction<PermutationType> remaining = type -> {
```
tjwatson, line 89: Both logger and remaining should be `final`

bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/report/resolution/ResolutionReport.java
```diff
@@ -193,4 +197,24 @@ public interface Listener {
...
+	 * @since 3.20
+	 */
+	int getProcessedPermutations();
+
+	/**
+	 * @return the number of permutations cause by a use-constraint violation -1 if
+	 *         unknown
+	 * @since 3.20
+	 */
+	int getUsesPermutations();
```
tjwatson, line 219: So we don't care about other permutation types?  Could we do without this method for now?

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Logger.java
```diff
@@ -131,4 +132,29 @@ public void logUsesConstraintViolation(Resource resource, ResolutionError error)
...
     }
+
+	/**
+	 * Called whenever a new permutation is added by the resolver.
+	 * 
+	 * @param type      the type of the permutation
+	 * @param remaining a function that can be used to query the now current number
+	 *                  of permutation types
+	 */
+	public void logPermutationAdded(PermutationType type, ToIntFunction<PermutationType> remaining)
```
tjwatson, line 143: Using a function here seems like overkill vs simply passing in the calculated integer.

bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/report/resolution/ResolutionReport.java
laeubi: For now I mostly care about uses permutations, other might become important later on but currently these are the ones I like to cut down as they often are a sign of suboptimal choices of the resolver, e.g. ideally the resolver will never choose a permutation that results in a use constraint violation and there ore this number should go down to zero for a testcase that shows such sub optimal behavior.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Logger.java
laeubi: My idea here was that for the noop implementation this means the numbers are not calculated, its a simple size() call so probably not that of an impact but something I considered here.
The second thing would be that the method needs to be enhanced once there is a different permutation type maybe in the future, while with this approach the method can stay the same.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/ResolverImpl.java
laeubi: Done

laeubi:
@tjwatson I rebased again an hopefully addressed all comments now.

laeubi:
I interpret the silence as agreement here, further improvements might be done in separate PRs then.

PR #541, opened 2024-02-27
Where is my bundle wiki page migrated to Equinox

vogella:


vogella:
Only docu, so the freeze does not apply and the configuration admin TCK fails always IIRC. @laeubi IIRC you did add this one? Is their work done for fixing the failure?

PR #542, opened 2024-02-27
Reduce usages of ShadowList

laeubi:
ShadowList currently extends CandidateSelector just to modify the underlying protected datastructure, instead it seems better to make CandidateSelector completely final and perform the actions on a copy of the original, this can lead to some extra objects created in the inital phase but has the benefit that afterwards CandidateSelectors are completely unmodifiable what makes them safe to implement hashcode/equals based on values instead of object identity.


laeubi:
@tjwatson from the codeflow it looks like a ShadowList can not be in the map (unless we assume prepare is called twice) at this point  in time, I added a throws now to see if maybe some test code trigger this.

laeubi:
Will be performed in smaller steps...

Issue #543, opened 2024-03-01
z/OS issues with behavior of canWrite


tjwatson:
In https://bugs.eclipse.org/bugs/show_bug.cgi?id=563697 we added back some niche behavior to better detect virtual directory characteristics on Windows with respect to being able to write.  In particular, the ability to write `.dll` files to the framework persistent storage location.

This approach causes issues on other OSes that have monitoring of unauthorized attempts to write content to the file system.  In particular, the z/OS has such a facility.  Users get concerned when they see the Eclipse process attempting to write some random `.dll` file to the file system in their monitoring logs. We should disable this extra code, at least for z/OS, but I can see where it should only be enabled for Windows also.  For now I would like to disable it for z/OS in particular.

tjwatson:
This was fixed with #544

PR #544, opened 2024-03-01
Issue #543 - avoid attempts to write .dll files when on z/OS for canWrite

tjwatson:


tjwatson:
This PR will wait to the next release.

PR #548, opened 2024-03-03
Add Felix Resolver Tests

laeubi:
Currently Equinox Embeeds the felix resolver source but does not execute the felix-resolver test.

This is an attempt to also execute the tests to make sure we do not introduce regressions covered by these.


laeubi:
@tjwatson what do you think? If you think it is useful to have some more tests I'll open a CQ for the imported felix code.

tjwatson:
> @tjwatson what do you think? If you think it is useful to have some more tests I'll open a CQ for the imported felix code.

Would it be better to do the dev work on the resolver in the felix project.  I find it easier to do the work there where I can maintain the proper source formatting.  Then once I get a PR there and merged I simply copy the relevant source to Equinox.

laeubi:
Main problem for me is that Felix has only a few test, equinox has some more and I don't want to break either felix nor equinox. So it seems better to develop the code at Equinox (that seems to have a larger test coverage) and then backport things to Felix, not sure what was the process in the past.

Another obstacle is that felix-devs are not necessarily equinox devs and vice versa ... e.g I'm not a committer there, and there is even not a PR validation at all...

laeubi:
@tjwatson as a first step I tried to "modernize" the felix resolver here and add minimal validation to it:
- https://github.com/apache/felix-dev/pull/301

laeubi:
Converted to draft as a CQ is possibly needed.

laeubi:
@tjwatson If you are fine with it I would create a CQ for this one?

laeubi:
@tjwatson any concerns here or should I proceed?

tjwatson:
> @tjwatson any concerns here or should I proceed?

You can proceed.  I'm not sure you need a CQ for this since we already had a CQ to pull in the org.apache.felix package from the resolver sub project of felix.

laeubi:
Then lets merge this as we also don't ship this code is seems safe to me.

PR #549, opened 2024-03-03
Extract retrieval of ShadowList into a method and invert conditions

laeubi:
Currently the code to work with the ShadowList is quite hard to understand as it references in a quite big block and uses a negated condition with if/else.

This extracts the retrieval of the ShadowList into a method to prevent accidentally using a wrong reference and inverts the condition so there is no negation involved.

This makes is slightly easier to understand whats going on.

### DIFF
--- a/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
+++ b/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
@@ -907,36 +907,24 @@ class Candidates
                             // matter if they come from the host or fragment,
                             // since we are completing replacing the declaring
                             // host and fragments with the wrapped host.
-                            CandidateSelector cands = m_candidateMap.get(r);
-                            ShadowList shadow;
-                            if (!(cands instanceof ShadowList))
+
+                            if (origCap.getResource().equals(hostResource.getDeclaredResource()))
                             {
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
```diff
@@ -907,36 +907,24 @@ public ResolutionError prepare()
...
-                            if (!origCap.getResource().equals(hostResource.getDeclaredResource()))
+                            if (origCap.getResource().equals(hostResource.getDeclaredResource()))
                             {
-                                shadow.insertHostedCapability(
-                                        m_session.getContext(),
-                                        (HostedCapability) c,
-                                        new SimpleHostedCapability(
-                                                hostResource.getDeclaredResource(),
-                                                origCap));
+								// If the original capability is from the host, then
```
tjwatson, line 913: Felix codebase does not use tab.

laeubi: Is this important to Equinox (so should we always have 1:1 code) or can it maybe addressed if/when back ported to felix? Can we somehow enable using space for the felix parts in JDT?)

tjwatson (approved):

PR #553, opened 2024-03-07
Pull down the replace method from ShadowList to Candidates

laeubi:
The replace method is currently used on one place and the reason we construct a new ShadowList, to remove references to this class the method can simply pulls down to Candidates as it only operates on the internal unmodifiable list.

This also adds a check that if the Capability to replace does not exits it is a noop to call this method instead of running into IndexOutOfBoundsException.

@tjwatson one small step to get rid of direct references to `ShadowList` as a preliminary step for further refactorings.

### DIFF
--- a/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
+++ b/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
@@ -18,17 +18,38 @@
  */
 package org.apache.felix.resolver;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.apache.felix.resolver.ResolverImpl.PermutationType;
 import org.apache.felix.resolver.ResolverImpl.ResolveSession;
 import org.apache.felix.resolver.reason.ReasonException;
-import org.apache.felix.resolver.util.*;
+import org.apache.felix.resolver.util.CandidateSelector;
+import org.apache.felix.resolver.util.CopyOnWriteSet;
+import org.apache.felix.resolver.util.OpenHashMap;
+import org.apache.felix.resolver.util.OpenHashMapList;
+import org.apache.felix.resolver.util.OpenHashMapSet;
+import org.apache.felix.resolver.util.ShadowList;
 import org.osgi.framework.Version;
-import org.osgi.framework.namespace.*;
-import org.osgi.resource.*;
+import org.osgi.framework.namespace.HostNamespace;
+import org.osgi.framework.namespace.IdentityNamespace;
+import org.osgi.framework.namespace.PackageNamespace;
+import org.osgi.resource.Capability;
+import org.osgi.resource.Requirement;
+import org.osgi.resource.Resource;
+import org.osgi.resource.Wire;
+import org.osgi.resource.Wiring;
 import org.osgi.service.resolver.HostedCapability;
 import org.osgi.service.resolver.ResolutionException;
 import org.osgi.service.resolver.ResolveContext;
@@ -868,7 +889,8 @@ class Candidates
                 // really be attached to the original host, not the wrapper.
                 if (!c.getNamespace().equals(HostNamespace.HOST_NAMESPACE))
                 {
-                    Capability origCap = ((HostedCapability) c).getDeclaredCapability();
+                    HostedCapability hostedCapability = (HostedCapability) c;
+                    Capability origCap = hostedCapability.getDeclaredCapability();
                     // Note that you might think we could remove the original cap
                     // from the dependent map, but you can't since it may come from
                     // a fragment that is attached to multiple hosts, so each host
@@ -912,7 +934,7 @@ class Candidates
                             {
                                 // If the original capability is from the host, then
                                 // we just need to replace it in the shadow list.
-                                getShadowList(r).replace(origCap, c);
+                                m_candidateMap.get(r).replaceHostedCapability(hostedCapability);
                             }
                             else
                             {
@@ -921,7 +943,7 @@ class Candidates
                                 // shadow copy of the list accordingly.
                                 getShadowList(r).insertHostedCapability(
                                         m_session.getContext(),
-                                        (HostedCapability) c,
+                                        hostedCapability,
                                         new SimpleHostedCapability(
                                                 hostResource.getDeclaredResource(),
                                                 origCap));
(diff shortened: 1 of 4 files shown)


laeubi:
All test are green so from my side it would be ready to merge!

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/Candidates.java
```diff
@@ -912,7 +912,7 @@ public ResolutionError prepare()
                             {
                                 // If the original capability is from the host, then
                                 // we just need to replace it in the shadow list.
-                                getShadowList(r).replace(origCap, c);
+                                m_candidateMap.get(r).replace(origCap, c);
```
tjwatson, line 915: This will be problematic if a fragment capability is attached to this host which is also a candidate for the given requirement `r`.  The reason is because it will become possible for the `ResolveContext` to get handed the modified `List<Capability>` with the `insertHostedCapability` which now will have some entries that may be wrapped and some that are not.  Yet the javadoc for `insertHostedCapability` says the list must be 
> The list returned from  {@link #findProviders(Requirement)}.

So basically we need to make sure the `ResolveContext` only gets a list of capabilities that it returned from `findProviders`, we cannot insert our internal wrappers and expose the `ResolveContext` to these internal types that it does not know about.

laeubi: So you mean the list must be the original list returned from that method correct? Is there any test or reasoning for this? If I look at the implementation, the list is defiantly modified by the ShaddowList and copied already but I'll double check that.

laeubi: Or must not the `List` be the same but only the items must only be the ones from `findProviders` correct?

tjwatson: > Or must not the `List` be the same but only the items must only be the ones from `findProviders` correct?

That is correct, the objects in the list itself must have come from `findProviders`.

tjwatson: If I recall, the order of the items should also be the same because they are supposed to be in preference order of what the resolver will try first to satisfy the requirement and that original order of the capabilities is supposed to be controlled by the order `findProviders` returned.

laeubi:
@tjwatson thanks for the explanation and hints, I have now adjusted the PR to reflect the original idea of the ShadowList method.

laeubi:
Okay looking at the  javadoc it even says

> The list returned from {@link #findProviders(Requirement)}

This is already violated as we make a _copy_ of the list (even copied twice) so it can never be *the* list, I also made the list unmodifiable here and it shows that Equinox at least even inserts items in the list directly, what seems valid but not mandatory, so if a resolve context really uses a cached list and only inserts items in this cache, or even relies on identity lookup this will break.

This also seems not really covered by the TCK, so I think before going further here we first need a TCK, then make sure Felix passes the TCK and then go on.

Taking this into account it seems the resolver would need to build the datastructure in a two step way, what likely will also make it easier to "finalize it":

1. Operate on the list returned from `findProviders(Requirement)` without wrapping and / or copy and do not replace / shadow things, but insert HostedCapabilities as needed.
2. Create a new (finalized) model where we make the copy as well as replace / shadow things because after this phase we should not insert things anymore.

There is just one tricky part `org.apache.felix.resolver.Candidates.processCandidates(LinkedList<Resource>, Requirement, List<Capability>)`  also calls `insertHostedCapability` but in this case relies on that the list is updated automatically, I'm not sure this is really covered by the spec!

laeubi:
See
- https://github.com/osgi/osgi/pull/690

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowedCapability.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public String getNamespace() {
+        return hosted.getNamespace();
+    }
+
+    @Override
+    public Map<String, String> getDirectives() {
+        return null;
```
tjwatson, line 52: why not return the original directives?

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowedCapability.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj instanceof Capability) {
+            Capability other = (Capability) obj;
+            return Objects.equals(getNamespace(), other.getNamespace())
+                    && Objects.equals(getDirectives(), other.getDirectives())
```
tjwatson, line 90: I don't see how this works if `getDirectives` returns `null` here.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowList.java
```diff
@@ -64,13 +65,18 @@ public void insertHostedCapability(ResolveContext context, HostedCapability wrap
...
+                                                                   // not modify ours
+        for (int i = 0; i < copy.length; i++) {
+            Capability capability = copy[i];
+            if (capability instanceof ShadowedCapability) {
+                // we unwrap the ShadowedCapability here as we always must pass only what we
+                // have got from the ResolveContext#findProviders.
+                copy[i] = ((ShadowedCapability) capability).getShadowed();
+            }
+        }
+        int insertIdx = context.insertHostedCapability(Arrays.asList(copy), toInsertCapability);
```
tjwatson, line 78: This seems odd and not in the spirit of the specification when multiple hosted capabilities are inserted into the list.  The `ResolveContext` will keep getting the same original list back and it will not have the past hosted capabilities there that the `ResolveContext` had inserted.  This means the `ResolveContext` will not have any control over the order of the inserted `HostedCapability` objects in relation to each other.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowedCapability.java
laeubi: > why not return the original directives?

Was missed here...

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowedCapability.java
laeubi: Fixed.

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowList.java
laeubi: I now fixed that, the main problem is that Felix resolver is violation the contract of insertHosteCapability, I [added a TCK that shows the problem](https://github.com/osgi/osgi/pull/690), but at least this should work as before.

laeubi:
I'll close this for now and maybe came up with a different aproach.

PR #554, opened 2024-03-09
Do not copy the list returned by findCandidates

laeubi:
The specification requires that we pass in the original list of candidates to the insertHostedCapability but currently always a copy is performed.

This replaces the copy by directly keeping the reference given as the resolver is woning the list anyways.

See
- https://github.com/osgi/osgi/pull/690

### DIFF
--- a/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowList.java
+++ b/bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowList.java
@@ -48,8 +48,8 @@ public class ShadowList extends CandidateSelector
     }
 
     private ShadowList(List<Capability> unmodifiable, List<Capability> original, AtomicBoolean isUnmodifiable) {
-        super(unmodifiable, isUnmodifiable);
-        m_original = new ArrayList<Capability>(original);
+        super(new ArrayList<Capability>(unmodifiable), isUnmodifiable);
+        m_original = original;
     }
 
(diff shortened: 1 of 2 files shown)


laeubi:
At least this seem not to break anything ...

laeubi:
@tjwatson what do you think?

bundles/org.eclipse.osgi/felix/src/org/apache/felix/resolver/util/ShadowList.java
```diff
@@ -48,8 +48,8 @@ private ShadowList(CandidateSelector shadow, List<Capability> original)
     }
 
     private ShadowList(List<Capability> unmodifiable, List<Capability> original, AtomicBoolean isUnmodifiable) {
-        super(unmodifiable, isUnmodifiable);
-        m_original = new ArrayList<Capability>(original);
+        super(new ArrayList<Capability>(unmodifiable), isUnmodifiable);
```
tjwatson, line 51: I don't understand the switch here to copy for the super call and not for the `m_original`

laeubi: This code a a bit cumbersome because it switches "unmodifiable" what actually *is* modified later on but the original is later passed to the ProvisonContext.

laeubi:
I'll close this for now as the specification will be changed in this aspect:
- https://github.com/osgi/osgi/pull/690#issuecomment-2023315446

PR #555, opened 2024-03-10
added jdk.net to the JavaSE-1.8.profile

axeluhl:
Java 8 has jdk.net as a package. However, it is missing from the Java 8 profile of the org.eclipse.osgi system bundle. This, in turn, causes other bundles that import the jdk.net package to fail resolve properly under Java 8. Shouldn't jdk.net therefore be declared in the profile?

See also https://docs.oracle.com/javase/8/docs/jre/api/net/socketoptions/spec/jdk/net/package-summary.html

### DIFF
--- a/bundles/org.eclipse.osgi/JavaSE-1.8.profile
+++ b/bundles/org.eclipse.osgi/JavaSE-1.8.profile
@@ -192,6 +192,7 @@ org.osgi.framework.system.packages = \
  javax.xml.ws.spi.http,\
  javax.xml.ws.wsaddressing,\
  javax.xml.xpath,\
+ jdk.net,\
  org.ietf.jgss,\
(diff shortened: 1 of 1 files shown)


mickaelistria:
Isn't `jdk.net` mostly an internal JVM package? ie there are most likely a lot of other types in it that aren't API and shouldn't be made available by default.
Note that as a workaround, you can create a system bundle fragment that exports java.net so it becomes visible to consumers; or you can set the `osgi.java.profile` system property to point to a custom profile.

laeubi:
`org.osgi.framework.system.packages.extra` is the standard way to expose any additional (maybe jvm specific) packages:
https://docs.osgi.org/specification/osgi.core/8.0.0/framework.lifecycle.html#framework.lifecycle.launchingproperties

PR #559, opened 2024-03-21
Add support for Windows on Arm64 (WoA).

chirontt:
## Add support for Windows on Arm64 (WoA)

A new launcher fragment for WoA is added: `org.eclipse.equinox.launcher.win32.win32.aarch64`

Various features are also updated to include the new fragment.

## To manually build the Equinox launcher binaries for WoA
On a WoA box, run the following command at the root directory of this repo:
```
mvn clean generate-resources -Dnative=win32.win32.aarch64
```
and the following launcher binaries for WoA are generated:
```
eclipse.exe
eclipse_11900.dll
eclipsec.exe
```
which are moved over to the `rt.equinox.binaries` repo which is supposedly at the same directory level as that of this repo, i.e. at `..\rt.equinox.binaries`, and to their corresponding sub-directories there as followed:
```
..\rt.equinox.binaries\org.eclipse.equinox.executable\bin\win32\win32\aarch64\eclipse.exe
..\rt.equinox.binaries\org.eclipse.equinox.executable\bin\win32\win32\aarch64\eclipsec.exe
..\rt.equinox.binaries\org.eclipse.equinox.launcher.win32.win32.aarch64\eclipse_11900.dll
```

### DIFF
--- a/features/org.eclipse.equinox.executable.feature/library/win32/build.bat
+++ b/features/org.eclipse.equinox.executable.feature/library/win32/build.bat
@@ -42,6 +42,31 @@ IF "%MSVC_EDITION%"=="" set "MSVC_EDITION=auto"
 @rem Specify VisualStudio Version: '2022', '2019' etc.
 IF "%MSVC_VERSION%"=="" set "MSVC_VERSION=auto"
 
+@REM Compose host architecture string for MSVC
+IF "%PROCESSOR_ARCHITECTURE%"=="AMD64" (
+  SET HOST_ARCH=x64
+  SET defaultOSArch=x86_64
+) ELSE IF "%PROCESSOR_ARCHITECTURE%"=="ARM64" (
+  SET HOST_ARCH=arm64
+  SET defaultOSArch=aarch64
+) ELSE (
+  CALL :ECHO "ERROR: Unknown host architecture: %PROCESSOR_ARCHITECTURE%."
+  EXIT /B 1
+)
+
+@REM %TARGET_ARCH% may be specified by the caller for cross-compiling.
+@REM If not, build for builder machine's architecture
+IF "%TARGET_ARCH%"=="" (
+  SET TARGET_ARCH=%HOST_ARCH%
+)
+
+@REM Compose build argument for MSVC
+IF "%TARGET_ARCH%"=="%HOST_ARCH%" (
+  SET BUILD_ARCH=%TARGET_ARCH%
+) ELSE (
+  SET BUILD_ARCH=%HOST_ARCH%_%TARGET_ARCH%
+)
+
 @rem Search for a usable Visual Studio
 @rem ---------------------------------
 IF "%MSVC_HOME%"=="" echo "'MSVC_HOME' was not provided, auto-searching for Visual Studio..."
@@ -58,14 +83,14 @@ IF EXIST "%MSVC_HOME%" (
 	echo "     Refer steps for SWT Windows native setup: https://www.eclipse.org/swt/swt_win_native.php"
 )
 IF EXIST "%JAVA_HOME%" (
-	echo "JAVA_HOME x64: %JAVA_HOME%"
+	echo "JAVA_HOME 64-bit: %JAVA_HOME%"
 ) ELSE (
-	echo "WARNING: x64 Java JDK not found. Please set JAVA_HOME to your JDK directory."
+	echo "WARNING: 64-bit Java JDK not found. Please set JAVA_HOME to your JDK directory."
 	echo "     Refer steps for SWT Windows native setup: https://www.eclipse.org/swt/swt_win_native.php"
 )
 set javaHome=%JAVA_HOME%
 set makefile=make_win64.mak
-call "%MSVC_HOME%\VC\Auxiliary\Build\vcvarsall.bat" x64
+call "%MSVC_HOME%\VC\Auxiliary\Build\vcvarsall.bat" %BUILD_ARCH%
 
 rem --------------------------
 rem Define default values for environment variables used in the makefiles.
@@ -74,7 +99,6 @@ set programOutput=eclipse.exe
 set programLibrary=eclipse.dll
 set defaultOS=win32
 set defaultWS=win32
-set defaultOSArch=x86_64
 
 rem --------------------------
 rem Parse the command line arguments and override the default values.
--- a/pom.xml
+++ b/pom.xml
@@ -219,6 +219,7 @@
 		    <module>bundles/org.eclipse.equinox.launcher.gtk.linux.aarch64</module>
 		    <module>bundles/org.eclipse.equinox.launcher.gtk.linux.ppc64le</module>
 		    <module>bundles/org.eclipse.equinox.launcher.gtk.linux.x86_64</module>
+		    <module>bundles/org.eclipse.equinox.launcher.win32.win32.aarch64</module>
 		    <module>bundles/org.eclipse.equinox.launcher.win32.win32.x86_64</module>
 		    <module>bundles/org.eclipse.equinox.launcher.tests</module>
 		
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,6 +48,18 @@ spec:
 
 /** Returns the download URL of the JDK against whoose C headers (in the 'include/' folder) and native libaries the natives are compiled.*/
 def getNativeJdkUrl(String os, String arch) { // To update the used JDK version update the URL template below
+	if('win32'.equals(os) && 'aarch64'.equals(arch)) {
+		// Temporary workaround until there are official Temurin GA releases for Windows on ARM that can be consumed through JustJ
+		dir("${WORKSPACE}/repackage-win32.aarch64-jdk") {
+			sh """
+				curl -L 'https://github.com/adoptium/temurin17-binaries/releases/download/jdk17u-2024-02-07-14-14-beta/OpenJDK17U-jdk_aarch64_windows_hotspot_2024-02-07-14-14.zip' > jdk.zip
+				unzip -q jdk.zip jdk-17.0.11+1/include/** jdk-17.0.11+1/lib/**
+				cd jdk-17.0.11+1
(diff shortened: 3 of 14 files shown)


laeubi:
@chirontt thanks for bringing this forward, in SWT we used (eclipse) linked folders to make the same sources available for different fragements, would something like this be suitable for the WoA fragemnt as well so we don't need an additional intermediate?

chirontt:
Yes, it may be doable. The benefits will be to remove the new `org.eclipse.equinox.security.win32` fragment (as not needed) and keep the existing fragment `org.eclipse.equinox.security.win32.x86_64` intact, while the new fragment for WoA `org.eclipse.equinox.security.win32.aarch64` will share the existing Java & C source code in the `x86_64` folder by making use of the `source.. =` property in the `build.properties` file to link to them. Is this what you mean, just to be sure?

laeubi:
Yes correctly you can take a look here how it is done:
https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/binaries/org.eclipse.swt.gtk.linux.x86_64/.project
https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/binaries/org.eclipse.swt.gtk.linux.x86_64/build.properties

chirontt:
@laeubi I've done what you suggested, but I encounter some small hiccup: how to include the `cpp` source files in the final `source` bundle (like what the `x86_64` counterpart is doing)? Here are my various attempts for the `build.properties` for this `org.eclipse.equinox.security.win32.aarch64` module (after removing all unrelated stuff for clarity): the 1st attempt which I consider the simplest and matching closely with the `x86_64` counterpart:
```
source.. = ../org.eclipse.equinox.security.win32.x86_64/src/
src.includes = ../org.eclipse.equinox.security.win32.x86_64/cpp/
```
but the resulting `source` bundle won't include the `cpp` files, as if the `src.includes` property just ignores the above setting. Then I try these:
```
source.. = ../org.eclipse.equinox.security.win32.x86_64/src/,\
           ../org.eclipse.equinox.security.win32.x86_64/cpp/
```
but then the resulting `binary` (jar) bundle would include those `cpp` source files too. The next try are these settings:
```
source.. = ../org.eclipse.equinox.security.win32.x86_64/src/,\
           ../org.eclipse.equinox.security.win32.x86_64/cpp/
bin.excludes = jnicrypt.*
```
and it almost achieves what I want, but the resulting `source` bundle won't have a `cpp` subfolder for those source files (they are left in the root of the bundle), unlike what the `source` bundle for `x86_64` is currently having.

Is there a better way? Otherwise I'd use the `bin.excludes` way for the PR then.

HannesWell:
@chirontt I also looked into this and in order to simplify the `org.eclipse.equinox.security.win32` I think it is worth investigating if we can just use JNA in order to perform the only native call. Just like it is done for Mac and Linux. If that works, we can then just use the same fragment for x86_64 and aarch64.

Would you be interested in looking into this in a separate PR? If not, I can do it.

chirontt:
@HannesWell  Yes, using JNA is a good idea, better than these clunky JNI/C files with re-compile for various platforms. But I have no experience with JNA, so please go ahead and see if JNA is applicable here. Once JNA is proven to work, there are other places in `eclipse.platform` repo to apply it as well (if possible):

https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/bundles/org.eclipse.core.filesystem/natives
https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/bundles/org.eclipse.core.resources/natives

HannesWell:
> Yes, using JNA is a good idea, better than these clunky JNI/C files with re-compile for various platforms. But I have no experience with JNA, so please go ahead and see if JNA is applicable here. 

Absolutely. It looks good, see https://github.com/eclipse-equinox/equinox/pull/564.

> Once JNA is proven to work, there are other places in `eclipse.platform` repo to apply it as well (if possible):
> 
> https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/bundles/org.eclipse.core.filesystem/natives
> https://github.com/eclipse-platform/eclipse.platform/tree/master/resources/bundles/org.eclipse.core.resources/natives

The filesystem-access implemented in `org.eclipse.core.filesystem` is very performance sensitive and I'm not sure if migrating it to JNA would be a performance regression for x86_64. As far as I know performance is not the No.1 priority for JNA. So that would have to be measured first.
But as written in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/577#issuecomment-2016421721, the default-impl should already work for WoA (IIRC it is about 10% which is not dramatic in general, but nevertheless not acceptable for some users).
Eventually we should try to migrate the native access to the new `Foreign Function and Memory Access API` finalized in Java-22, but it will probably take some time until we can use that (latest with the next LTS Java-25, which will be usable for Eclipse in about two years). So I'm not sure if it is worth the effort to have that intermediate step.

The situation for `org.eclipse.core.resources.win32` is similar. Without it you (probably) only loose refreshing out resources changed outside the workspace through native polling. That is probably not so performance sensitive, but a lot more work to migrate (equinox.security.win32 is even much simplified by using JNA) so it is also the question if it is worth the effort to migrate it to something else than FFM.

Therefore my suggestion is to first focus on the absolute requirements, namely SWT and the Equinox-Launcher.
SWT is on track, but for the Equinox-Launcher (i.e. this change), we also have to adjust the build-pipeline to build the executable.
Unlike SWT the pipeline is not yet migrated to a nice Jenkins pipeline stored in this git-Repo and still exists as manually configured Jenkins builds in https://ci.eclipse.org/releng/view/Launcher/.
I had the intention to migrate that too for a while but will take this as an opportunity to really do it. Once that work is done and with the work done for SWT to support WoA, it should be as simple as the adjustments in the SWT build-pipeline to land this.
Once the basics work and we have even have integration-tests for WoA we can still think about porting o.e.core.filesystem/resources.

HannesWell:
@chirontt please rebase this PR and resolve the conflicts. The changes to the equinox.security fragments should not be necessary anymore since #564 is submitted.

chirontt:
Branch updated with the latest from master, with PR description updated as well.

features/org.eclipse.equinox.core.sdk/feature.xml
```diff
@@ -74,13 +74,11 @@
    <plugin
          id="org.eclipse.equinox.security.win32"
          os="win32"
-         arch="x86_64"
```
HannesWell, line 77: Good catch. I forgot that in https://github.com/eclipse-equinox/equinox/pull/564.
Can you please apply these two changes in this file in a separate PR mentioning it was forgotten?

chirontt: I thought you intentionally left them out of #564 ! because it was quite logical to me that you left them out, for this PR to fix them for WoA :) But now I see it was a misunderstanding on my part, please create a new PR for them and I'll remove them from this PR.

HannesWell: Done via https://github.com/eclipse-equinox/equinox/pull/573.



> I thought you intentionally left them out of #564 ! because it was quite logical to me that you left them out, for this PR to fix them for WoA :)

That is a somewhat philosophical question since, at the moment, there is only the x86_64 arch supported on Windows so there is practically not a real difference. But since the plugin could/should run on all archs I think it is more suitable to remove it now.

HannesWell:
@chirontt I have now created https://github.com/eclipse-equinox/equinox/issues/575 in order to track the work on streamlining the Equinox launcher+executable build pipeline. Once that is done, it should be simple to complete this.

HannesWell:
> @chirontt I have now created #575 in order to track the work on streamlining the Equinox launcher+executable build pipeline. Once that is done, it should be simple to complete this.

Finally I have a draft PR for that up: https://github.com/eclipse-equinox/equinox/pull/603
It will need some adjustments in the Jenkins configuration done by the EF-infra structure team and some testing, but since it is relativly similar to the SWT natives build pipeline I assume that all of that is completed relatively soon.

In the meantime you could already rebase your PR and resolve the conflicts due to my earlier rework/clean-up of the native build.

chirontt:
> In the meantime you could already rebase your PR and resolve the conflicts due to my earlier rework/clean-up of the native build.

PR rebased to latest in master, and PR description is updated as well to reflect the `mvn` command now used for building the launcher binaries.

pom.xml
```diff
@@ -269,6 +270,13 @@
             <dependency-resolution>
               <optionalDependencies>${target-platform.optionalDependencies}</optionalDependencies>
             </dependency-resolution>
+            <environments combine.children="append">
+              <environment>
+                <os>win32</os>
+                <ws>win32</ws>
+                <arch>aarch64</arch>
+              </environment>
+            </environments>
```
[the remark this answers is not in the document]
HannesWell: Moved that temporary change to `org.eclipse.equinox.executable` to apply it to a more narrow scope.

features/org.eclipse.equinox.executable.feature/library/win32/build.bat
```diff
@@ -74,7 +99,7 @@ set programOutput=eclipse.exe
 set programLibrary=eclipse.dll
 set defaultOS=win32
 set defaultWS=win32
-set defaultOSArch=x86_64
+set OS=Windows
```
HannesWell, line 102: Setting the OS environment variable seems not to be necessary.
```suggestion
```

HannesWell (approved):
I just applied the suggested changes and fixed a small missing piece in the `org.eclipse.equinox.executable.feature\resources\build.xml` (although I'm not sure if that's even used anymore, see https://github.com/eclipse-equinox/equinox/pull/592).

With that this is ready for submission, once the Jenkins build has succeeded (all other builds cannot succeed without the initial build of the `win32.aarch64` launcher binaries).

Thank you for yet another great contribution.

pom.xml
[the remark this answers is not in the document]
HannesWell: As planned, this was now removed again via https://github.com/eclipse-equinox/equinox/pull/621.

PR #560, opened 2024-03-21
Update GH actions to reduce warnings

akurtakov:


akurtakov:
One generally unstable test. Merging.

PR #562, opened 2024-03-23
Remove deprecated core.launcher.Main and equinox.launcher.WebStartMain

HannesWell:
Both classes have been deprecated over five years ago via https://bugs.eclipse.org/bugs/show_bug.cgi?id=544262.

There have been requests to keep the classes four years ago in https://bugs.eclipse.org/bugs/show_bug.cgi?id=544262#c29, but as far as I can tell there have been no contribution of the requested tests?
@tjwatson should we proceed with the removal or ask again for feedback? If there are still real users out there, they can still maintain the code by themself. Since they are using WebStart in some way they can probably even test it better than Equinox can do, if there are really no tests.


tjwatson:
Lets proceed with deleting this.

tjwatson (approved):

HannesWell:
> Lets proceed with deleting this.

Great. But there is a surprising number of references to that launcher in the SDK that have to be migrated first:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1937
- https://github.com/eclipse-platform/eclipse.platform/pull/1266
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2221

The references in PDE look fine since they handle old target-platforms that probably didn't include the 'new' class.

tjwatson:
> Great. But there is a surprising number of references in code to that launcher in the SDK that have to be migrated first:

Thanks for finding these.  It is good to get them updated to the new class.

HannesWell:
> > Great. But there is a surprising number of references in code to that launcher in the SDK that have to be migrated first:
> 
> Thanks for finding these. It is good to get them updated to the new class.

Your welcome and yes, absolutely. I guess they were simply not discovered since the class was only referenced by its name.

Now that we had a new I-build with all the remaining references replaced (where suitable) and the CI builds passed I think we can finally submit this.

HannesWell:
Looks like this removal went fine and didn't blow up the world. Thei I-build test results look like before:
https://download.eclipse.org/eclipse/downloads/drops4/I20240327-1800/testResults.php

PR #565, opened 2024-03-24
Enforce encapsulation of equinox-launcher Constants and JNIBridge

HannesWell:
In the Eclipse SDK there are neither references to `org.eclipse.equinox.internal.launcher.Constants` nor to `org.eclipse.equinox.launcher.JNIBridge`.
@tjwatson or anybody else are you aware of external users that need those two classes public at the places where they are?

### DIFF
--- a/features/org.eclipse.equinox.executable.feature/library/org.eclipse.launcher.tests/src/org/eclipse/equinox/launcher/TestLauncherApp.java
+++ b/features/org.eclipse.equinox.executable.feature/library/org.eclipse.launcher.tests/src/org/eclipse/equinox/launcher/TestLauncherApp.java
@@ -11,7 +11,7 @@
  * Contributors:
  *     Umair Sair - initial API and implementation
  *******************************************************************************/
-package main;
+package org.eclipse.equinox.launcher;
 
 import java.io.BufferedReader;
 import java.io.DataOutputStream;
@@ -21,8 +21,6 @@ import java.net.Socket;
 import java.util.ArrayList;
 import java.util.List;
 
-import org.eclipse.equinox.launcher.JNIBridge;
-
 /**
  * Dummy test application used for eclipse launcher testing. JUnit tests
  * launches the eclipse launcher with startup application pointing to the jar file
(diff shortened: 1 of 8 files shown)


HannesWell:
Looks like at least the org.eclipse.launcher.tests introduced with https://github.com/eclipse-equinox/equinox/pull/401, needs some adjustments.
But this could probably be solved by making the test a fragment of the launcher and renaming the TestLauncherApp's package to `org.eclipse.equinox.launcher`.

features/org.eclipse.equinox.executable.feature/library/org.eclipse.launcher.tests/src/org/eclipse/equinox/launcher/TestLauncherApp.java
```diff
@@ -11,7 +11,7 @@
  * Contributors:
  *     Umair Sair - initial API and implementation
  *******************************************************************************/
-package main;
+package org.eclipse.equinox.launcher;
```
tjwatson, line 14: The JNIBridge import below should be removed if we move to this package.

tjwatson (commented):
I can't say if I am confident that JNIBridge is not needed outside of the launcher by anyone.  Regardless, I don't think it was ever intended to be API.  We can move forward here I think.

features/org.eclipse.equinox.executable.feature/library/org.eclipse.launcher.tests/src/org/eclipse/equinox/launcher/TestLauncherApp.java
HannesWell: Done.

HannesWell:
> I can't say if I am confident that JNIBridge is not needed outside of the launcher by anyone. Regardless, I don't think it was ever intended to be API. We can move forward here I think.

Ok, good. 
Since the build passes now, should we submit this or are there concerns by anybody?

tjwatson (approved):

tjwatson:
> Ok, good.
> Since the build passes now, should we submit this or are there concerns by anybody?

Not from me.

PR #566, opened 2024-03-25
Clean-up and simplify LauncherTests

HannesWell:


HannesWell:
I assume the test-failure is either due to the conversion to text-block (which uses the platform-specific line-separator instead of always `\n`) or maybe some charset issues due to the change in `writeEclipseIni()`. But I currently have don't have the nerve to investigate that.
Anyone interested, is welcome to pick this up, maybe also considering to convert this test-project into a 'regular' test fragment to get rid of the symbolic link to `JNIBridge` which also does not work well on windows.

PR #576, opened 2024-03-29
[Build] Use generated Maven toolchains.xml instead of own one

HannesWell:
The supplied own toolchains.xml provides JDKs that probably don't match the GitHub workflow environment.

@laeubi you created the `toolchains.xml` in https://github.com/eclipse-equinox/equinox/pull/1, can you tell why? And since there is only a Java-17 JDK is set up, I wonder if the toolchain is necessary at all.


laeubi:
The java action has not supported generating toolchain.xmls in the past.

HannesWell:
> The java action has not supported generating toolchain.xmls in the past.

Ah ok. But then we should leverage it. :)
Having JDK-1.8 and 17 configured should be sufficient since all projects are either at 1.8 or 17.

PR #577, opened 2024-03-30
Remove obsolete build scripts and make-files from equinox.executables

HannesWell:
- The script win32/build.sh probably aimed to be the bash/sh equivalent of the build.bat respectively an adaption of the gtk/build.sh. But it became out-dated.

- The architectures Windows 32-bit (x86 non 64) and Itanium 64 (IA64) are not supported anymore, so the corresponding make-files are unused.

The `win32/make_mingw.mak` could probably also be removed but is currently referenced in the CDT `.cproject`.
@akurtakov do you think it would be possible to migrate the CDT config to use the `make_win64.mak` which is also used in the CI? And could you help with that? I'm not familiar with CDT.

Furthermore I wonder if the `.cdtproject` is relevant besides the `.cproject` file?


akurtakov:
AFAIR and quick search confirm it(https://www.eclipse.org/lists/cdt-dev/msg33478.html) - .cdtproject files are just remnants from the past, so removing them should be fine. About help with windows - I have practically zero access and experience with it but I believe @sravanlakkimsetti should be able to help here.

HannesWell:
> AFAIR and quick search confirm it(https://www.eclipse.org/lists/cdt-dev/msg33478.html) - .cdtproject files are just remnants from the past, so removing them should be fine. 

Thank you for that hint, just removed the `.cdtproject` file.

About help with windows - I have practically zero access and experience with it but I believe @sravanlakkimsetti should be able to help here.

I have now adjusted the `.cproject` file to call the `make_win64.mak` instead of the `make_mingw.mak` file. But that file seems only work with Microsofts `nmake` and not with the GNU `make`. And if change the `buildCommand` to `nmake` and try to execute the target in CDT it says (at least on my computer) that nmake is not found and I can replicate that on the CLI calling
```
nmake -f make_win64.mak clean
```
But if I run `call "%MSVC_HOME%\VC\Auxiliary\Build\vcvarsall.bat" x64` in advance the `nmake` command is found. So I guess that `vcvarsall.bat` adds `nmake` to the PATH env variable of the current shell.
Since I see the call of `vcvarsall.bat` in the build-scripts of the equinox binaries for Windows and in SWT I wonder if it is common to call that on Windows with MSVC and to not have `nmake` on the global PATH?
And if Yes, is there a common way in CDT to call `vcvarsall.bat` in advance respectively to do the same setup?

sravanlakkimsetti (approved):

sravanlakkimsetti:
@niraj-modi This looks fine to me. Can you please take a look at windows files?

HannesWell:
> The ANT build.xml are not used in any CI build and are not functional out of the box. For local builds developers usually use the native build-scripts directly.

It has turned out that these `build.xml` files where actually referenced from the `launcher-binary-parent/pom.xml`. But this has been reworked in https://github.com/eclipse-equinox/equinox/pull/588 and those files have been removed there.

On the other hand I noticed that there are `.cdtproject` and `.cproject` files in the equinox.launcher project which does not contain C sources at all. So I assume that's a leftover from the time when the sources resided there, which can also be removed.

akurtakov:
What is the status here?

HannesWell:
> What is the status here?

I was awaiting the feedback of the other persons linked, but since there was non in the last two weeks I think there won't be any and this can be submitted.

PR #583, opened 2024-04-02
Convert equinox.launcher.tests into a test-fragment

HannesWell:
This fixes the following shortcomings associated to the launcher tests.

- When building the native launcher and executable, the entire test project was also copied to the specialized build agents, which unnecessarily increases their network-traffic.

- Test cannot be executed (as JUnit-Plugin test) in the IDE without further setup.

- Setting up tests and patching a local developer eclipse should not be done in a make-file. Its complexity should not be increased by such task, that also only work with simple installations but not distributed ones.

@umairsair since you intially wrote the tests, would you mind to have a look at this? I think I retained the purpose and setup of the tests, but it would be good to have a review.


umairsair:
The intention for keeping the test project along with native source was that we integrate tests in our jenkins job as well while building the binaries. Thats why it was simpler to hide it behind one command. But I think we want to just rely on tests ran by github workflow.

Also as a developer working with launcher code, I think its a bit more complex now.. previously I had to run just one command and new binaries were built and the tests were executed.. what would be the workflow now?

The `dev_build_install` was quite old logic. Though I have never used it but I think its useful if someone wants to test the new binaries with existing eclipse, so we are altogether dropping this support or giving some alternative?

I have skimmed through the PR and will check it in detail. But we need to fix macos tests failure..

bundles/org.eclipse.equinox.launcher.tests/src/org/eclipse/equinox/launcher/tests/LauncherTests.java
```diff
@@ -11,50 +11,163 @@
...
+		}
+		String os = System.getProperty("osgi.os");
+		String ws = System.getProperty("osgi.ws");
+		String arch = System.getProperty("osgi.arch");
+
+		// Assemble a Eclipse installation mock in the test-temporary directory
+
+		Path exePath = Path.of("org.eclipse.equinox.executable", "bin", ws, os, arch);
+		if (os.contains("mac")) {
+			exePath = exePath.resolve("Eclipse.app/Contents/MacOS/eclipse");
```
umairsair, line 104: can we move `"Eclipse.app/Contents/MacOS/eclipse"` as value of `ECLIPSE_EXE_NAME` in case of mac?

bundles/org.eclipse.equinox.launcher.tests/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,10 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Tests
+Bundle-SymbolicName: org.eclipse.equinox.launcher.tests
+Bundle-Version: 1.0.0.qualifier
+Fragment-Host: org.eclipse.equinox.launcher
```
umairsair, line 6: just wondering why we are making it fragment..

umairsair:
Thanks @HannesWell ! It would be good if we can add a readme in launcher code for developers..

Also I see following in test results comment..

```
org.eclipse.equinox.launcher.tests.LauncherTests ‑ Unknown test
```

overall LGTM!

bundles/org.eclipse.equinox.launcher.tests/META-INF/MANIFEST.MF
HannesWell: This allows us to access package private classes in the `org.eclipse.equinox.launcher` bundle in the this test fragment and therefore avoids the symbolic link.
Fragment and host share the same classloader, therefore if you put classes in a fragment they can also access package-private elements in the same package of the host.

bundles/org.eclipse.equinox.launcher.tests/src/org/eclipse/equinox/launcher/tests/LauncherTests.java
HannesWell: The problem is that for the test-setup it looks like it didn't work if the exe was moved to that sub-directory, but maybe I have to check that again.
I have no Mac and can therefore only test it in the CI.

HannesWell:
> Thanks @HannesWell ! It would be good if we can add a readme in launcher code for developers..

Yes, I added that with https://github.com/eclipse-equinox/equinox/pull/585.

> 
> Also I see following in test results comment..
> 
> ```
> org.eclipse.equinox.launcher.tests.LauncherTests ‑ Unknown test
> ```
> 
Where do you see that? I neither see it in Jenkins nor in the GH workflow.
The Windows workflow currently fails and altough I have a Windows build machine I cannot really reproduce it. Have to investigate that further.

> overall LGTM!

Thanks!

umairsair:
> Where do you see that?

Screenshot from 2024-04-08 16-19-56

HannesWell:
> Thanks @HannesWell ! It would be good if we can add a readme in launcher code for developers..
> 
> Also I see following in test results comment..
> 
> ```
> org.eclipse.equinox.launcher.tests.LauncherTests ‑ Unknown test
> ```
> 

Because that error happened also on the master branch I investigated again and it turned out that the codeql workflow is missing the `EQUINOX_BINARIES_LOC` environment variable which was set in this PR for the Jenkins pipeline and build workflow. Fixed that with:
- https://github.com/eclipse-equinox/equinox/pull/587

PR #585, opened 2024-04-05
[Build] Complete 'install' and 'clean' targets in native make files

HannesWell:
- Copy the launcher library too. To enable that introduce more specific environment variables for their different location.
- Complete the list of files that have to be deleted in a 'clean'
- Add a little documentation about how to build the native binaries.


HannesWell:
The verification build in https://github.com/eclipse-equinox/equinox/pull/583 shows that the updated install target is working fine on all platforms.

PR #586, opened 2024-04-08
[Build] Run full build and test workflow on all platforms

HannesWell:
This ensures the full equinox build and all tests run on all major OS along with the native binaries build. Furthermore it simplifies the GH workflows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,11 +11,20 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    name: Build ${{ matrix.config.name }}
+    runs-on: ${{ matrix.config.runner-os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config: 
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -49,66 +67,14 @@ jobs:
         -Dmaven.test.failure.ignore=true
         -Drt.equinox.binaries.loc=${{ github.workspace }}/rt.equinox.binaries
         clean verify
-    - name: Upload Test Results
-      uses: actions/upload-artifact@v4
-      with:
-        name: test-results
```
HannesWell, line 55: @laeubi can you tell if the changing this requires adapting the test-results workflow?
I searched for the name but didn't found a match that seemed to need adoption.

HannesWell: The test-result comments says everything looks good, so I assume it is fine.

HannesWell:
After adjusting the expectations in `ServletTest` to consider the different default line-separators the tests now pass on all platforms.

PR #592, opened 2024-04-10
Remove unused equinox.executable.feature resources

HannesWell:
The `build.xml` is currently shipped with the `o.e.equinox.executable.feature` which is odd and it at least seems not to be used for the build of Equinox.
I wonder if this is used by Tycho or probably more likely by the PDE Build Ant targets?
@iloveeclipse can you tell that with certainty? I have found some hits in PDE that indicate that, but I'm not sure and as far as I understand that content I'm in doubt it is useful for anything?

More or less the same question I want to raise for to the `o.e.equinox.executable.feature/resources/build.properties`. It is also shipped with the feature but is content is partly outdated (`linux.aarch64` is missing) and its content is inconsistent. Sometimes it uses `launcher` as name of the executable and sometimes `${launcherName}` (which I assume is supposed to be interpolated?

https://github.com/eclipse-equinox/equinox/blob/9fd42b025bf81534e51251ba59e7768171aac1b3/features/org.eclipse.equinox.executable.feature/resources/build.properties#L17-L36

@akurtakov or @laeubi maybe you know something about this?


merks:
This is needed to build products, at least with Tycho.  It, e.g., https://download.eclipse.org/eclipse/updates/4.32-I-builds/I20240410-1800/features/org.eclipse.equinox.executable_3.8.2500.v20240410-1706.jar, contains the launchers:

image

For Oomph we need to take special steps to ensure it's specified the target platform so that we can build the installer product:

image

When a product is built, you can see it in the target folder:

image

When I remove it from the target platform, the build fails like this:
```
[INFO] -------------------------[ eclipse-repository ]-------------------------
[INFO] Resolving dependencies of MavenProject: org.eclipse.oomph:org.eclipse.oomph.setup.installer.product:1.33.0-SNAPSHOT @ D:\Users\merks\oomph-1.33\git\oomph\products\org.eclipse.oomph.setup.installer.product\pom.xml
[INFO] {osgi.os=win32, osgi.ws=win32, org.eclipse.update.install.features=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: org.eclipse.oomph.setup.installer.product.with-jre 1.33.0.qualifier
[ERROR]   Missing requirement: org.eclipse.oomph.setup.installer.product.with-jre 1.33.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.equinox.executable.feature.group 0.0.0' but it could not be found
[ERROR] 
[ERROR] See https://wiki.eclipse.org/Tycho/Dependency_Resolution_Troubleshooting for help.
```
So products appears to have an implicit requirement on this feature.

HannesWell:
> So products appears to have an implicit requirement on this feature.

Yes, but this PR is not about removing the feature, it is only about to remove the build.xml (and build.properties) embedded in the `built` feature jar:

grafik

I have updated this PR to also remove the `build.properties` to show its 'full' state, given both are not required.
I would be surprised if Tycho relies on an ANT script delivered with this feature, but I'm not sure if the build.properties delivered with it is used somehow.
At the same time I would not be surprised if the PDE build ANT task rely on the shipped `build.xml` or `build.properties` in some way. For ordinary bundles and features the `build.properties` are not included in the built jar. But as Ed already elaborated, the equinox.executable feature is special so I will investigate Tycho and PDE build in detail unless somebody else can say with confident that any of those files is useful or not.

sravanlakkimsetti:
I remember build.xml being used to package native launchers for equinox download page. See https://download.eclipse.org/equinox/drops/I20240411-1800/index.php(section Native Launchers). 

If these zip files are created without build.xml we can get rid of build.xml.

akurtakov:
> I remember build.xml being used to package native launchers for equinox download page. See https://download.eclipse.org/equinox/drops/I20240411-1800/index.php(section Native Launchers).
> 
> If these zip files are created without build.xml we can get rid of build.xml.

I am surprised by the many N/A on that page 
image

This makes me wonder whether anyone is ever looking at this page as we never got any report about this at all.

laeubi:
I remember that the pde-build tests depend on these files, I already wondered about them a while back but gave up on this as keeping them is obviously not an issue but removing them will probably break some unknown consumers.

HannesWell:
Closed unintentionally.

Issue #575, opened 2024-03-29
Integrate Equinox launcher and executable natives build into the Equinox Jenkins pipeline


HannesWell:
At the moment the Equinox launcher and executable natives are build in a set of separate Jenkins Freestyle-job which are only triggered manually and push the built native binaries to https://github.com/eclipse-equinox/equinox.binaries:
- https://ci.eclipse.org/releng/view/Launcher/

This setup has the potential to [forget to trigger the natives rebuild](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1407) once a change was merged and makes it harder to understand and maintain the build-pipeline.
Therefore, even tough the Equinox natives are not built that often, I propose to migrate the mentioned set of Jenkins freestyle jobs into the Equinox Jenkins Pipeline and let that pipeline trigger the native builds on demand (i.e. if there have been changes), [similar to how it is done for SWT](https://github.com/eclipse-platform/eclipse.platform.swt/blob/9ccc8e724e1a47bee8b5a24ce39b6a82dceeacc8/Jenkinsfile).
This has the advantage that the exact build procedure is defined as code in this repository (always up-to-date), the natives can be build during PR validation, it reduces the number of Jobs in the RelEng JIPP and will also help for https://github.com/eclipse-equinox/equinox/pull/559.

The launcher/executable build is currently performed in the `RelEng` JIPP, which has specially prepared platform specific build machines available as Jenkins agents that are capable to perform these builds. 

The main question to answer is how we make these required special Jenkins agents available for the Equinox build pipeline?
When doing the same for SWT (in https://github.com/eclipse-platform/eclipse.platform.swt/pull/514) there where basically two options:
1. Move the [equinox Jenkins Job](https://ci.eclipse.org/equinox/job/equinox/) from the [Equinox JIPP](https://ci.eclipse.org/equinox/) to the [RelEng JIPP](https://ci.eclipse.org/releng/)
2. Add the special build machines as extra agents to the [Equinox JIPP](https://ci.eclipse.org/equinox/).

For SWT option 1 was chosen since it was easier for @fredg02 to just move the `eclipse.platform.swt` job from the `platform` to the `releng` JIPP. I assume it's the same in this case. 
Option 2 would be the cleaner solution from a build separation point of view, further more the RelEng JIPP is quite busy and contains the long running I- and Y-builds that regularly occupy a lot of resources, especially the special platform specific build machines, but sometimes even the basic ones.
[...]

PR #595, opened 2024-04-14
Add --launcher.noRestart option to launcher.

dhendriks:
This adds a `-norestart` option to the launcher. When the option is used, it suppresses the restart behavior of exit codes 23 and 24, such that an `IApplication` can return all exit codes if the restart behavior is not desired.

The diff looks more complicated than it needs to. In `eclipse.c`, the changes to the switch statement only involve an extra `if` around the body of two `case`s, with a comment at the end to explain the fall-through. The bodies of these two new `if` statements contain the original unmodified original code of the `case`s, indented one extra level.

See also the discussion at https://github.com/eclipse-equinox/equinox/discussions/378.

### DIFF
--- a/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/app/IApplication.java
+++ b/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/app/IApplication.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2018 IBM Corporation and others.
+ * Copyright (c) 2005, 2024 IBM Corporation and others.
  *
  *  This program and the accompanying materials
  *  are made available under the terms of the Eclipse Public License 2.0
@@ -7,7 +7,7 @@
  *  https://www.eclipse.org/legal/epl-2.0/
  *
  *  SPDX-License-Identifier: EPL-2.0
- * 
+ *
  *  Contributors:
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
@@ -23,7 +23,7 @@ package org.eclipse.equinox.app;
  * <p>
  * Clients may implement this interface.
  * </p>
- * 
+ *
  * @since 1.0
  */
 public interface IApplication {
@@ -34,7 +34,12 @@ public interface IApplication {
 	public static final Integer EXIT_OK = Integer.valueOf(0);
 
 	/**
-	 * Exit object requesting platform restart
+	 * Exit object requesting platform restart.
+	 *
+	 * <p>
+	 * Note: The handling of this special exit code may be disabled by launcher
+	 * argument {@code --launcher.noRestart}.
+	 * </p>
 	 */
 	public static final Integer EXIT_RESTART = Integer.valueOf(23);
 
@@ -44,6 +49,11 @@ public interface IApplication {
 	 * the executable is relaunched the command line will be retrieved from the
 	 * {@link IApplicationContext#EXIT_DATA_PROPERTY eclipse.exitdata} system
 	 * property.
+	 *
+	 * <p>
+	 * Note: The handling of this special exit code may be disabled by launcher
+	 * argument {@code --launcher.noRestart}.
+	 * </p>
 	 */
 	public static final Integer EXIT_RELAUNCH = Integer.valueOf(24);
 
@@ -64,7 +74,7 @@ public interface IApplication {
 	 * Note: This method is called by the platform; it is not intended to be called
 	 * directly by clients.
 	 * </p>
-	 * 
+	 *
 	 * @return the return value of the application
 	 * @see #EXIT_OK
 	 * @see #EXIT_RESTART
@@ -80,7 +90,7 @@ public interface IApplication {
 	 * running application is ready to exit. The {@link #start(IApplicationContext)}
 	 * should already have exited or should exit very soon after this method exits
 	 * <p>
-	 * 
+	 *
 	 * This method is only called to force an application to exit. This method will
 	 * not be called if an application exits normally from the
 	 * {@link #start(IApplicationContext)} method.
--- a/bundles/org.eclipse.equinox.launcher.tests/src/org/eclipse/equinox/launcher/tests/LauncherTests.java
+++ b/bundles/org.eclipse.equinox.launcher.tests/src/org/eclipse/equinox/launcher/tests/LauncherTests.java
@@ -16,6 +16,7 @@ package org.eclipse.equinox.launcher.tests;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import java.io.BufferedReader;
 import java.io.DataOutputStream;
@@ -23,6 +24,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.ServerSocket;
 import java.net.Socket;
+import java.net.SocketTimeoutException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -202,6 +204,66 @@ public class LauncherTests {
 		// Make sure launcher exited with code zero
 		launcherProcess.waitFor(5, TimeUnit.SECONDS);
 		assertEquals(0, launcherProcess.exitValue());
+	}
+
+	@Test
+	void test_appTerminatesWithoutNoRestartWithEXIT_OK() throws IOException, InterruptedException {
+		test_norestart(EXIT_OK, false);
+	}
+
+	@Test
+	void test_appTerminatesWithNoRestartAndEXIT_OK() throws IOException, InterruptedException {
+		test_norestart(EXIT_OK, true);
+	}
+
+	@Test
+	void test_appTerminatesWithNoRestartAndEXIT_RESTART() throws IOException, InterruptedException {
+		test_norestart(EXIT_RESTART, true);
+	}
+
+	@Test
+	void test_appTerminatesWithNoRestartAndEXIT_RELAUNCH() throws IOException, InterruptedException {
+		test_norestart(EXIT_RELAUNCH, true);
+	}
+
+	@Test
+	void test_appTerminatesWithoutNoRestartWithExitCode100() throws IOException, InterruptedException {
+		test_norestart(100, false);
+	}
+
+	@Test
+	void test_appTerminatesWithNoRestartAndExitCode100() throws IOException, InterruptedException {
+		test_norestart(100, true);
+	}
+
+	private void test_norestart(int exitCode, boolean addNoRestartArg) throws IOException, InterruptedException {
+		writeEclipseIni(DEFAULT_ECLIPSE_INI_CONTENT);
+
+		List<String> launcherArgs = addNoRestartArg ? List.of("--launcher.noRestart", "--launcher.suppressErrors")
+				: List.of("--launcher.suppressErrors");
+		Process launcherProcess = startEclipseLauncher(launcherArgs);
+
+		Socket socket = server.accept();
+
+		List<String> appArgs = new ArrayList<>();
+		analyzeLaunchedTestApp(socket, appArgs, null, exitCode);
+
+		// Make sure --launcher.noRestart arg is picked
+		assertEquals(addNoRestartArg, appArgs.contains("--launcher.noRestart"));
+		// Make sure launcher exited with expected exit value
+		launcherProcess.waitFor(5, TimeUnit.SECONDS);
+		assertEquals(exitCode, launcherProcess.exitValue());
+		try {
+			server.accept();
+			if (addNoRestartArg) {
+				fail("New eclipse started even with --launcher.noRestart arg and exit code " + exitCode);
+			} else {
+				fail("New eclipse started even with exit code " + exitCode);
+			}
+		} catch (SocketTimeoutException e) {
+			// No new instance launched
+			return;
+		}
 	}
 
 	@Test
--- a/features/org.eclipse.equinox.executable.feature/library/eclipse.c
+++ b/features/org.eclipse.equinox.executable.feature/library/eclipse.c
@@ -702,7 +702,7 @@ static int _run(int argc, _TCHAR* argv[], _TCHAR* vmArgs[])
 					}
 				}
 			} else {
-				 if (debug) {
+				if (debug) {
 					if (!suppressErrors)
 						displayMessage( title, shareMsg );
 					else
@@ -1103,12 +1103,15 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
 	(*vmArgv)[dst] = NULL;
 
 	/* Program arguments */
-    /*  OS <os> + WS <ws> + ARCH <arch> + LAUNCHER <launcher> + NAME <officialName> +
-     *  + LIBRARY <library> + SHOWSPLASH <cmd> + EXITDATA <cmd> + STARTUP <jar> + OVERRIDE/APPEND + argv[] + VM + <vm> +
-     * VMARGS + vmArg + requiredVMargs
+    /*  OS <os> + WS <ws> + ARCH <arch> + SHOWSPLASH <cmd> + LAUNCHER <program> + NAME <officialName>
+     *  + LIBRARY <eclipseLibrary> + STARTUP <jarFile> + PROTECT <protectMode> + APPEND/OVERRIDE
+     *  + EXITDATA <sharedId> + argv[] + VM <jniLib/javaVM> + VMARGS + vmArg[] + eeVMarg[] + reqVMarg[]
      *  + NULL)
      */
-    totalProgArgs  = 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 1 + argc + 2 + 1 + nVMarg + nEEargs + nReqVMarg + 1;
+    totalProgArgs = 2 + 2 + 2 + 2 + 2 + 2
+                  + 2 + 2 + 2 + 1
+                  + 2 + argc + 2 + 1 + nVMarg + nEEargs + nReqVMarg
+                  + 1;
 	*progArgv = malloc( totalProgArgs * sizeof( _TCHAR* ) );
     dst = 0;
 
--- a/features/org.eclipse.equinox.executable.feature/library/eclipse.c
+++ b/features/org.eclipse.equinox.executable.feature/library/eclipse.c
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2015 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -71,6 +71,7 @@
  *  -nosplash                  do not display the splash screen. The java application will
  *                             not receive the -showsplash command.
  *  -showsplash <bitmap>	   show the given bitmap in the splash screen.
+ *  --launcher.noRestart       disables the restart behavior of exit codes 23 and 24.
  *  -name <name>               application name displayed in error message dialogs and
  *                             splash screen window. Default value is computed from the
  *                             name of the executable - with the first letter capitalized
@@ -235,6 +236,7 @@ home directory.");
 static int     needConsole   = 0;				/* True: user wants a console	*/
 static int     debug         = 0;				/* True: output debugging info	*/
 static int     noSplash      = 0;				/* True: do not show splash win	*/
+static int     noRestart     = 0;				/* True: disables the restart behavior of the launcher */
 static int	   suppressErrors = 0;				/* True: do not display errors dialogs */
        int     secondThread  = 0;				/* True: start the VM on a second thread */
 static int     appendVmargs = 0;                /* True: append cmdline vmargs to launcher.ini vmargs */
@@ -286,6 +288,7 @@ static Option options[] = {
     { CONSOLELOG,	&needConsole,	VALUE_IS_FLAG,	0 },
     { DEBUG_ARG,	&debug,			VALUE_IS_FLAG,	0 },
     { NOSPLASH,     &noSplash,      VALUE_IS_FLAG,	1 },
+    { NORESTART,    &noRestart,     VALUE_IS_FLAG,	1 },
     { SUPRESSERRORS, &suppressErrors, VALUE_IS_FLAG, 1},
     { SECOND_THREAD, &secondThread, VALUE_IS_FLAG,  1 },
     { APPEND_VMARGS, &appendVmargs,	VALUE_IS_FLAG, 1 },
@@ -650,32 +653,35 @@ static int _run(int argc, _TCHAR* argv[], _TCHAR* vmArgs[])
 		case 0: /* normal exit */
 			break;
 		case RESTART_LAST_EC:
-			/* copy for relaunch, +1 to ensure NULL terminated */
-			relaunchCommand = malloc((initialArgc + 1) * sizeof(_TCHAR*));
-			memcpy(relaunchCommand, initialArgv, (initialArgc + 1) * sizeof(_TCHAR*));
-			relaunchCommand[initialArgc] = 0;
-			relaunchCommand[0] = program;
-			break;
-
+			if (!noRestart) {
+				/* copy for relaunch, +1 to ensure NULL terminated */
+				relaunchCommand = malloc((initialArgc + 1) * sizeof(_TCHAR*));
+				memcpy(relaunchCommand, initialArgv, (initialArgc + 1) * sizeof(_TCHAR*));
+				relaunchCommand[initialArgc] = 0;
+				relaunchCommand[0] = program;
+				break;
+			} // fall-through to default non-zero exit code handling.
 		case RESTART_NEW_EC:
-			if(launchMode == LAUNCH_EXE) {
-				if (exitData != NULL) free(exitData);
-				if (getSharedData( sharedID, &exitData ) != 0)
-					exitData = NULL;
-			}
-			if (exitData != 0) {
-				if (vmCommand != NULL) free( vmCommand );
-				vmCommand = parseArgList( exitData );
-				relaunchCommand = getRelaunchCommand(vmCommand);
-			} else {
-				if (debug) {
-					if (!suppressErrors)
-						displayMessage( officialName, shareMsg );
-					else
-						_ftprintf(stderr, _T_ECLIPSE("%s:\n%s\n"), officialName, shareMsg);
+			if (!noRestart) {
+				if(launchMode == LAUNCH_EXE) {
+					if (exitData != NULL) free(exitData);
+					if (getSharedData( sharedID, &exitData ) != 0)
+						exitData = NULL;
 				}
-			}
-			break;
+				if (exitData != 0) {
+					if (vmCommand != NULL) free( vmCommand );
+					vmCommand = parseArgList( exitData );
+					relaunchCommand = getRelaunchCommand(vmCommand);
+				} else {
+					if (debug) {
+						if (!suppressErrors)
+							displayMessage( officialName, shareMsg );
+						else
+							_ftprintf(stderr, _T_ECLIPSE("%s:\n%s\n"), officialName, shareMsg);
+					}
+				}
+				break;
+			} // fall-through to default non-zero exit code handling.
 		default: {
 			_TCHAR *title = _tcsdup(officialName);
 			errorMsg = NULL;
@@ -1104,12 +1110,12 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
 
 	/* Program arguments */
     /*  OS <os> + WS <ws> + ARCH <arch> + SHOWSPLASH <cmd> + LAUNCHER <program> + NAME <officialName>
-     *  + LIBRARY <eclipseLibrary> + STARTUP <jarFile> + PROTECT <protectMode> + APPEND/OVERRIDE
+     *  + LIBRARY <eclipseLibrary> + STARTUP <jarFile> + PROTECT <protectMode> + APPEND/OVERRIDE + NORESTART
      *  + EXITDATA <sharedId> + argv[] + VM <jniLib/javaVM> + VMARGS + vmArg[] + eeVMarg[] + reqVMarg[]
      *  + NULL)
      */
     totalProgArgs = 2 + 2 + 2 + 2 + 2 + 2
-                  + 2 + 2 + 2 + 1
+                  + 2 + 2 + 2 + 1 + 1
                   + 2 + argc + 2 + 1 + nVMarg + nEEargs + nReqVMarg
                   + 1;
 	*progArgv = malloc( totalProgArgs * sizeof( _TCHAR* ) );
@@ -1159,6 +1165,11 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
 
 	/* override or append vm args */
 	(*progArgv)[ dst++ ] = appendVmargs ? APPEND_VMARGS : OVERRIDE_VMARGS;
+
+	/* 'No restart' argument, if applicable. */
+	if (noRestart) {
+		(*progArgv)[ dst++ ] = NORESTART;
+	}
 
 	/* Append the exit data command. */
 	if (sharedID) {
--- a/features/org.eclipse.equinox.compendium.sdk/feature.xml
+++ b/features/org.eclipse.equinox.compendium.sdk/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.equinox.compendium.sdk"
       label="%featureName"
-      version="3.23.200.qualifier"
+      version="3.23.300.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/features/org.eclipse.equinox.executable.feature/library/eclipseCommon.h
+++ b/features/org.eclipse.equinox.executable.feature/library/eclipseCommon.h
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2009 IBM Corporation and others.
+ * Copyright (c) 2006, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -47,6 +47,7 @@
 #define INI                _T_ECLIPSE("--launcher.ini")
 #define APPEND_VMARGS      _T_ECLIPSE("--launcher.appendVmargs")
 #define OVERRIDE_VMARGS    _T_ECLIPSE("--launcher.overrideVmargs")
+#define NORESTART          _T_ECLIPSE("--launcher.noRestart")
 #define SECOND_THREAD      _T_ECLIPSE("--launcher.secondThread")
 #define PERM_GEN           _T_ECLIPSE("--launcher.XXMaxPermSize")
 #define OLD_ARGS_START     _T_ECLIPSE("--launcher.oldUserArgsStart")
--- a/features/org.eclipse.equinox.sdk/feature.xml
+++ b/features/org.eclipse.equinox.sdk/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.equinox.sdk"
       label="%featureName"
-      version="3.23.1300.qualifier"
+      version="3.23.1400.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">


merks (changes requested):
I wonder how many of these places where I see -nosplash that need some attention for this new argument?

image

merks:
@HannesWell 

Does this look okay to you?  Will anything special need to be done to actually build new launchers or will that "just work" given all your build improvements?

@tjwatson 

Do you think something with deeper knowledge about the launcher needs to review this.  It appears okay to me...

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -235,6 +236,7 @@ home directory.");
 static int     needConsole   = 0;				/* True: user wants a console	*/
 static int     debug         = 0;				/* True: output debugging info	*/
 static int     noSplash      = 0;				/* True: do not show splash win	*/
+static int     noRestart     = 0;               /* True: disables the restart behavior of the launcher */
```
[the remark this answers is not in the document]
dhendriks: I adapted it to match `noSplash`.

bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/app/IApplication.java
```diff
@@ -34,7 +34,11 @@ public interface IApplication {
 	public static final Integer EXIT_OK = Integer.valueOf(0);
 
 	/**
-	 * Exit object requesting platform restart
+	 * Exit object requesting platform restart.
+	 *
+	 * <p>
+	 * Note: This special exit code may be disabled by a launcher argument.
```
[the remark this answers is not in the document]
dhendriks: Thanks for the review.

* I checked the plugins within this repo for `nosplash` and made changes as needed for `norestart`.
* If other changes are needed in other repos, I assume we need to do that as follow-ups, in different merge requests? Seems to concern only documentation.
* Updated the JavaDocs in `IApplication` to address your comments.
* I did some other fixes, as the build/tests also failed.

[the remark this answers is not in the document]
merks: Yes, changes in other repos must be done as separate pull requests that of course can come later.

HannesWell:
> Will anything special need to be done to actually build new launchers or will that "just work" given all your build improvements?
> 
The GitHub workflows already pick up the change and build and test the new code.
I'm currently working on https://github.com/eclipse-equinox/equinox/issues/575 and hope to complete this within the next one or two weeks. Actually this would be a nice first real-world use and test case.
So unless it is urgent, it would be great if we could await the completion of https://github.com/eclipse-equinox/equinox/issues/575 before submitting this which would also avoid the need to subsequently trigger the launcher build manually.

> Does this look okay to you?

I'm not very familiar with the code itself, but what I see looks sane to me.

@dhendriks it would be great if you could add a test-case to the existing equinox.launcher.tests. @umairsair could you give some advice how to test this best?

umairsair:
> @umairsair could you give some advice how to test this best?

I think following three tests should be added.. I haven't run the tests though..

```
	@Test
	void test_appTerminatesWithNoRestartAndEXIT_OK() throws IOException, InterruptedException {
		test_norestart(EXIT_OK);
	}
	
	@Test
	void test_appTerminatesWithNoRestartAndEXIT_RESTART() throws IOException, InterruptedException {
		test_norestart(EXIT_RESTART);
	}
	
	@Test
	void test_appTerminatesWithNoRestartAndEXIT_RELAUNCH() throws IOException, InterruptedException {
		test_norestart(EXIT_RELAUNCH);
	}

	private void test_norestart(int exitCode) throws IOException, InterruptedException {
		writeEclipseIni(DEFAULT_ECLIPSE_INI_CONTENT);

		Process launcherProcess = startEclipseLauncher(List.of("-norestart"));

		Socket socket = server.accept();

		List<String> appArgs = new ArrayList<>();
		analyzeLaunchedTestApp(socket, appArgs, null, exitCode);

		// Make sure -norestart arg is picked
		assertTrue(appArgs.contains("-norestart"));
		// Make sure launcher exited with expected exit value
		launcherProcess.waitFor(5, TimeUnit.SECONDS);
		assertEquals(exitCode, launcherProcess.exitValue());
		try {
			server.accept();
			fail("New eclipse started even with -norestart arg and exit code " + exitCode):
		} catch (SocketTimeoutException e) {
			// No new instance launched
			return;
		}
	}
```

dhendriks:
> > @umairsair could you give some advice how to test this best?
> 
> I think following three tests should be added.. I haven't run the tests though..

I added the tests. Let's see if they run OK.

dhendriks:
> I'm currently working on #575 and hope to complete this within the next one or two weeks. Actually this would be a nice first real-world use and test case.
> So unless it is urgent, it would be great if we could await the completion of #575 before submitting this which would also avoid the need to subsequently trigger the launcher build manually.

Let's try to get this in a good shape to be merged now already. For me it is OK to then merge in a few weeks.

umairsair:
so we are not passing `-norestart` as parameter to java application, see `getVMCommand(..)`, thats why following assertion is failing..

```
assertTrue(appArgs.contains("-norestart"));
```

IMO we should pass it to application so that if java application wants to make any decision based on `-norestart` arg..

laeubi:
> IMO we should pass it to application so that if java application wants to make any decision based on -norestart arg..

I just wanted to note that one should not confuse JVM argument, Application Arguments and launcher/runtime Arguments.

If you [look here](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fruntime-options.html) then launcher / runtime arguments are usually mapped as **java system properties** and not passed directly to the invoked application!

dhendriks:
So, `-norestart` is a launcher argument, I assume? Then somewhere in `eclipse.c` I need to add a `-Dlauncher.noRestart=true` or so? I tried looking for examples of where this is done already, but didn't find any. Could you point me in the right direction?

umairsair:
see following for reference

https://github.com/eclipse-equinox/equinox/blob/49be75d81307d259b2ec5f01493c284253660fe3/features/org.eclipse.equinox.executable.feature/library/eclipse.c#L1158

you can add following code after above line..

```
if (noRestart)
    (*progArgv)[ dst++ ] = "--launcher.noRestart";
```

you can create define for `--launcher.noRestart` in eclipseCommon.h and update the tests accordingly to check for `--launcher.noRestart`

dhendriks:
> see following for reference
> [...]
> you can create define for `--launcher.noRestart` in eclipseCommon.h and update the tests accordingly to check for `--launcher.noRestart`

I added it all:
* Not sure what the define in `eclipseCommon.h` is for, as it doesn't seem to be used then.
* When updating `eclipse.c`, I found it difficult to link the argument comment to the argument size computation, so I made them more alike. I also updated the argument description, as it didn't fully match the code (also making it more difficult to keep it in sync).
* I updated the test. Let's see if it succeeds now.

dhendriks:
@umairsair I implemented it as you indicated. But the tests fail. Partial stack trace:

```
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
	at org.eclipse.equinox.launcher.tests.LauncherTests.test_norestart(LauncherTests.java:234)
	at org.eclipse.equinox.launcher.tests.LauncherTests.test_appTerminatesWithNoRestartAndEXIT_RESTART(LauncherTests.java:215)
```

Stdout:

```
--- start ----
-os
linux
-ws
gtk
-arch
x86_64
-showsplash
-launcher
/tmp/junit15564327931400162296/eclipse/eclipse
-name
Eclipse
--launcher.library
/tmp/junit15564327931400162296/eclipse//plugins/org.eclipse.equinox.launcher/eclipse_11900.so
-startup
/tmp/junit15564327931400162296/eclipse//test.launcher.jar
--launcher.overrideVmargs
-exitdata
4
-norestart
-vm
/opt/tools/java/temurin/jdk-17/latest/bin/java
-vmargs
-Xms40m
-jar
/tmp/junit15564327931400162296/eclipse//test.launcher.jar
--- end ----
```

It seems `-norestart` is present, while `--launcher.noRestart` is not. Do you know what is going on?

umairsair:
Probably some local setup issue.

I see that tests are not executed for latest commit and requires approval. @merks , @HannesWell  can you please approve the run so that we can analyze the latest test results?

dhendriks:
All 3 tests fail, like this:

```
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
	at org.eclipse.equinox.launcher.tests.LauncherTests.test_norestart(LauncherTests.java:234)
	at org.eclipse.equinox.launcher.tests.LauncherTests.test_appTerminatesWithNoRestartAndEXIT_OK(LauncherTests.java:210)
```

So, on the assertion:

```
assertTrue(appArgs.contains("--launcher.noRestart"));
```

I looked at the code, and the changes I did previously. I have to clue what to do next. Did I not change it correctly, to pass that argument on from the launcher to the application?

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -1157,6 +1166,10 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
 	/* override or append vm args */
 	(*progArgv)[ dst++ ] = appendVmargs ? APPEND_VMARGS : OVERRIDE_VMARGS;
 
+	/* 'No restart' argument, if applicable. */
+	if (noRestart)
+		(*progArgv)[ dst++ ] = "--launcher.noRestart";
```
umairsair, line 1171:
```suggestion
		(*progArgv)[ dst++ ] = LAUNCHER_NORESTART;
```

umairsair:
last run tells that tests on windows are unstable but failing on linux.. on windows, I hope the suggestion I made above will solve it.. for linux, launcher is not exiting.. that needs to be checked..

https://github.com/eclipse-equinox/equinox/pull/595/checks?check_run_id=24349592476

also macos build is failing..

https://github.com/eclipse-equinox/equinox/actions/runs/8833610849/job/24349296333?pr=595

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -650,32 +653,35 @@ static int _run(int argc, _TCHAR* argv[], _TCHAR* vmArgs[])
...
-			relaunchCommand[0] = program;
-			break;
-
+			if (!noRestart) {
+				/* copy for relaunch, +1 to ensure NULL terminated */
+				relaunchCommand = malloc((initialArgc + 1) * sizeof(_TCHAR*));
+				memcpy(relaunchCommand, initialArgv, (initialArgc + 1) * sizeof(_TCHAR*));
+				relaunchCommand[initialArgc] = 0;
+				relaunchCommand[0] = program;
+				break;
```
umairsair, line 662: the break of case statment should still be outside of if condition, right? otherwise any data put in exitData buffer by the application will be considered error message in `default` case..

dhendriks: The essence is:

```java
		case RESTART_LAST_EC:
			if (!noRestart) {
                                ...
				break;
			} // else: fall-through to 'default' case.
		case RESTART_NEW_EC:
			if (!noRestart) {
                                ...
				break;
			} // else: fall-through to 'default' case.
		default: {
			...
```

If the exit code is a restart exit code, then if we allow restart (we do not disallow restarting), we do something special (whatever was there already before) and then break. But, if restart is disabled, we instead fall through to the `default` case, and handle it like any other non-zero exit code.

So, that part still looks good to me.

dhendriks:
macOS build is indeed failing. But it indicates:

>>>
**Set up JDK 17**
Run actions/setup-java@v4
Installed distributions
  Trying to resolve the latest version from remote
  Error: Could not find satisfied version for SemVer '8'. 
  Available versions: 22.0.1+8, 22.0.0+36, 21.0.3+9.0.LTS, 21.0.2+13.0.LTS, 21.0.1+12.0.LTS, 21.0.0+35.0.LTS, 20.0.2+9, 20.0.1+9, 20.0.0+36, 19.0.2+7, 19.0.1+10, 19.0.0+36, 18.0.2+101, 18.0.2+9, 18.0.1+10, 18.0.0+36, 17.0.11+9, 17.0.10+7, 17.0.9+9, 17.0.8+101, 17.0.8+7, 17.0.7+7, 17.0.6+10, 17.0.5+8, 17.0.4+101, 17.0.4+8, 17.0.3+7, 17.0.2+8, 17.0.1+12, 17.0.0+35, 11.0.23+9, 11.0.22+7.1, 11.0.22+7, 11.0.21+9, 11.0.20+101, 11.0.20+8, 11.0.19+7, 11.0.18+10, 11.0.17+8, 11.0.16+101, 11.0.16+8, 11.0.15+10
>>>

This seems to be a setup thing even before any code from the repo is used, so I doubt it is because of anything I changed.

dhendriks:
There appear to be two runs for each test. Not sure why. The first ones fail like on Windows. The 2nd ones indeed because `process hasn't exited`. Hopefully this is also fixed by your proposed change (which I applied).

dhendriks:
@merks / @HannesWell: Can you please approve the new build/test runs?

dhendriks:
The `org.opentest4j.AssertionFailedError: expected: <true> but was: <false>` issues have been resolved. Tests were run for all three platforms now. The `test_appTerminatesWithNoRestartAndEXIT_OK` test now succeeds. The `test_appTerminatesWithNoRestartAndEXIT_RESTART` and `test_appTerminatesWithNoRestartAndEXIT_RELAUNCH` tests both fail with the following error on all three platforms:

```
process has not exited
java.lang.IllegalThreadStateException: process has not exited
	at java.base/java.lang.ProcessImpl.exitValue(ProcessImpl.java:560)
	at org.eclipse.equinox.launcher.tests.LauncherTests.test_norestart(LauncherTests.java:238)
```

I have no clue how to debug what is going wrong, and why the process does not terminate.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
umairsair: and this is infact the reason of tests failure..

test app set exit data here if case from within application user select restart.

https://github.com/eclipse-equinox/equinox/blob/99a85789ed29f87d4ed04d15bb640de325f4750f/bundles/org.eclipse.equinox.launcher.tests/src/org/eclipse/equinox/launcher/TestLauncherApp.java#L89

and in launcher, as `-norestart` is mentioned and `default` is executed, this exitData is shown as error..

https://github.com/eclipse-equinox/equinox/blob/1551906df8e01c74351712db0d602495458fa7fd/features/org.eclipse.equinox.executable.feature/library/eclipse.c#L735

HannesWell:
Just as a quick note: all the launcher binaries are now build in the Jenkins pipeline if anything changed.
Therefore please rebase your future adjustments on the master.
Also please squash all your commits into one with a message that covers the total change (there is no need to record the development loops in git)

I'm not very familiar with the code itself, but I can try to have a look at this tomorrow.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
dhendriks: I was trying to only skip the `RESTART_*` cases, but I think you're saying also `exitData` may contain something related to restart behavior? I'm really not sure what that variable is about, at all.

`eclipse.c` has more than 2000 lines. I have to say there is a lot of stuff going on there, and I don't know what half of it means. It is complicated.

Can the `default` case really be skipped fully? It seems to handle reporting various errors. For instance, the `errorMsg == NULL` seems relevant, as it reports problems launching Java? Are you sure we can safely skip all of that? I don't want it to print anything in case of a non-zero exit code, but it should probably still report issues for failing to launch Java to begin with.

Should we do `if (exitData != NULL) free(exitData);` before falling through? And what is `getSharedData` about? That can also set `exitData` again?

umairsair: IMO moving `break` out of `if` conditions will fix the issue..

dhendriks: > IMO moving `break` out of `if` conditions will fix the issue..

Yes, I got that impression as well. But, while that may fix the test, will that not prevent too many other errors from being reported as well (see my previous comment).

umairsair: it would become as first case where application is expected to just exit..

https://github.com/eclipse-equinox/equinox/blob/2ef1133fd652c1aef50e41a0b3369ad6ed11eb81/features/org.eclipse.equinox.executable.feature/library/eclipse.c#L653-L654

dhendriks: Yes, I get that. But is that desired? I don't think users would expect that if they disable restart behavior, that they suddenly also get errors for failing to start Java reported in a different way. It would fit my use case just fine, but I'm not sure it is really logical for others?

umairsair: >  if they disable restart behavior, that they suddenly also get errors for failing to start Java reported in a different way.

this will not happen if you break from all cases.. in which code flow you think this will happen?

dhendriks: If `javaResults->launchResult` is `23` or `24` and `javaResults->runResult` is zero, we'd get into the restart cases and fall through to the `default` (assuming the keep the `break`s where they are now). If `launchMode != LAUNCH_EXE` and `exitData == 0`, we'd go into `errorMsg == null` and since `javaResults->runResult` is not taken as branch in the `if`, and assuming `javaResults->errorMessage != NULL`, we set `errorMsg`, and just below we'd print that. By moving the `break` statements, and thus not falling through to the `default` case, this is not reported anymore. I think it is still valuable to report issues in launching Java, as otherwise a normal exit code can't be distinguished anymore from a failure to launch Java.

dhendriks: @umairsair What do you think of these cases?

umairsair: > assuming javaResults->errorMessage != NULL, we set errorMsg, and just below we'd print that. 

the code comment on it says that we have a problem launching java, which means that in this flow user app will not have been run and not returned 23 or 24.. we need to completely follow the same code flow as before, the only change we are doing in code flow is to not restart java/jvm if --norestart is set and it'll be achieved by simply conditioning the code that is doing the restart and let the remaining code flow same, and it would be achieved if you simply put case 23 and 24 code under condition except break statement..

BTW did you try running tests after moving break out of condition? does it solve the test failures?

dhendriks: * I've moved the `break` statements as suggested by @umairsair. Let's see if that works. Could someone allow another build?
* I've rebased on `master` and squashed all changes to a single commit, as requested.

dhendriks: Build failed:

```
Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.equinox.app: Only qualifier changed for (org.eclipse.equinox.app/1.7.100.v20240608-1512). Expected to have bigger x.y.z than what is available in baseline (1.7.100.v20240321-1445) -> [Help 1]
```

I tried changing the version in the manifest of `org.eclipse.equinox.app` from `1.7.100.qualifier` to `1.7.101.qualifier`, but that gives a warning: `The service version is increased unnecessarily since either the major or minor or service version is already increased`.

What should I do to get the build working?

umairsair: update version to `1.7.200.qualifier` at following location

https://github.com/eclipse-equinox/equinox/blob/5851137b9bbfad795a9f0c00c5646f9aa80bde78/bundles/org.eclipse.equinox.app/META-INF/MANIFEST.MF#L5

dhendriks: I changed the version.

dhendriks: I also changed the feature, as it then complained about that. It seems the latest build was 'aborted'. Could someone approve or restart the build/test?

iloveeclipse: I've retriggered the build

dhendriks: I had to recursively adapt 4 more features. I hope that is all the versions that need updating. Needs another build approval.

dhendriks: If I understand it correctly, the new tests pass. There is one test that fails: `testCoordinatedConfigurationOnBeforeRegisteredManagedService (org.osgi.test.cases.cm.junit.CMCoordinationTestCase) failed. Not sure what that is, or where it comes from, and whether that is even related to the changes I made? Also, 2/23 checks were aborted. I also have no idea why that is.

HannesWell:
In case it is helpful for you: I'm currently working on improving the documentation about how the binaries can be built locally in https://github.com/eclipse-equinox/equinox/pull/640. This also adds a maven build launch config to simply build the binaries for the current platform. With that you could build this change locally and debug the tests.
But I can also look into this next week.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
dhendriks: The `CMCoordinationTestCase` failed test also fails in #640 that @HannesWell [refers to](https://github.com/eclipse-equinox/equinox/pull/595#issuecomment-2164057367) in a different thread of this PR. I don't think that failure is related to my changes.

I still don't know what caused the 2 other checks to be aborted. Their output doesn't help me much. Although, I see they were 'aborted'; they didn't 'fail'. I also in Eclipse ESCET recently got timeout aborts due to the Jenkins cluster being slower than usual, and I had to increase the timeout there. Could that be the reason here as well?

dhendriks:
> In case it is helpful for you: I'm currently working on improving the documentation about how the binaries can be built locally in #640. This also adds a maven build launch config to simply build the binaries for the current platform. With that you could build this change locally and debug the tests. But I can also look into this next week.

Sure, that would be / have been helpful. But, the [current build failures/aborts](https://github.com/eclipse-equinox/equinox/pull/595#discussion_r1637656279) do not seem related to this.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
HannesWell: I have not yet read to this entire thread, but my first impression when looking into the change itself is that in case a no-restart is present the code should fall through to the default branch and tread the exit code as error (which it in fact is, isn't it)?

By default the executable pops-up a native error dialog if the app exited with an error code. 
Could the test-failure with a fall-through to default be because of that? Because at the server such pop-up is closed by nobody.
Maybe the test would be more correct if you use from the [Eclipse runtime options](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fruntime-options.html)
```
--launcher.suppressErrors (Executable)
    If specified the executable will not display any error or message dialogs.  This is useful if the executable is being used in an unattended situation.
```
Then the error code is just returned by the command and could be checked in the invoking java test code?

HannesWell (commented):
> > In case it is helpful for you: I'm currently working on improving the documentation about how the binaries can be built locally in #640. This also adds a maven build launch config to simply build the binaries for the current platform. With that you could build this change locally and debug the tests. But I can also look into this next week.
> 
> Sure, that would be / have been helpful. But, the [current build failures/aborts](https://github.com/eclipse-equinox/equinox/pull/595#discussion_r1637656279) do not seem related to this.

This happened because one of the build-agents went offline, but I reported it yesterday and it is back online:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4726

The current failure is because all the native launcher fragments need a version bumps since this is the first time in this release cycle they get changed.
But I have just submitted another PR that had to bump them as well, so you should be saved from that now when rebasing on the latest master.

bundles/org.eclipse.equinox.launcher.tests/src/org/eclipse/equinox/launcher/tests/LauncherTests.java
```diff
@@ -204,6 +206,45 @@ void test_appTerminatesWithCodeZeroOnExit() throws IOException, InterruptedExcep
...
+		// Make sure launcher exited with expected exit value
+		launcherProcess.waitFor(5, TimeUnit.SECONDS);
+		assertEquals(exitCode, launcherProcess.exitValue());
+		try {
+			server.accept();
+			fail("New eclipse started even with -norestart arg and exit code " + exitCode);
+		} catch (SocketTimeoutException e) {
+			// No new instance launched
+			return;
+		}
```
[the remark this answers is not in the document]
dhendriks: You proposal does not close the server, leading to a resource leak warning. I kept it as is.

dhendriks:
> I have not yet read to this entire thread, but my first impression when looking into the change itself is that in case a no-restart is present the code should fall through to the default branch and tread the exit code as error (which it in fact is, isn't it)?

I completely agree. I had that at first. @umairsair [didn't think that was correct](https://github.com/eclipse-equinox/equinox/pull/595#discussion_r1597044496). I tried to argue for keeping it, but gave up. It's been 2 months since I created this merge request! It is not easy to adapt this stuff and know what to adapt exactly, when you're not familiar with this code base, build infrastructure, test setup, etc. I'm happy you now agree with me. I want to only skip the restart behavior, making it be treated as any other non-zero exit code. That's all I'm trying to do.

I adapted the code to fall-through again. I also added the `--launcher.suppressErrors` argument, as suggested. Can you approve another build, to see how this works out now?

dhendriks:
> I adapted the code to fall-through again. I also added the `--launcher.suppressErrors` argument, as suggested. Can you approve another build, to see how this works out now?

It works! The build and all tests related to this PR succeed!

But, there is one test that fails: `testCoordinatedConfigurationOnBeforeRegisteredManagedService (org.osgi.test.cases.cm.junit.CMCoordinationTestCase) failed`. Does anyone know what this test tests, where it comes from, and what it has to do with the changes in this branch?

laeubi:
The test failure is "normal" see:
- https://github.com/osgi/osgi/issues/718

merks:
@tjwatson 

Do you plan to review?   

I think @umairsair  is probably the most qualified having working on details not so long ago...

In any case, early in the release cycle is a good ti to merge...

HannesWell (changes requested):
> > I adapted the code to fall-through again. I also added the `--launcher.suppressErrors` argument, as suggested. Can you approve another build, to see how this works out now?
> 
> It works! The build and all tests related to this PR succeed!
> 

Awesome! In general this change looks fine to me now.
But I think we should should test cases for the following cases:
- `test_norestart` with some other error code, e.g. `100` with and without `-norestart` being passed.
- test that `restart` and `relaunch` still work as expected when `-norestart` is not supplied.

If necessary I can suggest specific test-cases.

Btw. was it already discussed if the flag should be named  `-norestart` or `--launcher.norestart`? From the top of my head I don't remember the difference, but IIRC there is one altough on first sight, it is not clear when `--launcher` is prefixed.

Furthermore could you please move all the version bumps into a separate first commit of this PR with message `Version bumps for 4.33 stream` to make it easier to find.
I know I have asked you to squash the changes before, but this is different since it separates the outcome of the PR.


A review from somebody who is more an expert at the launcher code would still be welcome.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -1157,6 +1166,10 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
 	/* override or append vm args */
 	(*progArgv)[ dst++ ] = appendVmargs ? APPEND_VMARGS : OVERRIDE_VMARGS;
 
+	/* 'No restart' argument, if applicable. */
+	if (noRestart)
+		(*progArgv)[ dst++ ] = LAUNCHER_NORESTART;
+
```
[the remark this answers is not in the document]
dhendriks: I added the parentheses.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -1103,12 +1109,15 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
...
+    /*  OS <os> + WS <ws> + ARCH <arch> + SHOWSPLASH <cmd> + LAUNCHER <program> + NAME <officialName>
+     *  + LIBRARY <eclipseLibrary> + STARTUP <jarFile> + PROTECT <protectMode> + APPEND/OVERRIDE + NORESTART
+     *  + EXITDATA <sharedId> + argv[] + VM <jniLib/javaVM> + VMARGS + vmArg[] + eeVMarg[] + reqVMarg[]
      *  + NULL)
      */
-    totalProgArgs  = 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 1 + argc + 2 + 1 + nVMarg + nEEargs + nReqVMarg + 1;
+    totalProgArgs = 2 + 2 + 2 + 2 + 2 + 2
+                  + 2 + 2 + 2 + 1 + 1
+                  + 2 + argc + 2 + 1 + nVMarg + nEEargs + nReqVMarg
+                  + 1;
```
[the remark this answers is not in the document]
dhendriks: I moved the cleanup part to a separate commit.

dhendriks:
> [...] In general this change looks fine to me now. But I think we should should test cases for the following cases:
> 
> * `test_norestart` with some other error code, e.g. `100` with and without `-norestart` being passed.
> * test that `restart` and `relaunch` still work as expected when `-norestart` is not supplied.

I added tests for exit code 100, with and without `-norestart`. I also added a test for exit code zero without `-norestart`.

There are already tests for testing the restart/relaunch behaviors (exit codes 23 and 24). They restart/relaunch even multiple times, etc. They don't supply `-norestart`. So I think that is already covered by existing tests.

Probably somebody needs to approve another build to see if the new/changed tests succeed.

> Btw. was it already discussed if the flag should be named `-norestart` or `--launcher.norestart`? From the top of my head I don't remember the difference, but IIRC there is one altough on first sight, it is not clear when `--launcher` is prefixed.

We have both `-norestart` and `--launcher.noRestart` for the app and launcher. I was advised to use these names. Let me know if it needs to be changed.

> Furthermore could you please move all the version bumps into a separate first commit of this PR with message `Version bumps for 4.33 stream` to make it easier to find. I know I have asked you to squash the changes before, but this is different since it separates the outcome of the PR.

I moved all separate changes to separate commits.

> A review from somebody who is more an expert at the launcher code would still be welcome.

You asked earlier if @umairsair and @tjwatson could take a look. Maybe they can now have a look?

umairsair:
> I have not yet read to this entire thread, but my first impression when looking into the change itself is that in case a no-restart is present the code should fall through to the default branch and tread the exit code as error (which it in fact is, isn't it)?

I differ here.. consider the case of eclipse where user passed -norestart to eclipse launcher and later used restart option from IDE.. in this case, the shared buffer will contain the launcher args which will be appearing as error pop-up on restart..

if we really want to go this route, we should add support in equinox application as well to either hide restart option or give error to user that this action is unavailable..

dhendriks:
> > I have not yet read to this entire thread, but my first impression when looking into the change itself is that in case a no-restart is present the code should fall through to the default branch and tread the exit code as error (which it in fact is, isn't it)?
> 
> I differ here.. consider the case of eclipse where user passed -norestart to eclipse launcher and later used restart option from IDE.. in this case, the shared buffer will contain the launcher args which will be appearing as error pop-up on restart..
> 
> if we really want to go this route, we should add support in equinox application as well to either hide restart option or give error to user that this action is unavailable..

I think that it does not matter for this whether we fall through or handle exit codes 23 and 24 as other exit codes. In both cases, if a user requests a restart in the IDE and we disabled the special handling of these exit codes, then restart does not work.

I don't think the code for the IDE restart command isn't in this repo. Together with other identified points that need to be done in other repos / other pull requests, this gives the following list of next steps after this merge request:

* Adapt equinox application 'restart' command to inform user that action is unavailable if launcher was given the no restart option.
* Update documentation of the launcher for the new no restart option.

@umairsair Is that OK?

dhendriks:
@tjwatson Could you also still review this pull request?

dhendriks:
I rebased to fix a merge conflict on version bumps.

HannesWell:
> > [...] In general this change looks fine to me now. But I think we should should test cases for the following cases:
> > 
> > * `test_norestart` with some other error code, e.g. `100` with and without `-norestart` being passed.
> > * test that `restart` and `relaunch` still work as expected when `-norestart` is not supplied.
> 
> I added tests for exit code 100, with and without `-norestart`. I also added a test for exit code zero without `-norestart`.
> 
> There are already tests for testing the restart/relaunch behaviors (exit codes 23 and 24). They restart/relaunch even multiple times, etc. They don't supply `-norestart`. So I think that is already covered by existing tests.

Yes that's indeed already fine. Perfect, thanks for the updates.
Having the indentation fix in a separate commit is IMHO a bit too fine-grained. I have now squashed that into the other clean-up commit and moved the addition of the norestart option into the commit that adds the processing (I believe this was unintentional).

> > Btw. was it already discussed if the flag should be named `-norestart` or `--launcher.norestart`? From the top of my head I don't remember the difference, but IIRC there is one altough on first sight, it is not clear when `--launcher` is prefixed.
> 
> We have both `-norestart` and `--launcher.noRestart` for the app and launcher. I was advised to use these names. Let me know if it needs to be changed.

I'll look it up. Until know I have not yet fully understand the logic here. Is it necessary to have both defined?
@tjwatson can you advice here?

> > > I have not yet read to this entire thread, but my first impression when looking into the change itself is that in case a no-restart is present the code should fall through to the default branch and tread the exit code as error (which it in fact is, isn't it)?
> > 
> > 
> > I differ here.. consider the case of eclipse where user passed -norestart to eclipse launcher and later used restart option from IDE.. in this case, the shared buffer will contain the launcher args which will be appearing as error pop-up on restart..
> > if we really want to go this route, we should add support in equinox application as well to either hide restart option or give error to user that this action is unavailable..
> 
> I think that it does not matter for this whether we fall through or handle exit codes 23 and 24 as other exit codes. In both cases, if a user requests a restart in the IDE and we disabled the special handling of these exit codes, then restart does not work.
> 
> I don't think the code for the IDE restart command isn't in this repo. Together with other identified points that need to be done in other repos / other pull requests, this gives the following list of next steps after this merge request:
> 
>     * Adapt equinox application 'restart' command to inform user that action is unavailable if launcher was given the no restart option.
> 
>     * Update documentation of the launcher for the new no restart option.
> 

I understand @umairsair's point there, but I'm not yet sure about the conclusion. If one passes `-norestart` it effectively means that these exit code should be treated as errors and as @dhendriks said it means a restart/relaunch does not work.
Furthermore I assume that the norestart mode is mainly used for command-line applications and not for applications with an UI, isn't it?
Of course the restart menu entry could be disabled in the UI in this case, the code is probably in https://github.com/eclipse-platform/eclipse.platform.ui. But I would consider this as a nice to have but not a strict requirement. Still it should be documented.

@tjwatson do you have an opinion on this topic?

HannesWell (commented):
@dhendriks sorry for the delay.
I have not found any definitive documentation/guide-line how to name the arguments, but since the argument to add is definitively only handled by the launcher.
Therefore I would be in favor of naming the flag `--launcher.noRestart`, which is similar to for example `--launcher.suppressErrors`.
I have made a few comments below what I think needs to be changed.

Can you please try if that works, i.e. apply the changes and see the results in the CI? If it works I think this can be submitted.

If I understand the current implementation correct we have a `-noRestart` argument that controls the launcher behavior but only the presence of `--launcher.noRestart` is passed to the Eclipse application?

dhendriks:
> @dhendriks sorry for the delay. I have not found any definitive documentation/guide-line how to name the arguments, but since the argument to add is definitively only handled by the launcher. Therefore I would be in favor of naming the flag `--launcher.noRestart`, which is similar to for example `--launcher.suppressErrors`. I have made a few comments below what I think needs to be changed.
> 
> Can you please try if that works, i.e. apply the changes and see the results in the CI? If it works I think this can be submitted.
> 
> If I understand the current implementation correct we have a `-noRestart` argument that controls the launcher behavior but only the presence of `--launcher.noRestart` is passed to the Eclipse application?

The way I understood it, is: There are options for the launcher. They are use short names. When passed on to the application, they get the longer form to indicate it is an option from the launcher, as an application also has options of itself.
Hence, in my version, the launcher got a `-norestart` and the Eclipse application got `--launcher.noRestart`.
There are separate lists in `eclipseCommon.h` for this very reason, I thought.
But, it was already not consistently applied, so I may be misinterpreting the naming idea: I checked the defined constants, and the ones that are options of the launcher have a one space indentation:

```
 #define CONSOLE            _T_ECLIPSE("-console")
 #define CONSOLELOG         _T_ECLIPSE("-consoleLog")
 #define DEBUG_ARG          _T_ECLIPSE("-debug") // Post-fixing with _ARG. Otherwise it causes failure in macOS build
 #define OS                 _T_ECLIPSE("-os")
 #define OSARCH             _T_ECLIPSE("-arch")
 #define NOSPLASH           _T_ECLIPSE("-nosplash")
 #define NORESTART          _T_ECLIPSE("-norestart")
#define LAUNCHER           _T_ECLIPSE("-launcher")
 #define SHOWSPLASH         _T_ECLIPSE("-showsplash")
#define EXITDATA           _T_ECLIPSE("-exitdata")
 #define STARTUP            _T_ECLIPSE("-startup")
 #define VM                 _T_ECLIPSE("-vm")
 #define WS                 _T_ECLIPSE("-ws")
 #define NAME               _T_ECLIPSE("-name")
#define VMARGS             _T_ECLIPSE("-vmargs")  /* special option processing required */
#define CP                 _T_ECLIPSE("-cp")
#define CLASSPATH          _T_ECLIPSE("-classpath")
#define JAR                _T_ECLIPSE("-jar")
 #define PROTECT            _T_ECLIPSE("-protect")
#define ROOT               _T_ECLIPSE("root")  /* the only level of protection we care now */

 #define OPENFILE           _T_ECLIPSE("--launcher.openFile")
 #define DEFAULTACTION      _T_ECLIPSE("--launcher.defaultAction")
 #define TIMEOUT            _T_ECLIPSE("--launcher.timeout")
 #define LIBRARY            _T_ECLIPSE("--launcher.library")
 #define SUPRESSERRORS      _T_ECLIPSE("--launcher.suppressErrors")
 #define INI                _T_ECLIPSE("--launcher.ini")
 #define APPEND_VMARGS      _T_ECLIPSE("--launcher.appendVmargs")
 #define OVERRIDE_VMARGS    _T_ECLIPSE("--launcher.overrideVmargs")
#define LAUNCHER_NORESTART _T_ECLIPSE("--launcher.noRestart")
 #define SECOND_THREAD      _T_ECLIPSE("--launcher.secondThread")
 #define PERM_GEN           _T_ECLIPSE("--launcher.XXMaxPermSize")
#define OLD_ARGS_START     _T_ECLIPSE("--launcher.oldUserArgsStart")
#define OLD_ARGS_END       _T_ECLIPSE("--launcher.oldUserArgsEnd")
#define SKIP_OLD_ARGS      _T_ECLIPSE("--launcher.skipOldUserArgs")

#define XXPERMGEN          _T_ECLIPSE("-XX:MaxPermSize=")
#define ADDMODULES         _T_ECLIPSE("--add-modules")
#define ACTION_OPENFILE    _T_ECLIPSE("openFile")
 #define GTK_VERSION        _T_ECLIPSE("--launcher.GTK_version")
```

It gets things from multiple lists, and none of the lists is completely present. So, I guess I also don't get the naming convention.

Your proposal does look weird though, as in `eclipse.c` there is a list of options for the command line help text, and none of them have `--launcher.xxx` syntax...
Are you sure the names you proposed are a good idea? I personally find the current names more logical.

> I have made a few comments below what I think needs to be changed.

That is incomplete, as it does not account for JavaDocs, tests, etc. I can make the necessary changes, but I'm not sure we should, since I don't see the logic of the new names you propose just yet.

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -1157,6 +1166,11 @@ static void getVMCommand( int launchMode, int argc, _TCHAR* argv[], _TCHAR **vmA
 	/* override or append vm args */
 	(*progArgv)[ dst++ ] = appendVmargs ? APPEND_VMARGS : OVERRIDE_VMARGS;
 
+	/* 'No restart' argument, if applicable. */
+	if (noRestart) {
+		(*progArgv)[ dst++ ] = LAUNCHER_NORESTART;
```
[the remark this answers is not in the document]
dhendriks: As I understand it, here we pass an argument from the launcher to the application.

HannesWell:
Thank you for your detailed answer, I have checked also the surrounding code in more detail.
> 
> The way I understood it, is: There are options for the launcher. They are use short names. When passed on to the application, they get the longer form to indicate it is an option from the launcher, as an application also has options of itself. Hence, in my version, the launcher got a `-norestart` and the Eclipse application got `--launcher.noRestart`. There are separate lists in `eclipseCommon.h` for this very reason, I thought. 

While it is correct that there are options that are only processed by the launcher (and not passed to the application) and options that are only passed to the application (and not processed by the launcher), I didn't find an option that is processed by the launcher and then passed to the app under a different name (maybe there are options that imply other options?). And I would also find it confusing as a user if I specify some option `-A` on the CLI and get another option `-B` passed into my application.
I think it should really be the same name. Which name that is, is something that can be discussed.

> 
> Your proposal does look weird though, as in `eclipse.c` there is a list of options for the command line help text, and none of them have `--launcher.xxx` syntax... Are you sure the names you proposed are a good idea? I personally find the current names more logical.
> 
The reasons I find `--launcher.norestart` more suitable is that is it similar to `--launcher.suppressErrors` and since both affect a similar area (i.e. the handling of exit codes) I think they should follow the same naming schema. Furthermore the _norestart_ flag is actually just processed by the launcher. It is just passed to the application for informational purpose and that we can disable the restart button.
It is similar to the `--launcher.library` argument, which is also passed through respectively the `-launcher` argument (even though that doesn't have the `--launcher.X` prefix again).
And because I assume that that the norestart flag is used very often I also don't think it is very important to make it short.

Why do you prefer just `-norestart`?

The doc gives the full list of supported options:
https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fruntime-options.html
The sources for it are in:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/3bd05fbd8610fa57766a0399b59ac256e5f50424/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html#L114-L117

Eventually we should update that as well.

> There are separate lists in eclipseCommon.h for this very reason, I thought.
> But, it was already not consistently applied,

Yes it looks like that. For example I cannot find a any definition nor a usage of the `-library` argument, although it is listed in the `Eclipse program launcher` argument list in the beginning of `eclipse.c`. Therefore I have the impression that list is outdated already and since it is only in this c file, I assume it is not 'normative' in any way.

> > I have made a few comments below what I think needs to be changed.
> 
> That is incomplete, as it does not account for JavaDocs, tests, etc. 

That's right. I have adapted to and it should succeed now (at least my local tests after a local rebuild did). I hope I didn't forget something.

dhendriks:
I changed `--launcher.norestart disables the restart behavior of exit codes 23 and 24.` to `--launcher.noRestart disables the restart behavior of exit codes 23 and 24.`, capitalizing the `R` of `noRestart`. With that, it is now consistently `--launcher.noRestart`.

dhendriks:
I updated the merge request title to account for now having `--launcher.noRestart`, rather than `-norestart`.

dhendriks:
> I think it should really be the same name. Which name that is, is something that can be discussed.

Alright. Lets go with only `--launcher.noRestart` then.

@HannesWell Probably another build needs to be approved. If that succeeds, can we merge this?

HannesWell (approved):
> > I think it should really be the same name. Which name that is, is something that can be discussed.
> 
> Alright. Lets go with only `--launcher.noRestart` then.

Great.

> I changed `--launcher.norestart disables the restart behavior of exit codes 23 and 24.` to `--launcher.noRestart disables the restart behavior of exit codes 23 and 24.`, capitalizing the `R` of `noRestart`. With that, it is now consistently `--launcher.noRestart`.

Yes, that makes sense. Thanks.
I have now squashed the commit that applies the suggestions from code review into your last commit and also adapted the commit message to the new name.

> @HannesWell Probably another build needs to be approved. If that succeeds, can we merge this?

Yes, absolutely. I'm fine with this now and I assume we won't get more reviews from other committers, but I think it's a sane enhancements now.

Thanks for this enhancement and your patience. Also thanks to @umairsair for your help!

HannesWell:
The test-failure is unrelated, submitting.

@dhendriks would you be so kind and also provide a PR to add documentation for this new flag next to
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/3bd05fbd8610fa57766a0399b59ac256e5f50424/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html#L114-L117

Which eventually will be part of
https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fruntime-options.html

dhendriks:
> Thanks for this enhancement and your patience. Also thanks to @umairsair for your help!

Indeed, thanks @umairsair. And thanks to you as well, @HannesWell. And to the others that participated/helped.

dhendriks:
> @dhendriks would you be so kind and also provide a PR to add documentation for this new flag [...]

I created a pull request for it: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2195

merks:
@dhendriks 

Thanks for your persistence and patience!  🏆

dhendriks:
I was looking at how to pick up the option in the IDE, and I'm not sure passing it as an argument is the right way to go. I can't figure out how to get them. Maybe it should be passed as a property, as @laeubi [suggested](https://github.com/eclipse-equinox/equinox/pull/595#issuecomment-2067593216)?

@HannesWell What do you think?

dhendriks:
Ah, they seem to already be available, as Java property `eclipse.commands`. I can use that.

dhendriks:
I've created a PR for the workbench restart not being available. See https://github.com/eclipse-platform/eclipse.platform.ui/pull/2125.

Issue #571, opened 2024-03-28
JRE-1.1.profile has no system packages


laeubi:
The system packages for the profile is empty:

https://github.com/eclipse-equinox/equinox/blob/cf85dc1c86b3df415610a30c981fa7ed2ffef6a1/bundles/org.eclipse.osgi/JRE-1.1.profile#L14

This currently leads to a warning in Tycho builds

> [09:31:29.975] [WARNING] No system packages found in profile nor toolchain for JRE-1.1, using current JRE system packages.

to reduce warnings in builds I'd like to get rid of this, the question is is it correct that for a JRE-1.1 VM there are literally no system packages or can be simply copy the ones from here for example:
https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi/OSGi_Minimum-1.1.profile

tjwatson:
I'm surprised there are any projects that try to use JRE-1.1, let alone fine a java compiler that can compile down to something that low.  I doubt it makes any difference what you put in there since I doubt it can really be used for a real scenario anymore.  With that said, I guess you can copy the content of https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi/OSGi_Minimum-1.0.profile for the system packages.

A while ago I wanted to delete all these profiles except for the ones related to Java 8 since these are the only ones the framework could ever use at runtime.  The Java 9 one is never used because we generate the system packages based on JPMS.  But it surprised me that we have all these users of these files.  At any rate, there are no profiles for Java versions > 9 so tycho and others must get this information some other way, right?

laeubi:
> I'm surprised there are any projects that try to use JRE-1.1, let alone find a java compiler that can compile down to something that low.

It is not really used in a build, it is just that Equinox mention it in the [profile.list](https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi/profile.list) and there it is parsed and issues an warning because it has no system packages. So another one would be to simply delete it altogether.

> At any rate, there are no profiles for Java versions > 9 so tycho and others must get this information some other way, right?

For any JVM > 8 that does not have a profile Tycho automatically generates one.

tjwatson:
> So another one would be to simply delete it altogether.

+1 to delete.

HannesWell:
> At any rate, there are no profiles for Java versions > 9 so tycho and others must get this information some other way, right?

[...]

PR #596, opened 2024-04-15
Revert removal of JRE-1.1 execution environment profile and add its system-packages

HannesWell:
This reverts https://github.com/eclipse-equinox/equinox/pull/572 and instead instead adds the system-packages for the JRE-1.1 execution environment profile that where derived from ancient Java-1.1 JREs as described in https://github.com/eclipse-pde/eclipse.pde/pull/693#issuecomment-1673903650.

Fixes https://github.com/eclipse-equinox/equinox/issues/571
Part of https://github.com/eclipse-pde/eclipse.pde/issues/1231 (fixing that completely requires the reversion of https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/425)

Once this is merged, we can remove the hard-coded system-packages from PDE:
https://github.com/eclipse-pde/eclipse.pde/blob/860100e5b0d57bca42cbe26078ea5c3093828c60/build/org.eclipse.pde.build/src/org/eclipse/pde/internal/build/site/PDEState.java#L580-L603

### DIFF
--- a/bundles/org.eclipse.osgi/JRE-1.1.profile
+++ b/bundles/org.eclipse.osgi/JRE-1.1.profile
@@ -1,0 +1,27 @@
+###############################################################################
+# Copyright (c) 2003, 2012 IBM Corporation and others.
+#
+# This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License 2.0
+# which accompanies this distribution, and is available at
+# https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+# 
+# Contributors:
+#     IBM Corporation - initial API and implementation
+###############################################################################
+org.osgi.framework.system.packages = 
+org.osgi.framework.bootdelegation = \
+ sun.*,\
+ com.sun.*
+org.osgi.framework.executionenvironment = \
+ JRE-1.1
+org.osgi.framework.system.capabilities = \
+ osgi.ee; osgi.ee="JRE"; version:List<Version>="1.0, 1.1"
+osgi.java.profile.name = JRE-1.1
+org.eclipse.jdt.core.compiler.compliance=1.3
+org.eclipse.jdt.core.compiler.source=1.3
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.1
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=ignore
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=ignore
--- a/bundles/org.eclipse.osgi/JRE-1.1.profile
+++ b/bundles/org.eclipse.osgi/JRE-1.1.profile
@@ -11,7 +11,31 @@
 # Contributors:
 #     IBM Corporation - initial API and implementation
 ###############################################################################
-org.osgi.framework.system.packages = 
+org.osgi.framework.system.packages = \
+java.applet,\
+java.awt,\
+java.awt.datatransfer,\
+java.awt.event,\
+java.awt.image,\
+java.awt.peer,\
+java.beans,\
+java.io,\
+java.lang,\
+java.lang.reflect,\
+java.math,\
+java.net,\
+java.rmi,\
+java.rmi.dgc,\
+java.rmi.registry,\
+java.rmi.server,\
+java.security,\
+java.security.acl,\
+java.security.interfaces,\
+java.sql,\
+java.text,\
+java.text.resources,\
+java.util,\
+java.util.zip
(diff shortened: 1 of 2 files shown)


tjwatson (approved):
In the end adding the JRE-1.1 is the simple thing to do.  But I really wish PDE and others could somehow just use the "next best" EE when the one requested isn't available.  After all the JavaSE-1.2 profile claims to provide:

```
 osgi.ee; osgi.ee="JRE"; version:List<Version>="1.0, 1.1"
```

HannesWell:
> In the end adding the JRE-1.1 is the simple thing to do. But I really wish PDE and others could somehow just use the "next best" EE when the one requested isn't available. 

It would probably be possible somehow. But for now I agree that reverting is the simplest thing to do.
If someone has the time and interest to investigate it in depth and all details it can be removed again, which would be great.

laeubi (commented):
Thanks for fixing

merks:
I’m not asking for sympathy. If you want some background about one specific case, this is one such case

https://github.com/eclipse-pde/eclipse.pde/issues/1231#issuecomment-2056841517

tjwatson:
> I’m not asking for sympathy. If you want some background about one specific case, this is one such case
> 
> [eclipse-pde/eclipse.pde#1231 (comment)](https://github.com/eclipse-pde/eclipse.pde/issues/1231#issuecomment-2056841517)

I saw that, but I do not understand why we need to keep building that particular bundle if it will never have real source code changes.

merks:
I could change it and will do that if/when the signature expires.  There has never been any suggestion that something was wrong, bad, or otherwise.  Also, projects depend on really old maven crap and those might contain such things or when wrapped by bnd might produce such things.  I’d rather such old crap were not used. It would make my life easier.

laeubi:
@tjwatson that's why I initially though it is "safe" to remove but in the end its just a random text file that literally never changes so any more effort here seems wasted time compared to the rumors this has already produced.

HannesWell:
> One thing that concerns me with the \"support\" objectives of PDE and other tools is I don't think there is a possibility to use Java 21 to compile down to Java 1.1. In fact, I would not be surprised if Java 11 doesn't let you compile to Java 1.1. Even ECJ will be dropping the ability to compile to really old Java versions.

At least what I can say from the top of my head in PDE the main 'issue' is that the EEs provided in the PDE TP state and their system packages are the ExecutionEnvironment defined by JDT, which get the system-packages from the embedded Equinox profiles.
If JDT removes some of the EEs in the future that mechanism has to be changed in PDE to be the JDT-EEs plus the ancient EE that are still required in some old existing manifests.
We then probably have hard-code the system-packages in PDE (again).

So it is possible just (probably) needs more planning/investigations.
But since only bundles in the TP can then require the removed EEs it is probably save to just reuse the system-packages of the next supported EE and only maintain a map of EE name to Java-Version.

laeubi:
I already mentioned it in the past that I think the behavior of PDE is wrong in this regard. It is matching each (target) bundle against the exact matching profile this does not happens in a runtime and therefore produces all this kind of strange errors sometimes. Instead PDE should use the target platforms EE as this is what is "targeting" and will be used at runtime, especially for bundles from the target platform the user can't influence the EE anyways in any other way.

merks:
It seems to me that each project is bound to a JRE too and when it doubt, that would be used instead.  

Also, looking closely here, I'm surprised that the information jar produced by Tycho is not really what I would expect based on the BREE in the MANIFEST.MF:

image

In any case, note that the source MANIFEST.MF makes no mention on JRE 1.1.  That detail is produced elsewhere.

HannesWell:
> I already mentioned it in the past that I think the behavior of PDE is wrong in this regard. It is matching each (target) bundle against the exact matching profile this does not happens in a runtime and therefore produces all this kind of strange errors sometimes. Instead PDE should use the target platforms EE as this is what is "targeting" and will be used at runtime, especially for bundles from the target platform the user can't influence the EE anyways in any other way.

The fundamental problem here is that a workspace is not the runtime and the eventual runtime JVM is not known with certainty. Furthermore bundles in the TP are different from the one in the workspace. But no solution for that problem and there are always possibilities for false positive or negative errors. But we have discussed that in length in https://github.com/eclipse-pde/eclipse.pde/issues/429 and I actually don't want to start that discussion again. The solution chosen there is in my onion a good trade-off.

Nevertheless this issue in combination of the discussed removal of support for ancient java version made we wonder if these EEs should be removed from JDT, which would in the end also allow to remove the profiles from Equinox when PDE is adapted:
- https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/433

PR #605, opened 2024-05-02
[Build] Ensure GH-workflow for MacOS runs on x86_64 image

HannesWell:
GitHub Actions changed their 'latest' images for MacOS to use the 'aarch64' instead of the 'x86_64' architecture:
https://github.com/actions/runner-images/blob/7bb1d84f7071bfa9c350d7552d9631d9a69bfdb0/README.md

But for macosx.aarch64 no JDK-8 is available.

This is similar to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1201.


HannesWell:
Mac build is green again.

HannesWell:
As a back-reference: This is actually not a good solution, see https://github.com/eclipse-equinox/equinox/pull/614.

PR #614, opened 2024-05-07
Remove macos-latest-large label

fredg02:
See https://www.eclipse.org/lists/cross-project-issues-dev/msg19891.html for context


fredg02:
FYI: @HannesWell @laeubi

HannesWell:
> See https://www.eclipse.org/lists/cross-project-issues-dev/msg19891.html for context

Thank you Fred. I wasn't aware that these images are not free.
The only reason to use them was that java-8 is not available on the new arm nodes.

PR #618, opened 2024-05-07
[Build] Only archive just built jars and launcher binaries in Jenkins

HannesWell:
Currently all jars in the workspace of a Jenkins build run are archived, which also includes the entire local Maven repository and all baseline jars. This leads to a total archive size of about 670MB (compressed).

With this only the built jars plus the native binaries are archived, leading to a much smaller storage demands and a shorter runtime of the archive step.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -314,7 +314,7 @@ pipeline {
 			}
 			post {
 				always {
-					archiveArtifacts artifacts: '**/*.log, **/*.jar', allowEmptyArchive: true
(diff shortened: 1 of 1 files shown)


HannesWell:
> Currently all jars in the workspace of a Jenkins build run are archived, which also includes the entire local Maven repository and all baseline jars. This leads to a total archive size of about 670MB (compressed).

The (compressed) archive is now down to 29MB and additionally contains the launcher binaries. Furthermore the runtime of the archiving step went down from almost 1min30s to ~5s.

PR #619, opened 2024-05-07
[Build] Speed-up recording of compiler- and test-logs

HannesWell:
Make the log-file patterns more specific to speed up the search. This is possible since the equinox and equinox.binaries repository are checkout side-by-side and avoids search the local Maven repository in the workspace.


HannesWell:
This does not have a significant impact on the runtime of the log-recordings, thus keeping the pipeline as it is.

Issue #630, opened 2024-05-14
[Mac Sonoma] "WARNING: Secure coding is not enabled for restorable state" warning message


Phillipus:
**The issue**
When running Eclipse or an RCP app on macOS Sonoma a warning message is written to the console:

`eclipse[4647:170325] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.`

I've Googled this and it seems that this could lead to a vulnerabilty and needs to be addressed. For example see here:

https://stackoverflow.com/questions/77283578/sonoma-and-nsapplicationdelegate-applicationsupportssecurerestorablestate?rq=2

**To Reproduce**
The only way to see this warning message is to run Eclipse on Mac Sonoma from the command line so the message is written to the console in the Terminal app:

1. Open the Mac Terminal app and run the following command:
`Eclipse.app/Contents/MacOS/eclipse`
2. The warning message is written to the terminal window

**Expected behavior**
The warning message should not appear.

**Environment:**
macOS Sonoma 14.5
MacBook M2 aarch64
Temurin JDK 17.0.11+9
Eclipse 4.31 and latest 4.32 I builds

Phillipus:
The warning message appears at this point:

https://github.com/eclipse-equinox/equinox/blob/c4be7f2d1bfab28d171e7ead3a95632a924505cf/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java#L2175

And so, if it is possible to implement this, it will be in the native JNI code here:

https://github.com/eclipse-equinox/equinox/tree/master/features/org.eclipse.equinox.executable.feature/library/cocoa

Phillipus:
I first reported this issue in the SWT issue tracker, and then closed that one and opened it here because the warning message appearing at the `bridge.showSplash` point. But after debugging with the `-noSplash` option the warning message now appears immediately after this line in `WorkbenchAdvisor`:

https://github.com/eclipse-platform/eclipse.platform.ui/blob/deb75d4e977d767d3c5249e8a953fd2ef986baab/bundles/org.eclipse.ui.workbench/Eclipse%20UI/org/eclipse/ui/application/WorkbenchAdvisor.java#L768

So I'm not sure whether to continue to trace this to SWT or Equinox. @HannesWell @HeikoKlare WDYT?

mdaloia:
[...]

PR #632, opened 2024-05-15
Fix warning message on Mac about secure coding not enabled

Phillipus:
- See https://github.com/eclipse-equinox/equinox/issues/630

- On macOS 14 and later a warning message is written to console: "WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES."

- This adds an NSApplicationDelegate that implements applicationSupportsSecureRestorableState

- This is a draft PR to show what I think needs to change to fix this issue. However, I don't know if the additional code could be better placed. Feedback welcome!


HannesWell:
@sravanlakkimsetti can you review this? 

@Phillipus thank you for being pro-active on this topic. You can try out the executables/launcher-libraries archived in the Jenkins verification build to test if it has the desired effect.

Phillipus:
> @Phillipus thank you for being pro-active on this topic. You can try out the executables/launcher-libraries archived in the Jenkins verification build to test if it has the desired effect.

@HannesWell Thanks for the heads up. I had already compiled and built the `so` file locally and tested it. But I also tested the one you refer to at Jenkins [here](https://ci.eclipse.org/releng/job/equinox/job/PR-632/2/artifact/equinox.binaries/org.eclipse.equinox.launcher.cocoa.macosx.aarch64/). That does indeed have the desired effect.

Phillipus:
I'm not certain that setting an app delegate where I have set it in the PR is the right place.  There are places that reference a an existing delegate:

https://github.com/eclipse-equinox/equinox/blob/c4be7f2d1bfab28d171e7ead3a95632a924505cf/features/org.eclipse.equinox.executable.feature/library/cocoa/eclipseCocoa.c#L197-L198

https://github.com/eclipse-equinox/equinox/blob/c4be7f2d1bfab28d171e7ead3a95632a924505cf/features/org.eclipse.equinox.executable.feature/library/cocoa/eclipseCocoa.c#L211-L212

https://github.com/eclipse-equinox/equinox/blob/c4be7f2d1bfab28d171e7ead3a95632a924505cf/features/org.eclipse.equinox.executable.feature/library/cocoa/eclipseCocoa.c#L234-L235

I'm not sure if setting a new delegate where I have might over-write a possibly existing one and then those functions might not work.

Phillipus:
I did a check, and at the point where this PR sets the AppDelegate it is `NULL` but I'm not sure where or if an AppDelegate is set somewhere else. @lshanmug can you advise?

sravanlakkimsetti:
@subyssurendran666 Can you please take a look?

Lets merge this after 4.32

Phillipus:
Don't merge this one without further investigation. The equivalent implementation in SWT caused a crash to desktop. See https://github.com/eclipse-platform/eclipse.platform.swt/pull/1240

PR #639, opened 2024-06-09
[Build] Derive TARGET_ARCH from -arch argument in make_win64.mak

HannesWell:
This allows cross-compilation on Windows between x86_64 and aarch64 by specifying the `-arch` argument as usual when calling the win32/build.bat script.

@chirontt do you see any problem with this?

### DIFF
--- a/bundles/org.eclipse.equinox.launcher.win32.win32.aarch64/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.launcher.win32.win32.aarch64/META-INF/MANIFEST.MF
@@ -3,8 +3,8 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.equinox.launcher.win32.win32.aarch64;singleton:=true
-Bundle-Version: 1.2.1000.qualifier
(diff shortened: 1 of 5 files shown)


chirontt:
Sounds good to me to re-use the `-arch` argument for cross-compiling purpose. This removes the need to set a TARGET_ARCH environment variable for cross-compiling, as originally intended.

But cross-compiling also requires a `target JDK` for the relevant headers and/or libraries suitable for the target architecture. Can the `-java` argument be used for this purpose?

HannesWell:
> Sounds good to me to re-use the `-arch` argument for cross-compiling purpose. This removes the need to set a TARGET_ARCH environment variable for cross-compiling, as originally intended.

Exactly. And it also verifies that the jdk passed with the -java argument exists.

> 
> But cross-compiling also requires a `target JDK` for the relevant headers and/or libraries suitable for the target architecture. Can the `-java` argument be used for this purpose?

That's also right and yes the -java argument can be used for that.
I'm currently enhancing the documentation about how to build the Equinox binaries in https://github.com/eclipse-equinox/equinox/pull/640 and mentioned in there as an example.

HannesWell:
Build is green now (besides the usual TCK failure and an occasional DS test failure).

PR #642, opened 2024-06-10
Update org.apache.felix.scr to 2.2.12

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2115


merks:
FYI the first build failed with just the qualifier change, so I incremented the feature version.

PR #646, opened 2024-06-20
Fix BasicLocation set method when read-only and lock requested

### LINKED ISSUE #460: BasicLocationTests#testCreateLocation04() fails silently
The test case `org.eclipse.osgi.tests.services.datalocation.BasicLocationTests#testCreateLocation04()` fails silently. The test runs into a `fail` statement, which is caught and ignored by a surrounding try-catch:

https://github.com/eclipse-equinox/equinox/blob/60a2e63e9c40fc480c90b3800d80b7c03b0602b9/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/services/datalocation/BasicLocationTests.java#L149-L155

I tried to replace this block with an `assertThrows` in #455, which fails because nothing is actually thrown.

I have not yet investigated whether the test or the tested functionality is faulty.

tjwatson:
https://github.com/eclipse-equinox/equinox.framework/pull/39 PR introduced a breaking change that always returned true with a read-only location and lock was requested to the set method.

The testCreateLocation04 method also needed fixing to properly test this case.

Fixes #460

### DIFF
--- a/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Core OSGi Tests
 Bundle-SymbolicName: org.eclipse.osgi.tests;singleton:=true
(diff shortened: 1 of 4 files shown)


laeubi (approved):
looks fine to me, the control flow is really hard to follow :-\

PR #647, opened 2024-06-20
Remove unnecessary casts

stephan-herrmann:
When https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2471 is merged, ecj will signal more casts as unnecessary.

With this PR I suggest to remove affected casts before new warnings will show up in the build.

### DIFF
--- a/bundles/org.eclipse.equinox.app/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.app/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.app; singleton:=true
-Bundle-Version: 1.7.100.qualifier
+Bundle-Version: 1.7.200.qualifier
 Bundle-Vendor: %providerName
 Bundle-Activator: org.eclipse.equinox.internal.app.Activator
 Bundle-Localization: plugin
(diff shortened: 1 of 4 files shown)


akurtakov:
Version bump needed:
 Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.equinox.app: Only qualifier changed for (org.eclipse.equinox.app/1.7.100.v20240620-1529). Expected to have bigger x.y.z than what is available in baseline (1.7.100.v20240321-1445) -> [Help 1]

stephan-herrmann:
` testCoordinatedConfigurationOnBeforeRegisteredManagedService (org.osgi.test.cases.cm.junit.CMCoordinationTestCase) failed`

Sorry, I can't find that test so I cannot test if this failure could possibly be related.

tjwatson:
> Sorry, I can't find that test so I cannot test if this failure could possibly be related.

It is not related.

stephan-herrmann:
> > Sorry, I can't find that test so I cannot test if this failure could possibly be related.
> 
> It is not related.

thanks.

merks:
I think it’s best for there to be one commit not 3. 

I think I can do that with squash and merge button, but I’ve not tried that before.

merks:
That worked okay.

stephan-herrmann:
> I think I can do that with squash and merge button, but I’ve not tried that before.

There's a first time for everything :)

(On other occasions people wanted to have version bumps separate from code changes, but perhaps that's only relevant in projects with BETA_JAVAXY branches).

merks:
Yes sometimes people want that.  I think to make revert easier. But we don’t expect to revert this and the preceding state would never build successfully. 😬

merks:
And probably two separate version bump commits never make sense. 😱

Personally I generally amend and force push until I’m done. No one has complained so far. 😜

tjwatson:
> Yes sometimes people want that. I think to make revert easier. But we don’t expect to revert this and the preceding state would never build successfully.

Yes, that was the reason for the policy.  Generally, it caused problems if we reverted and one I-Build to the next we had a version decrease.  The other issue was if other changes happened after then the version decrease would become an issue also. But a straight revert in this project seems pretty rare vs. a new PR that fixes the issue on top of the existing changes.

merks:
Nothing I commit ever needs to be reverted. 👼 😜

Issue #652, opened 2024-07-02
@user.home can be used in -eclipse.keyring argument but not with -eclipse.password.


mPorhel:
-eclipse.keyring and -eclipse.password runtime options can be used to change the location of the secure storage and bypass the password provider mecanism.

They are handled in org.eclipse.equinox.internal.security.storage.SecurePreferencesMapper::open.
`@user.home` is replaced with the corresponding value.path during retrieval of the secure storage / keyring file in  content in  SecurePreferencesMapper::getKeyringFile but not in password file retrieval in SecurePreferencesMapper::processPassword

The documentation of those options clearly indicates that protection of the password file must be considered. Having the possibility to use `@user.home`can help to avoid absolute path or computation from user name on the command line if user profile folder is used to store this file (or as indirect mounting point).

@https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Freference%2Fref-securestorage-options.htm

> Password file
The password provider mechanism can be bypassed if you specify the "-eclipse.password <file path>" runtime option. In this case the contents of the file you specified as the argument will be used as a master password. While this option is valuable in some circumstances (such as headless applications), the protection of the password file becomes a consideration. The password file can be protected by the operating system access rights and/or by placing it on a removable storage, such as a USB key.

HannesWell:
Fixed with https://github.com/eclipse-equinox/equinox/pull/653

Thank you @mPorhel for providing a fix. Much appreciated.

PR #653, opened 2024-07-02
#652 Handle @user.home in -eclipse.password file retrieval

mPorhel:


bundles/org.eclipse.equinox.security/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.security;singleton:=true
-Bundle-Version: 1.4.300.qualifier
+Bundle-Version: 1.4.301.qualifier
```
merks, line 5: We generally increment by +100 in the master stream so that there is room for increments in a maintenance stream that don't overlap with the  master stream's version.

https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md

merks (changes requested):
Other than the version, it looks fine.

bundles/org.eclipse.equinox.security/META-INF/MANIFEST.MF
mPorhel: Thanks for the pointer Ed.

merks (approved):

PR #654, opened 2024-07-05
Migrate ConfigurationSessionTests in configuration package to JUnit 5

HeikoKlare:
The test classes `EclipseStarterConfigurationAreaTest`, `MovableConfigurationAreaTest`, and `ReadOnlyConfigurationAreaTest` rely on the JUnit-3-specific `ConfigurationSessionTest`. The `SessionTestExtension` has been introduced as a JUnit-5-based replacement. This change adapts the test classes to be executed with JUnit 5, using the `SessionTestExtension`.

This contribution serves two purposes:
1. Modernization of the Equinox tests
2. Modernization of the overall Eclipse Platform tests and getting rid of JUnit-3-based functionality (such as the `ConfigurationSessionTest`), see https://github.com/eclipse-platform/eclipse.platform/issues/903

I will also provide PRs for updating the other `ConfigurationSessionTest` usages in Equinox once this update for the test classes in the `configuration` package has been approved.


HannesWell:
Thanks Heiko for this.
I'll try to have a detailed look as soon as possible (maybe already tomorrow).
Just one first question: Is a deterministic order of execution for the test cases annotated with @Order strictly necessary or is this just nice to have? In case it's the latter I would suggest to remove that.

HeikoKlare:
Thank you, Hannes!
In short: yes, the order annotations are (usually) necessary. The existing `ConfigurationSessionTest`/`ConfigurationSessionTestSuite` executes test methods ordered by their names by default. First, I find this slightly confusing, as you have to know about how that test suite behaves to understand and define a proper execution order. Second, the new JUnit 5 `SessionTestExtension` only hooks into the execution of a single test case rather than orchestrating the whole test class, which makes it impossible to enforce an execution order. There would, of course, have been other means than a JUnit 5 extension to realize session tests, but I found the current solution of having the configurable `SessionTestExtension` together with the possibility to define for each test case whether it is executed in a separate session (default) or in the host (via `@0ExecuteInHost`) to be most appropriate.
As an additional information: I have already adapted all the Equinox session tests while introducing the `SessionTestExtension` to the Eclipse Platform to be quite sure that it will work properly. I just wanted to start with a smaller set of tests in this PR as a starting point for potential initial discussions on the solution.

akurtakov:
Is there plan to finish this one?

HeikoKlare:
> Is there plan to finish this one?

It is basically finished, it just waited for review 🙂 

I can rebase the change if someone wants to have a look and eventually merge it. I can then also provide the follow-up PRs for the remaining `ConfigurationSessionTests` in Equinox, as we can then get rid of that class in the Eclipse Platform.

akurtakov:
Please rebase and I'll review/merge. Sorry for the delays, this one I just missed.

PR #656, opened 2024-07-12
[Build] Define k8-agent for linux.x86_64 natives completely in pipeline

HannesWell:
This avoids the need to use customized docker images maintained in `eclipse.platform.releng.aggregator`.

This is similar to https://github.com/eclipse-platform/eclipse.platform.swt/pull/1332.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,12 +17,18 @@ def runOnNativeBuildAgent(String platform, Closure body) {
 	def final nativeBuildStageName = 'Perform native launcher build'
 	if (platform == 'gtk.linux.x86_64') {
 		podTemplate(inheritFrom: 'centos-latest' /* inhert general configuration */, containers: [
-			containerTemplate(name: 'launcherbuild', image: 'eclipse/platformreleng-centos-swt-build:8',
+			containerTemplate(name: 'jnlp', image: 'eclipsecbi/jiro-agent-centos-8',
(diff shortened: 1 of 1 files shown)


HannesWell:
Looks like we can't install the GTK development packages using `yum -y install gtk3-devel` since it requires elevated permissions.
Using `sudo yum -y install gtk3-devel` also does not work since it says `sudo` is unknown. Furthermore the specifying the container with `privileged: true` is also not permitted.

I also thought about levering [nested-pod-templates](https://plugins.jenkins.io/kubernetes/#plugin-content-nesting-pod-templates) but I didn't found a docker image for GTK development (on centos), that is not one of the ones we build our-self in releng.aggregator (and I actually would like to get rid of).

@akurtakov any idea how to solve this?

In the end I would like to define in something in this repo, ideally in the Jenkins pipeline, what needs to be installed in the container on top of the `eclipsecbi/jiro-agent-centos-8` image.

HannesWell:
Can something like this be a solution?
https://stackoverflow.com/questions/36651091/how-to-install-packages-in-linux-centos-without-root-user-with-automatic-depen

HannesWell:
The idea of this does not work because docker-images are 'immutable' and one cannot simply install extra software.
So I'm closing this one.

The simplification of the definition was extracted into https://github.com/eclipse-equinox/equinox/pull/671.

PR #657, opened 2024-07-21
[Win32] Opt-in for long path capability in Eclipse application manifest

HannesWell:
This allows an Eclipse application to avoid the path length limitation on Windows if additionally the limit disabled in the registry.

See
- https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#application-manifest-updates-to-declare-long-path-capability
- https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longPathAware

### DIFF
--- a/features/org.eclipse.equinox.executable.feature/library/win32/eclipse.exe.manifest
+++ b/features/org.eclipse.equinox.executable.feature/library/win32/eclipse.exe.manifest
@@ -8,8 +8,9 @@
 		</dependentAssembly>
 	</dependency>
 	<asmv3:application>
-		<asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
-			<ms_windowsSettings:dpiAware xmlns:ms_windowsSettings="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</ms_windowsSettings:dpiAware>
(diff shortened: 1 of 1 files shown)


HannesWell:
@sravanlakkimsetti or @niraj-modi can you tell if that change makes sense?

The goal is to fix the failures of PDE-tests in I-builds (see https://github.com/eclipse-pde/eclipse.pde/issues/1333) since the paths on the test machines got longer and now exceed the default limit.
The test machine has already opt-ed in for long paths (https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4832) but the application has to opt-in too. I hope that this is the thing to do.

vogella:
Great, ran into this multiple times with clients and Oomph IIRC.

Once this is in, please add to N&N with the information which registry key to set

HannesWell:
After some more research I found [JDK Bug 8315405](https://bugs.openjdk.org/browse/JDK-8315405) which states that there is no perfect way to circumvent the MAX_PATH length limit for the working directory of launched processes on Windows.
This is probably also the reason why the JDK itself does not declar long-path compatibility in its executable manifest:
https://github.com/openjdk/jdk/blob/d63d6e23d123e82cef1372aed31b5846fb6d23fd/src/java.base/windows/native/launcher/java.manifest#L39-L43

But with https://github.com/eclipse-platform/eclipse.platform/pull/1477 I have created a work-around based on shortened MS-DOS naming convention. At least this should permit longer paths than currently possible. It is most effective if a path consists of only few long segments.

This can be closed.

PR #661, opened 2024-08-09
Leverage OSGi-DS component-xml file generation of PDE-DS/Tycho

HannesWell:
Remove all generated xml-files from git.

Especially the component-xml files are annoying since the state in git is not up to date. Consequently they are changed on each build and make the git staging dirty.
See https://gitlab.eclipse.org/eclipse-wg/ide-wg/community/-/issues/47#note_2617895


laeubi:
The `EventAdmin` is currently not generated xml maybe better place that in a separate PR?

HannesWell:
> The `EventAdmin` is currently not generated xml maybe better place that in a separate PR?

Yes, see https://github.com/eclipse-equinox/equinox/pull/662.

Although the test failure was in equinox.ds.tests.

HannesWell:
I really don't understand why the test is failing.
I could reproduce it locally when the component XML files are missing, but the logs say they are generated before the tests are executed?

bundles/org.eclipse.equinox.ds.tests/.gitignore
```diff
@@ -1 +1,2 @@
 /scr_test
+OSGI-INF/org.eclipse.equinox.*.xml
```
laeubi, line 2:
```suggestion
OSGI-INF/
```

laeubi:
I assume the problem is that the bundle is not packed before executed as a test and then the launcher looks for `OSGI-INF` (that is empty) but the files are generated at `target/classes/OSGI-INF`

HannesWell:
> I assume the problem is that the bundle is not packed before executed as a test and then the launcher looks for `${basedir}/OSGI-INF` (that is empty) but the files are generated at `${basedir}/target/classes/OSGI-INF`

Yes. :/
I now remembered the work-around I implemented in eclipse-platform until this works out of the box:
https://github.com/eclipse-platform/eclipse.platform/pull/790/files#diff-899b68f39b8bdb3938d5d7807842a815c76276df4e51e7bad26f8272f55994bb

bundles/org.eclipse.equinox.ds.tests/.gitignore
HannesWell: I reduced it now to `OSGI-INF/org.eclipse.*.xml`, ignoring all XML could be too restrictive. And since in the fast majority of cases the classes in eclipse projects are in a `org.eclipse...` package this should be fine. And if not, one will notice it quickly and can adjust the pattern in that specific case.

HannesWell:
Looks like this finally works. 🎉

PR #662, opened 2024-08-11
Generate OSGi-DS component-xml file for equinox.event via PDE-DS/Tycho

HannesWell:


HannesWell:
This is actually the simpler case of https://github.com/eclipse-equinox/equinox/pull/661.
Keep this in https://github.com/eclipse-equinox/equinox/pull/661, no need to artificially increase the number of commits.

PR #663, opened 2024-08-27
EclipsePreferences: use simple HashMap

jukzi:
Using a concurrent Map is a waste as every access is guarded by childAndPropertyLock anyway.

+ add missing guards (was used unguarded in RootPreferences)
+ make getChild(String key, Object context, boolean create) atomic


jukzi:
If there is no objection i plan to submit this tomorrow as is - ignoring the failing tck test.

PR #671, opened 2024-09-14
[Build] Simplify native build agent definition and platform list

HannesWell:


Jenkinsfile
```diff
@@ -16,33 +16,21 @@
...
-  volumes:
-  - name: tools
-    persistentVolumeClaim:
-      claimName: tools-claim-jiro-releng
-  - name: volume-known-hosts
-    configMap:
-      name: known-hosts
-''') { node(POD_LABEL) { stage(nativeBuildStageName) { container('launcherbuild') { body() } } } }
+		podTemplate(inheritFrom: 'centos-latest' /* inhert general configuration */, containers: [
+			containerTemplate(name: 'launcherbuild', image: 'eclipse/platformreleng-centos-swt-build:8',
```
akurtakov, line 20: As centos 8 is deprecated and for gtk4 we are looking into moving to newer version it would be better to try with eclipse/platformreleng-centos-swt-build:9.

HannesWell: I would like to do that in a separate PR dedicated only to that especially since using centos-9 has the non-trivial implications.

Btw. `inheritFrom: 'centos-latest'` to inherit the very basic configuration like volumes, credentials etc. The image used is the one specified in the line below.

PR #669, opened 2024-09-10
Add support for Linux on riscv64

yuzibo:

PR #672, opened 2024-09-14
Version bumps for 4.34 stream

yuzibo:


yuzibo:
Adding the PR to the [commit](https://github.com/eclipse-equinox/equinox/pull/669/commits/a84cdd4a7bed2c5c23f42e43d2269570118410b0), so close it.

PR #676, opened 2024-09-21
Update tycho-build to 4.0.9

akurtakov:


merks:
FYI, this looks like the problem I saw yesterday where the API baseline check appears to hang along with the new warning:

image

PR #677, opened 2024-09-25
[Build] Replace use of coactions/setup-xvfb

akurtakov:
It will stop working on Oct 15 and is claimed to not be needed anymore in https://github.com/coactions/setup-xvfb/issues/29#issuecomment-2374414522

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,8 +11,6 @@ on:
 
 jobs:
   build:
-    name: Build ${{ matrix.config.name }}
-    runs-on: ${{ matrix.config.runner-os }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


HannesWell:
@akurtakov I just pushed an update that should fix the failures and also restores the previous step order. Not that the order is important in that case, but I think we should minimize changes.

HannesWell:
Something is still wrong with Windows.
Can have a look tomorrow.

HannesWell:
> Something is still wrong with Windows.

The problem seems to be that powershell(-core), which is the default shell on Windows runners seems to interpret dots in variables differently. To solve this we would either have to escape all arguments that contain a dot or use cmd as shell.
I went now for the latter.

HannesWell:
Finally all builds have been successful without the `coactions/setup-xvfb` action. 🎉

PR #684, opened 2024-10-03
Do not expose the ServicePermission to users of URL

tjwatson:
An important doPriv was removed which protected callers of URL from the SerivcePermissin check to get the handler service.


tjwatson:
This is to fix the changes from #290 that removed the doPriv for getting the handler service.

PR #689, opened 2024-10-12
[Oomph-Setup] Change Equinox configuration setup to contain only equinox

HannesWell:
Add a `equinox.core` configuration setup that only contains the `Equinox` and `Equinox Binaries` repository, opposed to the 'full' equinox configuration that also contains P2.

I used Equinox Core to refer to this repository to resolve the name clash that exists between this repository and the overall organization name. Alternatively we could find another name for the configuration handling the 'full' Equinox organization and rename it. Actually I would be in favor of that but didn't found a good name for it, `Equinox Full`, `Equinox Organization`, `Equinox SDK`, `Equinox Complete`, `Equinox All`, all didn't sound like a perfect solution for me.

@laeubi, @merks do you have any suggestions or opinions on that?

This also adds a reference the new `equinox.p2` configuration added in the eclipse-equinox/p2 repository.
Additionally change the styled and drag&drop-able Oomph Configuration button in the main README to just display the Equinox Core configuration, not the 'full' Equinox organization config.

The 'full' equinox configuration is also added to the organizations overview page via (currently it's only shown in the CONTRIBUTING file for Equinox
- https://github.com/eclipse-equinox/.github/pull/2

Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2430

### DIFF
--- a/releng/org.eclipse.equinox.releng/EquinoxConfiguration.setup
+++ b/releng/org.eclipse.equinox.releng/EquinoxConfiguration.setup
@@ -25,7 +25,7 @@
         value="equinox"/>
     <productVersion
         href="index:/org.eclipse.setup#//@productCatalogs[name='org.eclipse.applications']/@products[name='eclipse.platform.sdk']/@versions[name='latest']"/>
-    <description>The Equinox installation provides the latest tools needed to work with the Eclipse Equinox's source code.</description>
+    <description>The Equinox installation provides the latest tools needed to work with the project's source code.</description>
   </installation>
   <workspace
       name="equinox.workspace"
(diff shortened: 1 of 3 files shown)


merks:
Perhaps `equinox+p2` captures the name of the content without trying to describe or characterize the content...

HannesWell:
> Perhaps `equinox+p2` captures the name of the content without trying to describe or characterize the content...

This could work for the Badge/Button, but I assume one cannot use the + in variable names like for the workspace or installation, can you? At the same time it seems to be a bit specific to me to list the project.
But since the Equinox (repo) and P2 are actually not so strongly related another option could be that we simply drop the organization wide configuration and just have one for `Equinox` and `P2`? Respectively remove P2 from the Equinox configuration.

Then we just have a Equinox config (with the equinox and equinox.binaries repository) and a P2 configuration (containing only the p2 repo). On the organization README and CONTRIBUTING file we could then simply list both configs.
Having a config with Equinox+P2 does not seem to be so useful as it can be e.g. for all of JDT or for the Platform SDK (if one really wants to have everything).
WDYT?

releng/org.eclipse.equinox.releng/EquinoxConfiguration.setup
```diff
@@ -61,29 +61,24 @@
             value="Equinox"/>
       </setupTask>
     </setupTask>
-    <setupTask
-        xsi:type="setup:VariableTask"
-        name="eclipse.git.authentication.style"
-        defaultValue="anonymous"/>
```
HannesWell, line 67: @merks is it necessary to define this variable? As far as I can tell, the default is already `anonymous` anyways:

HannesWell (commented):
> But since the Equinox (repo) and P2 are actually not so strongly related another option could be that we simply drop the organization wide configuration and just have one for `Equinox` and `P2`? Respectively remove P2 from the Equinox configuration.

I have now implemented this in this PR.
With that this is ready from my side.

I'll also update the GH organization pages accordingly later, but I have to leave for now.

releng/org.eclipse.equinox.releng/EquinoxConfiguration.setup
merks: I think not. I think I changed this from the olden days when one typically wanted a gerrit URI for contribution, but now one can use anonymous and commit to a fork, and even the committers do that.

HannesWell: Yes, personally I usually fetch from the upstream/original repo using https and push to my fork via SSH.

Ok, good. Then we can probably remove it from all other configurations. I can take care of that.
Thanks.

HannesWell:
Since there were no objecting remarks in the meantime I think this is ready for submission.

PR #691, opened 2024-10-19
Introduce robust URL to java.nio.file.Path conversion

HannesWell:
Converting a URL into a `Path` or `File` is not a trivial task, especially since URLs are often malformed and therefore the straight forward way `Path.of(url.toURI())` or `new File(url.toURI())` lead to a `URISyntaxException` being thrown. Therefore multiple different workarounds are used throughout the SDK, e.g. `Path.of(url.getPath())` or `Path.of(url.getFile())`, but they are again have their problems.

This PR suggest a new utility method in the `URIUtil` that is capable of all the different conversion methods and workarounds and has the goal to reliably convert `file`-URLs into `java.nio.file.Path`s regardless of if the URL is malformed or perfectly complying to the standard.

This could also help for https://github.com/eclipse-platform/eclipse.platform/issues/35 because one does not have to rely on malformed URLs anymore if they are just passed around internally (URLs displayed to the user are another topic).

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
@@ -15,6 +15,7 @@ package org.eclipse.core.runtime;
 
 import java.io.File;
 import java.net.*;
+import java.nio.file.Path;
 
 /**
  * A utility class for manipulating URIs. This class works around some of the
@@ -312,11 +313,12 @@ public final class URIUtil {
 		// URL behaves differently across platforms so for file: URLs we parse from
 		// string form
 		if (SCHEME_FILE.equals(url.getProtocol())) {
-			String pathString = url.toExternalForm().substring(5);
+			String pathString = url.toExternalForm().substring(SCHEME_FILE.length() + 1);
 			// ensure there is a leading slash to handle common malformed URLs such as
 			// file:c:/tmp
-			if (pathString.indexOf('/') != 0)
+			if (!pathString.startsWith("/")) { //$NON-NLS-1$
 				pathString = '/' + pathString;
+			}
 			return toURI(SCHEME_FILE, null, pathString, null);
 		}
 		try {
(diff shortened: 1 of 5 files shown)


bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
```diff
@@ -328,6 +330,33 @@ public static URI toURI(URL url) throws URISyntaxException {
...
 
+	/**
+	 * Returns the {@code Path} specified by the given {@code file} {@link URL}.
+	 * 
+	 * @throws IllegalArgumentException if the specified URL does not use the
+	 *                                  {@code file} protocol
+	 * 
+	 * @since 3.20
+	 */
+	public static Path toFilePath(URL url) {
```
HannesWell, line 341: That the name contains `File` and ` Path` seems to be a duplication at first, but since any kind of URL can have a path I think it should be made clear that it's a file-system path (although nio-Path can also handle non-classical file-systems).
On the other hand `toFileSystemPath()` seemed to be a bit verbose to me.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
```diff
@@ -328,6 +330,33 @@ public static URI toURI(URL url) throws URISyntaxException {
...
+		if (!SCHEME_FILE.equals(url.getProtocol())) {
+			throw new IllegalArgumentException("Not a 'file' url: " + url); //$NON-NLS-1$
+		}
```
HannesWell, line 344: Alternatively this method could return an empty optional if the URL is not a file-URL.
But since callers that are able to handle other protocols can simple check the precondition before calling this method, I think it's fine to keep it simple for those cases where a file-URL is used with certainty.

laeubi: Optional seem much nicer, calles can throw exception if the want anyways with `Optional.orElseThrow(...)`

Then we can easily handel other error conditions here as well for example if invalid characters are in the path.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
HannesWell: I'm torn here as well. Currently the `URIUtil` class has both kind of methods: `toXYZ()` and `fromXYZ()`.
On one hand, it's clear from the method signature that a file-system Path is returned so we could reflect the file-URL requirement in the name. On the other hand, since this class is named `URIUtil`, just the method name `fromFileUrl()` could imply that a URI is created (which is indeed the case for the `fromString()` method).

A 'full' name could be `filePathFromFileURL()` or just `pathFromFileURL()` but I fear the latter misses the 'file-system' aspect and both are quite lengthy.

Maybe a different style like `fileURLToPath()` would also fit?

bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/URIUtil.java
HannesWell: In some cases it's indeed nicer, but there are many cases where the callers have already checked or are sure that it's a file-URL and then it's more cumbersome.
Furthermore, if we check more error cases where using an optional erases detailed error information as one just gets an empty optional without more context. Using exceptions instead allows us to add more details.

laeubi:
> This PR suggest a new utility method in the `URIUtil` that is capable of all the different conversion methods and workarounds and has the goal to reliably convert `file`-URLs into `java.nio.file.Path`s regardless of if the URL is malformed or perfectly complying to the standard.

Why not have a new `PathUtil` then if its not about URI...? Or put it into `FileLocator`, as we already have `public static URL toFileURL(URL url) throws IOException` there...

PR #693, opened 2024-10-30
Set the original cause of the IOException for the tryLock call

### LINKED ISSUE #692: Locker_JavaNio.lock() doesn't wrap I/O exceptions
We have this exception printed on standard error:

```
Caused by: java.io.IOException: An error occurred while locking file "...": "No locks available". A common reason is that the file system or Runtime Environment does not support file locking for that location. Please choose a different location, or disable file locking by passing "-Dosgi.locking=none" as a VM argument.
	at org.eclipse.osgi.internal.location.Locker_JavaNio.lock(Locker_JavaNio.java:53)
	at org.eclipse.osgi.storagemanager.StorageManager.lock(StorageManager.java:403)
	at org.eclipse.osgi.storagemanager.StorageManager.open(StorageManager.java:716)
	at org.eclipse.osgi.storage.Storage.getChildStorageManager(Storage.java:2241)
	at org.eclipse.osgi.storage.Storage.getInfoInputStream(Storage.java:2258)
	at org.eclipse.osgi.storage.Storage.<init>(Storage.java:253)
	at org.eclipse.osgi.storage.Storage.createStorage(Storage.java:181)
	at org.eclipse.osgi.internal.framework.EquinoxContainer.<init>(EquinoxContainer.java:108)
```

While the error message tries to explain what went wrong, I would still like to know what the actual occurred `IOException` is. It should get wrapped, not just used to extend the error message:

```
	@Override
	public synchronized boolean lock() throws IOException {
		raFile = new RandomAccessFile(lockFile, "rw"); //$NON-NLS-1$
		try {
			/*
			 * fix for bug http://bugs.sun.com/view_bug.do?bug_id=6628575 and
			 * https://bugs.eclipse.org/bugs/show_bug.cgi?id=44735#c17
			 */
			fileLock = raFile.getChannel().tryLock(0, 1, false);
		} catch (IOException ioe) {
			// print exception if debugging
			if (debug)
				System.out.println(NLS.bind(Msg.location_cannotLock, lockFile));
			// produce a more specific message for clients
			String specificMessage = NLS.bind(Msg.location_cannotLockNIO, new Object[] {lockFile, ioe.getMessage(), "\"-D" + LocationHelper.PROP_OSGI_LOCKING + "=none\""}); //$NON-NLS-1$ //$NON-NLS-2$
			throw new IOException(specificMessage);
		} catch (OverlappingFileLockException e) {
			// handle it as null result
			fileLock = null;
		} finally {
			if (fileLock != null)
				return true;
			raFile.close();
			raFile = null;
		}
		return false;
	}
```

I essentially have to `grep` OpenJDK code to find where "No locks available" is coming from.

tjwatson:
Fixes #692

### DIFF
--- a/bundles/org.eclipse.osgi/supplement/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi/supplement/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.supplement
-Bundle-Version: 1.11.0.qualifier
+Bundle-Version: 1.11.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.equinox.log;version="1.1",
--- a/bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/internal/location/Locker_JavaNio.java
+++ b/bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/internal/location/Locker_JavaNio.java
@@ -15,7 +15,10 @@
  *******************************************************************************/
 package org.eclipse.osgi.internal.location;
 
-import java.io.*;
+import java.io.File;
+import java.io.FileNotFoundException;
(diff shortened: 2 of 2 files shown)


trancexpress:
Thanks @tjwatson !

PR #698, opened 2024-11-15
Get loader-classname lock for calls to findLoadedClass

tjwatson:


tjwatson:
Closing this PR to create a new one.  Something got mixed up with the branches.  I tried to force push back to the original, but still has the `Stop throwing IllegalStateException` commit here.

PR #699, opened 2024-11-15
Stop throwing IllegalStateException

### LINKED ISSUE #172: OSGi core R9 will remove ISE on some methods for BundleContext, ServiceObjects and ServiceRegistration
See:
https://github.com/osgi/osgi/issues/454
https://github.com/osgi/osgi/pull/462
And
https://github.com/osgi/osgi/issues/486
https://github.com/osgi/osgi/pull/522

The following methods are removing the ISE to improve usability.

- Bundle.uninstall

- BundleContext.getBundle
- BundleContext.removeServiceListener
- BundleContext.removeBundleListener
- BundleContext.removeFrameworkListener
- BundleContext.ungetService

- ServiceObjects.ungetService

- ServiceRegistration.unregister

I think we can safely remove these now, even in our R8 implementation.

tjwatson:
The following methods can stop throwing IllegalStateExceptions

Bundle.uninstall
BundleContext.createFilter
BundleContext.getBundle
BundleContext.getProperty
BundleContext.removeServiceListener
BundleContext.removeBundleListener
BundleContext.removeFrameworkListener
BundleContext.ungetService
ServiceObjects.ungetService
ServiceRegistration.unregister

Fixes #172

### DIFF
--- a/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Core OSGi Tests
 Bundle-SymbolicName: org.eclipse.osgi.tests;singleton:=true
-Bundle-Version: 3.20.300.qualifier
+Bundle-Version: 3.22.100.qualifier
 Bundle-Vendor: Eclipse.org
 Require-Bundle: 
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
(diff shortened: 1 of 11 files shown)


laeubi:
The TCK is not using usual surefire but https://tycho.eclipseprojects.io/doc/latest/tycho-surefire-plugin/bnd-test-mojo.html

tjwatson:
> The TCK is not using usual surefire but https://tycho.eclipseprojects.io/doc/latest/tycho-surefire-plugin/bnd-test-mojo.html

If I understand that correctly, it appears I can use `<exclude/>` to exclude complete test classes, but not single test methods?  I would not want to exclude all the tests in https://github.com/osgi/osgi/blob/main/org.osgi.test.cases.framework/src/org/osgi/test/cases/framework/junit/service/ServiceRegistryTests.java

laeubi:
We "only" call the bnd-test here so if there is an option to skip one test you can do it here

https://github.com/eclipse-equinox/equinox/blob/2815608394cba2ec5529c7a5ec861e7c34113dd2/pom.xml#L116-L129

as far as I know last time I checked it was not that fine grained but might have changed (or can be improved) at bnd-lib.

PR #700, opened 2024-11-18
Get loader-classname lock for calls to findLoadedClass

tjwatson:


tjwatson:
I do not plan to merge this until when the development branches open for 2025-03.  At that time I will need to bump the version of org.eclipse.osgi.

PR #702, opened 2024-11-23
Add manual trigger for 'updateRelease'-workflow and use project-bot

HannesWell:
Adding a manual trigger allows to re-run that job manually with the latest changes applied to the referenced workflow from the releng.aggregator repository.

And since we have a eclipse bot specifically for each project we can use that bot as author of these automated changes instead of the generic GH bot.

Uses https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2571

### DIFF
--- a/.github/workflows/updateRelease.yml
+++ b/.github/workflows/updateRelease.yml
@@ -2,11 +2,23 @@ name: Update to next release
 on:
   milestone:
     types: [created]
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


HannesWell:
This is obsolete as https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2571 is obsolete as well.

Issue #704, opened 2024-11-25
Secure storage was unable to save the master password


Mailaender:
I am on KUbuntu 24.10 and the Secure Storage Linux Integegration fails at setting the master password. `libsecret-1-0` is installed.

```
java.lang.SecurityException: couldn't calculate mac
	at org.eclipse.equinox.internal.security.linux.LinuxPasswordProvider.requireNoError(LinuxPasswordProvider.java:141)
	at org.eclipse.equinox.internal.security.linux.LinuxPasswordProvider.saveMasterPassword(LinuxPasswordProvider.java:133)
	at org.eclipse.equinox.internal.security.linux.LinuxPasswordProvider.getPassword(LinuxPasswordProvider.java:168)
	at org.eclipse.equinox.internal.security.storage.PasswordProviderModuleExt.getPassword(PasswordProviderModuleExt.java:44)
	at org.eclipse.equinox.internal.security.storage.SecurePreferencesRoot.onChangePassword(SecurePreferencesRoot.java:338)
	at org.eclipse.equinox.internal.security.storage.SecurePreferences.passwordChanging(SecurePreferences.java:517)
	at org.eclipse.equinox.internal.security.storage.SecurePreferencesWrapper.passwordChanging(SecurePreferencesWrapper.java:201)
	at org.eclipse.equinox.internal.security.storage.friends.ReEncrypter.switchToNewPassword(ReEncrypter.java:115)
	at org.eclipse.equinox.internal.security.ui.storage.TabPassword.changePassword(TabPassword.java:336)
	at org.eclipse.equinox.internal.security.ui.storage.TabPassword$3.widgetSelected(TabPassword.java:145)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:265)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1617)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:5065)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4517)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)
	at org.eclipse.jface.window.Window.open(Window.java:799)
[...]

PR #705, opened 2024-11-25
Fixed couldn't calculate mac due to attributes null pointer

Mailaender:
Closes https://github.com/eclipse-equinox/equinox/issues/704


akurtakov:
Please add one more commit bumping the bundle version to 1.1.400

Mailaender:
Actually saving does not work either. I just made the exception go away.

PR #678, opened 2024-09-25
Set up workflow for automated version increments in pull-requests 

HannesWell:
Set up for this repository the workflow for automated version increments in pull-requests added via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352

Before this can work complete, the EF infra team has to add the required PAT of the Platform-bot to this repo as described in the linked PR and requested in
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5020#note_2825941

PR #706, opened 2024-11-26
fix Unnecessary @SuppressWarnings("deprecation")

jukzi:
jdt does not require @SuppressWarnings("deprecation")
 for imports anymore.


HannesWell:
> @HannesWell this repo does not have your version increment bot yet?

Yes, that's unfortunately right, see https://github.com/eclipse-equinox/equinox/pull/678.
I didn't have the time for a 'perfect' solution in Tycho yet and didn't want to spend extra time for a work-around just for Equinox. But maybe I should do the latter.

PR #707, opened 2024-11-27
Updated bundles with comparator errors

trancexpress:
Changes to ecj are for handling of switch on String.

See: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2595


akurtakov:
@HannesWell Is there a chance to make the autobump handle features too (like in this case)?

HannesWell:
> @HannesWell Is there a chance to make the autobump handle features too (like in this case)?

The workflow handles features already.
It's just not enabled for the repository at all :/

- https://github.com/eclipse-equinox/equinox/pull/678

akurtakov:
How easy it is to get used to automation and take it for granted!

PR #712, opened 2024-12-04
Move product to common

laeubi:
Currently we have a split-package between equinox-common and eclipse-platform.

This moves the interface
- IProductProvider
- IProduct

from platform to commons to reduce the amount of split-api, I tried to recover the history as much as possible for these files even though there is not much changed in the last years. If anyone wants to verify this it would be appreciated.

When this PR is merged the classes can be deleted from the platform bundle making it only 5 classes left to migrate before the split is resolved.


laeubi:
The ECA check fails as this seem to contain some verrrryyyy old commits that don't have a suitable author to be identified by the check.

laeubi:
Only the usual TCK fail, @tjwatson what did you think?

laeubi:
>  In the long run it may be best to move all the split "parts" of org.eclipse.core.runtime back to org.eclipse.core.runtime bundle in platform.

I don't think that's really possible as `org.eclipse.core.runtime` requires equinox common so there is not really a backward compatible path in my opinion. So my goal would be more to get rid of most things in `org.eclipse.core.runtime` and then simply stop using it.

PR #714, opened 2024-12-05
[#713] API to identify bundle resources (like images)

### LINKED ISSUE #713: API to identify bundle resources (like images)
We have a [discussion in PDE](https://github.com/eclipse-pde/eclipse.pde/discussions/1315) and it partially addresses the problem of identifying package resources (like images) inside Java code to establish PDE API control for them.

This issue was created to discuss a specific implementation proposal, 
an excerpt from it is given here:
```java
public record ImageIdentityRecord(BundleIdentity bundle, String path, String key) implements ImageIdentity {

	@Override
	public String id() {
		return bundle.symbolic() + ".image." + key; //$NON-NLS-1$
	}

	@Override
	public String url() {
		return "platform:/plugin/" + bundle.symbolic() + "/$nl$/" + path; //$NON-NLS-1$ //$NON-NLS-2$
	}

}

```

PR with details will follow shortly

ruspl-afed:
`ImageIdentity` interface with surrounding types

Fixes #713

### DIFF
--- a/bundles/org.eclipse.equinox.common/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.common/META-INF/MANIFEST.MF
@@ -16,6 +16,7 @@ Export-Package: org.eclipse.core.internal.boot;x-friends:="org.eclipse.core.reso
    org.eclipse.equinox.security",
  org.eclipse.core.runtime;common=split;version="3.7.0";mandatory:=common,
  org.eclipse.core.text;version="3.13.0",
+ org.eclipse.equinox.common.identity;x-internal:=true,
  org.eclipse.equinox.events;version="1.0.0"
 Bundle-Vendor: %providerName
 Bundle-Activator: org.eclipse.core.internal.runtime.Activator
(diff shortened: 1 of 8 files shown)


fedejeanne:
Hi Alex. Just following up yesterday's discussion in the community call. I have a question: would it be possible with this approach to somehow "swap" images/icons in runtime?

The reason I ask is because a student of mine is taking a look at how to switch (all) icons at runtime based in some property (let's call it "icon theming") and I thought the solution you propose here might be useful for that?

Our initial idea was to provide the themes in bundles and to put (all) icons inside a folder with the name of the theme in there, something like:
```
themeA (bundle)
'-- resources
  '-- icons
       '-- themeA
           '-- ... (the icons)

themeB (bundle)
'-- resources
  '-- icons
       '-- themeB
           '-- ... (the icons)
```

So if one could switch from `themeA` to `themeB` in runtime then icons would be taken from another folder. I assume one would have to add one placeholder to the URL of the icons in the consumers (something like today's `$nl`). 

Would that be doable with this approach?

ruspl-afed:
> would it be possible with this approach to somehow "swap" images/icons in runtime

This approach suggest rather normalized way to work with image resources than direct solution for "swap" problem.
However with suggested approach it may be a bit easier to implement what you want.
1) wrap `ImageRegistry` with another type `ImageIdentityRegistry` that works with `ImageIdentity`
2) provide Theme-aware implementation of `ImageIdentity` that will calculate URL based on active Theme
3) implement "change theme handling" for `ImageIdentityRegistry` -> it will clear and refill underlying `ImageRegistry` with images for new active Theme.

PR #717, opened 2024-12-12
Remove old junit 4.13.1 from o.e.osgi pom

akurtakov:
It's covered by the JUNIT_CONTAINER/4 on the classpath so no need to have junit explicitly listed in the pom as that requires extra maintenance to update it.


akurtakov:
Failing TCK is long standing. Merging.

PR #720, opened 2024-12-12
Move TextMatcher from Platform UI to Equinox Common

ptziegler:
The TextMatcher class is used in the UI component, despite not depending on any UI classes. By moving it to Equinox, it can be used anywhere in the Eclipse Platform.

### DIFF
--- a/bundles/org.eclipse.equinox.common/src/org/eclipse/core/text/StringMatcher.java
+++ b/bundles/org.eclipse.equinox.common/src/org/eclipse/core/text/StringMatcher.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2020 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,11 +10,12 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Thomas Wolf, Patrick Ziegler - support for matching individual words
  *******************************************************************************/
 package org.eclipse.core.text;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
+import java.util.regex.Pattern;
 
 /**
  * A string pattern matcher. Supports '*' and '?' wildcards.
@@ -22,6 +23,10 @@ import java.util.List;
  * @since 3.12
  */
 public final class StringMatcher {
+
+	private static final Pattern NON_WORD = Pattern.compile("\\W+", Pattern.UNICODE_CHARACTER_CLASS); //$NON-NLS-1$
+
+	private static final StringMatcher[] NO_MATCHERS = new StringMatcher[0];
 
 	private final String fPattern;
 
@@ -34,6 +39,8 @@ public final class StringMatcher {
 	private boolean fHasLeadingStar;
 
 	private boolean fHasTrailingStar;
+
+	private StringMatcher fParts[]; // the given pattern is split into space-separated sub-patterns
 
 	private String fSegments[]; // the given pattern is split into * separated segments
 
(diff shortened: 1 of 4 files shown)


ptziegler:
Prerequisite to https://github.com/eclipse-platform/eclipse.platform.ui/pull/2567

vogella:
@tjwatson any concerns here?

HannesWell:
I'm really not sure if this is the right place. Equinox common contains code related to the runtime (in the broadest sense).
Having text-handling here seems not to be a good fit.
But while writing this, I see there is already `org.eclipse.core.text.StringMatcher`, which really surprises me.

ptziegler:
> I'm really not sure if this is the right place. Equinox common contains code related to the runtime (in the broadest sense).
Having text-handling here seems not to be a good fit.

The main motivator for all of this are plugin dependencies. Several UI plugins use the TextMatcher and to prevent having to add any new dependencies, I have to find the a common denominator, which is `org.eclipse.core.runtime`.

But none of the re-exported plugins really have anything to do with text handling.
image

So the reason I picked Equinox is, as you noticed, the StringMatcher, because that's at least somewhat related.

laeubi:
`TextMatcher` even references `StringMatcher` in its API doc and as it is more about Strings (in contrast to JFace Documents) I think it is fine to have it here.

tjwatson:
> `TextMatcher` even references `StringMatcher` in its API doc and as it is more about Strings (in contrast to JFace Documents) I think it is fine to have it here.

The API signature does not reference `StringMatcher`.  The implementation detail does.

> @tjwatson any concerns here?

It seems in the spirit of https://bugs.eclipse.org/bugs/show_bug.cgi?id=508611 so I don't have an issue with the functionality existing in this package of Equinox common.

But TextMatcher vs. StringMatcher have arbitrary names that are confusing to one that has never used either of these classes.  It seems overkill to add a new slightly differently named class just to add the following public APIs, what which is a static method that could go anywhere:

```java
	/**
	 * Determines whether the given {@code text} matches the pattern.
	 *
	 * @param text String to match; must not be {@code null}
	 * @return {@code true} if the whole {@code text} matches the pattern;
	 *         {@code false} otherwise
	 * @throws IllegalArgumentException if {@code text == null}
	 */
	public boolean match(String text)

	/**
	 * Determines whether the given sub-string of {@code text} from {@code start}
	 * (inclusive) to {@code end} (exclusive) matches the pattern.
	 *
	 * @param text  String to match in; must not be {@code null}
	 * @param start start index (inclusive) within {@code text} of the sub-string to
	 *              match
	 * @param end   end index (exclusive) within {@code text} of the sub-string to
	 *              match
	 * @return {@code true} if the given slice of {@code text} matches the pattern;
	 *         {@code false} otherwise
	 * @throws IllegalArgumentException if {@code text == null}
	 */
	public boolean match(String text, int start, int end)

	/**
	 * Splits a given text into words.
	 *
	 * @param text to split
	 * @return the words of the text
	 */
	public static String[] getWords(String text) {
```

I think it would be better to enhance `StringMatcher` to have this functionality instead of introducing a new class.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/text/StringMatcher.java
```diff
@@ -141,19 +169,79 @@ public String toString() {
...
 		if (pattern == null) {
 			throw new IllegalArgumentException();
 		}
 		fIgnoreCase = ignoreCase;
 		fIgnoreWildCards = ignoreWildCards;
-		fPattern = pattern;
-		fLength = pattern.length();
+		fIgnoreWords = ignoreWords;
+		fPattern = pattern.trim();
+		fLength = fPattern.length();
```
ptziegler, line 233: I wonder if there is a better way to handle this here... Originally, the Platform was using a copy of this StringMatcher, where it would explicitly trim the input pattern. See https://bugs.eclipse.org/bugs/show_bug.cgi?id=570390

ptziegler:
> I think it would be better to enhance StringMatcher to have this functionality instead of introducing a new class.

That makes sense. Luckily, the TextMatcher effectively extends the StringMatcher, so it was fairly easy to add this new functionality.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/text/StringMatcher.java
tjwatson: It is unclear to me on if this could break existing uses of `StringMatcher` where `ignoreWords` would be `true`.

I also doubt you intended to `trim` here and also in `splitPattern` (https://github.com/eclipse-equinox/equinox/pull/720/files#diff-ba6befa586c50334dce15c73bd18a046e4c1194bd272fd52f461ceacf7b1ee7aR479)

ptziegler: > It is unclear to me on if this could break existing uses of StringMatcher where ignoreWords would be true.

That's what I wasn't sure about, either. I've updated it so that the string is only trimmed if `ignoredWords` is `false`, i.e. when the new functionality is used.

> I also doubt you intended to trim here and also in splitPattern (https://github.com/eclipse-equinox/equinox/pull/720/files#diff-ba6befa586c50334dce15c73bd18a046e4c1194bd272fd52f461ceacf7b1ee7aR479)

True, that was a left-over from the original implementation, which I've now also updated.

bundles/org.eclipse.equinox.common/src/org/eclipse/core/text/StringMatcher.java
```diff
@@ -141,14 +169,75 @@ public String toString() {
...
 		if (pattern == null) {
 			throw new IllegalArgumentException();
 		}
 		fIgnoreCase = ignoreCase;
 		fIgnoreWildCards = ignoreWildCards;
-		fPattern = pattern;
-		fLength = pattern.length();
-
+		fIgnoreWords = ignoreWords;
+		if (fIgnoreWords) {
```
tjwatson, line 232: With only my API hat on I think it would make more sense if we didn't add a new constructor to take an additional `boolean` for word matching.  Instead could we add a new method `public boolean matchWords(List<String> words)` or `public boolean matchWords(String[] words)`

I'm unsure how much that complicates the implementation.  I would imagine you would want to use an `AtomicReference<fParts[]>` to create the parts on demand if/when `matchWords` is called.

ptziegler: >  Instead could we add a new method public boolean matchWords(List<String> words) or public boolean matchWords(String[] words)

I suppose that would work.  Though I think the method has to accept the whole text, given that the sub-patterns are matched against the text as well as the individual words. I can't really predict the impact if I were to change this behavior.

>  I would imagine you would want to use an AtomicReference<fParts[]> to create the parts on demand if/when matchWords is called.

I'm not sure dead-set on the `final` modifier, so an `AtomicReference` sounds like overkill if I can do a lazy initialization by checking whether `fParts == null`.

> I'm unsure how much that complicates the implementation.

It'll primarly make the check more verbose. So instead of `matcher.match(text)` it'll become `matcher.match(text) || matcher.matchWords(text)`. There are 4-5 Platform classes that need to be updated, so I think that's a doable amount of work.

It obviously also means that the trimming of the pattern has to be done explicitly, but that's also fine.

ptziegler: > It'll primarly make the check more verbose. So instead of matcher.match(text) it'll become matcher.match(text) || matcher.matchWords(text). There are 4-5 Platform classes that need to be updated, so I think that's a doable amount of work.

Alternatively, I can also simply call `match(text)` inside `matchWords(text)` which I think is preferable.

tjwatson: > > It'll primarly make the check more verbose. So instead of matcher.match(text) it'll become matcher.match(text) || matcher.matchWords(text). There are 4-5 Platform classes that need to be updated, so I think that's a doable amount of work.
> 
> Alternatively, I can also simply call `match(text)` inside `matchWords(text)` which I think is preferable.

That was my intention, make the new method do what you wanted to do with the existing method.

laeubi:
@tjwatson any more concerns or can we merge this?

tjwatson (approved):

PR #721, opened 2024-12-15
Skip SecurityManager related tests if SecurityManager is disallowed

HannesWell:
Part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2623

See also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2659#issuecomment-2543020833.

### DIFF
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/bundles/EquinoxBundleAdaptTests.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/bundles/EquinoxBundleAdaptTests.java
@@ -24,6 +24,8 @@ import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.osgi.container.Module;
 import org.eclipse.osgi.internal.framework.EquinoxBundle;
 import org.eclipse.osgi.signedcontent.SignedContent;
+import org.eclipse.osgi.tests.securityadmin.SecurityManagerTests;
+import org.junit.Assume;
 import org.junit.Test;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/bundles/EquinoxBundleAdaptTests.java
```diff
@@ -42,17 +43,23 @@ public void testAdapt_Module() throws Exception {
...
 		SecurityManager previousSM = System.getSecurityManager();
+		boolean unsetSecurityManager = false;
 		try {
-			System.setSecurityManager(new SecurityManager() {
-				@Override
-				public void checkPermission(Permission perm) {
-					// Just permit everything
-				}
-			});
+			if (SecurityManagerTests.isSecurityManagerAllowed()) {
```
laeubi, line 48: can we probabbly use something like `asumeThat(SecurityManagerTests.isSecurityManagerAllowed())` instead?

HannesWell: For this test-case I wasn't 100% sure if it's useless if the security-manager is disallowed or removed.
It could still test that `bundle.adapt(java.security.ProtectionDomain.class)` returns not null, even when the  `SecurityManager` is still gone/disabled. But of course the main value is checking if the `SecurityManager` is queried for the corresponding permission.

HannesWell: OK, it looks like this test does not work at all if a security-manager is not allowed, because a ProtectionDomain is never returned if the security-manager is null:
https://github.com/eclipse-equinox/equinox/blob/372338078b3d5c908fcc5ed39a10683e9ea10dff/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/BundleInfo.java#L332-L335

Even better, just skipping the test is simpler.

PR #728, opened 2024-12-19
Use Java 21 in Github actions

laeubi:
Platorm requires Java 21 now on some places so we should be prepared.


laeubi:
This now fails to execute (or more formally skip) test that require a security manager, the ConfigAdmin text seem to not run at all anymore.

PR #730, opened 2024-12-21
Add java 21 to the list of installed JVMs

laeubi:
Lets see if adding Java 21 already make it work (but still run/test with 17 as default).


laeubi:
Only the usual CM failure.

PR #731, opened 2024-12-21
Add "replace" as a default available transformer

laeubi:
A while back in the incubation phase [there was a dedicated bundle](https://eclipse.googlesource.com/equinox/rt.equinox.incubator/+/refs/tags/I20120516-1900/components/bundles/org.eclipse.equinox.transforms.replace) that provided a "replace" transformer.

As this is the very basic transformation and do not require any special dependency it is now added back (in a slightly enhanced implementation) as being available by default.

This contributes to
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/14#issuecomment-2558026326

PR #732, opened 2024-12-21
Add Extender that allows bundles to declare a transformer resource

laeubi:
Currently it is rather inconvenient to make a transformer list available as it requires to register an URL as service and the service itself also needs to be registered with some properties.

This now adds an bundle extender that scans for a header 'Equinox-Transformer' where the value is pointing to a resource in the bundle (optionally specify the transformer type). If such bundle is found the resource is looked up and registered as a service with the required properties so it can be discovered by the transformer.

Contributes to
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/14#issuecomment-2558026326

Requires:
- https://github.com/eclipse-equinox/equinox/pull/731


mickaelistria:
Can you please add a tip about this powerful addition to the release notes?

PR #734, opened 2024-12-31
Mark deprecated methods for removal

laeubi:


laeubi:
We can't use since / forRemoval here due to java 1.8 compatibility. @tjwatson since this is deprecated since ages (somewhere back to 2006 if I look at the git history) should we probably just remote them without further notice?

tjwatson:
> We can't use since / forRemoval here due to java 1.8 compatibility. @tjwatson since this is deprecated since ages (somewhere back to 2006 if I look at the git history) should we probably just remote them without further notice?

I'm not that motivated to remove individual methods from what I consider a completely deprecated package.

laeubi:
> I'm not that motivated to remove individual methods from what I consider a completely deprecated package.

Maybe we should then add the new deprecated directive to that package? I assumed until now that only the implementation (that already resides in a fragment) is deprecated but not the API itself (what is used in P2 and PDE...).

PR #738, opened 2025-01-12
Fix maven group id to match the actual deployment

merks:
Fix the remaining bundles not addressed by the previous commit.

https://github.com/eclipse-equinox/equinox/pull/737


merks:
Another one of Oomph's handy dandy tools builds an overview of all the POMs in a folder location and with that it's easy to see if any are using an inappropriate group ID:

image

This helps to easily find ones that still needed fixing.    (Of course the group ID only matters for bundles.)

laeubi:
Thanks, I missed that ones because I assumed they are all pomless already!

merks:
I believe that one TCK always fails.

PR #740, opened 2025-01-14
Create an SLF4j provider implementation for the Equinox Log

laeubi:
Currently when running an application with Equinox that uses other logging techniques one has some kind if mixture when it comes to logging as some parts are logging to Equinox Log (especially eclipse platform) and some are logging to something else (e.g. jul, log4j, ...).

A common approach for this is to use bridges that redirect different loggings to one api (e.g. slf4j) and then installing one desired provider to use that for unified logging.

To support such deployments, this now adds a slf4-provider that acts as such provider for slf4j so one can use it with slf4j api and also with such bridges to get a unified logging.

### To test this do the following:

1. Make sure no other provider is included into your product, e.g. start it with -console and type in `ss slf4j`, if needed remove all and if you start your product you should see:
```
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
```

2. Now add the bundle `org.eclipse.equinox.slf4j` to your product and make sure it includes `org.apache.aries.spifly.dynamic.bundle` as well and is started at level 1

An example product can be found here:

```
<?xml version="1.0" encoding="UTF-8"?>
<?pde version="3.5"?>

<product uid="testslf4j" version="1.0.0.qualifier" type="bundles" includeLaunchers="true" autoIncludeRequirements="false">

   <configIni use="default">
   </configIni>

   <launcherArgs>
      <programArgs>-console
      </programArgs>
      <vmArgs>-Declipse.ignoreApp=true
-Dosgi.noShutdown=true
      </vmArgs>
      <vmArgsMac>-XstartOnFirstThread -Dorg.eclipse.swt.internal.carbon.smallFonts
      </vmArgsMac>
   </launcherArgs>

   <launcher>
      <win useIco="false">
         <bmp/>
      </win>
   </launcher>

   <vm>
   </vm>

   <plugins>
      <plugin id="my.logging.bundle"/>
      <plugin id="org.apache.aries.spifly.dynamic.bundle"/>
      <plugin id="org.apache.felix.gogo.command"/>
      <plugin id="org.apache.felix.gogo.runtime"/>
      <plugin id="org.apache.felix.gogo.shell"/>
      <plugin id="org.apache.felix.scr"/>
      <plugin id="org.eclipse.equinox.common"/>
      <plugin id="org.eclipse.equinox.console"/>
      <plugin id="org.eclipse.equinox.slf4j"/>
      <plugin id="org.eclipse.osgi"/>
      <plugin id="org.objectweb.asm"/>
      <plugin id="org.objectweb.asm.commons"/>
      <plugin id="org.objectweb.asm.tree"/>
      <plugin id="org.objectweb.asm.tree.analysis"/>
      <plugin id="org.objectweb.asm.util"/>
      <plugin id="org.osgi.service.component"/>
      <plugin id="org.osgi.util.function"/>
      <plugin id="org.osgi.util.promise"/>
      <plugin id="slf4j.api"/>
   </plugins>

   <configurations>
      <plugin id="my.logging.bundle" autoStart="true" startLevel="0" />
      <plugin id="org.apache.aries.spifly.dynamic.bundle" autoStart="true" startLevel="1" />
      <plugin id="org.apache.felix.scr" autoStart="true" startLevel="2" />
      <plugin id="org.eclipse.core.runtime" autoStart="true" startLevel="0" />
      <plugin id="org.eclipse.equinox.common" autoStart="true" startLevel="2" />
      <plugin id="org.eclipse.equinox.event" autoStart="true" startLevel="2" />
      <plugin id="org.eclipse.equinox.simpleconfigurator" autoStart="true" startLevel="1" />
   </configurations>

</product>
```

that has a `my.logging.bundle` bundle using SLF4J and then you will see log messages like this:

```
!SESSION 2025-02-09 08:49:41.906 -----------------------------------------------
eclipse.buildId=unknown
java.version=21
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=de_DE
Framework arguments:  -application 
Command-line arguments:  -application  -data /opt/eclipse/platform-sdk/ws/../runtime-testslf4j.product -dev file:///opt/eclipse/platform-sdk/ws/.metadata/.plugins/org.eclipse.pde.core/testslf4j.product/dev.properties -os linux -ws gtk -arch x86_64 -consoleLog -console

!ENTRY my.logging.bundle 2 0 2025-02-09 08:49:42.376
!MESSAGE I warn you!

!ENTRY my.logging.bundle 1 0 2025-02-09 08:49:42.377
!MESSAGE This is an info

!ENTRY my.logging.bundle 0 0 2025-02-09 08:49:42.377
!MESSAGE Debug
```

### DIFF
--- a/bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
+++ b/bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
@@ -1,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.equinox.slf4j;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.eclipse.core.internal.runtime.PlatformLogWriter;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.service.log.LogService;
+import org.osgi.service.log.Logger;
+import org.slf4j.ILoggerFactory;
+
+/**
+ * An implementation of the SLF4J {@link ILoggerFactory} that is using the
+ * equinox platform log like done with the ILog interface
+ */
+public class EquinoxLoggerFactory implements org.slf4j.ILoggerFactory {
+
+	private static final StackWalker STACK_WALKER = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE);
+	private static final Bundle FACTORY_BUNDLE = FrameworkUtil.getBundle(EquinoxLoggerFactory.class);
+	private static final Bundle SLF4J_BUNDLE = FrameworkUtil.getBundle(org.slf4j.ILoggerFactory.class);
+	static final ConcurrentMap<Bundle, EquinoxLogger> LOGGER_MAP = new ConcurrentHashMap<>();
+	static final AtomicReference<LogService> logService = new AtomicReference<>();
+
+	@Override
+	public org.slf4j.Logger getLogger(String name) {
+		Bundle bundle = getBundle();
+		if (bundle == null) {
+			return new EquinoxLogger(name, null);
+		}
+		return LOGGER_MAP.computeIfAbsent(bundle, b -> {
+			LogService service = logService.get();
+			@SuppressWarnings("restriction")
+			Logger bundleLogger = service == null ? null
+					: service.getLogger(b, PlatformLogWriter.EQUINOX_LOGGER_NAME, Logger.class);
+			return new EquinoxLogger(name, bundleLogger);
+		});
+	}
+
+	private Bundle getBundle() {
+		return STACK_WALKER.walk(s -> s.map(frame -> FrameworkUtil.getBundle(frame.getDeclaringClass()))
+				.dropWhile(stackFrameBundle -> stackFrameBundle == FACTORY_BUNDLE || stackFrameBundle == SLF4J_BUNDLE)
+				.findFirst().orElse(null));
+	}
+
+}
--- a/bundles/org.eclipse.equinox.slf4j/.classpath
+++ b/bundles/org.eclipse.equinox.slf4j/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="output" path="bin"/>
+</classpath>
--- a/bundles/org.eclipse.equinox.slf4j/.project
+++ b/bundles/org.eclipse.equinox.slf4j/.project
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.eclipse.equinox.slf4j</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.pde.PluginNature</nature>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+	</natures>
+</projectDescription>
--- a/bundles/org.eclipse.equinox.slf4j/.settings/org.eclipse.core.resources.prefs
+++ b/bundles/org.eclipse.equinox.slf4j/.settings/org.eclipse.core.resources.prefs
@@ -1,0 +1,2 @@
+eclipse.preferences.version=1
+encoding/<project>=UTF-8
--- a/bundles/org.eclipse.equinox.slf4j/.settings/org.eclipse.jdt.core.prefs
+++ b/bundles/org.eclipse.equinox.slf4j/.settings/org.eclipse.jdt.core.prefs
@@ -1,0 +1,9 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=17
+org.eclipse.jdt.core.compiler.compliance=17
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enablePreviewFeatures=disabled
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=warning
+org.eclipse.jdt.core.compiler.release=enabled
+org.eclipse.jdt.core.compiler.source=17
--- a/bundles/org.eclipse.equinox.slf4j/.settings/org.eclipse.pde.core.prefs
+++ b/bundles/org.eclipse.equinox.slf4j/.settings/org.eclipse.pde.core.prefs
@@ -1,0 +1,4 @@
+eclipse.preferences.version=1
+pluginProject.equinox=false
+pluginProject.extensions=false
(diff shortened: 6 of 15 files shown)


bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
```diff
@@ -0,0 +1,89 @@
...
+ * An implementation of the SLF4J {@link ILoggerFactory} that is using the
+ * equinox platform log like done with the ILog interface
+ */
+public class EquinoxLoggerFactory implements org.slf4j.ILoggerFactory {
+
+	private static final StackWalker STACK_WALKER = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE);
+	private static final Bundle FACTORY_BUNDLE = FrameworkUtil.getBundle(EquinoxLoggerFactory.class);
+	private static final Bundle SLF4J_BUNDLE = FrameworkUtil.getBundle(org.slf4j.ILoggerFactory.class);
+	private static final ConcurrentMap<Bundle, EquinoxLogger> LOGGER_MAP = new ConcurrentHashMap<>();
+	private ExtendedLogService logService;
```
tjwatson, line 39: Can you get by with the LogService here?

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+	@Override
+	public org.slf4j.Logger getLogger(String name) {
+		Bundle bundle = getBundle();
+		if (bundle == null) {
+			return new EquinoxLogger(name, null, null);
+		}
+		return LOGGER_MAP.computeIfAbsent(bundle, b -> {
```
tjwatson, line 47: This will leak `LogService` objects when the bundle is uninstalled.  This bundle should have a BundleListener that clears this map when the bundle is uninstalled.  It should also clear this MAP when the slf4j bundle is stopped I think.  Also the logService object itself will need to be nulled out on stop so that if the bundle is restarted then it will reget the log service.

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
laeubi: Not really because `ExtendedLogService` has a special method with name+bundle that is also used internally in Platform, and I wanted to be en-par with that.

tjwatson: Pretty sure you can get by with `org.osgi.service.log.LoggerFactory.getLogger(Bundle, String, Class<L>)` which `LogService extends LoggerFactory`

laeubi: I first thought of that as well but it has additional check on the bundle and bundlestate :-\

Also we here explicitly reference the framework logger so I don't see much of an benefit using "standard" API.

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
laeubi: > This will leak LogService objects when the bundle is uninstalled. This bundle should have a BundleListener that clears this map when the bundle is uninstalled

Good point.

>  It should also clear this MAP when the slf4j bundle is stopped I think

Hm... I think it would not really matter because the slf4j do not really use services anywhere and as long as it is resolved client bundles can use the loggers. if it is really uninstalled this bundle will be stopped as well and all its services and classes released so I don't think any action is required here.

> Also the logService object itself will need to be nulled out on stop so that if the bundle is restarted then it will reget the log service.

Actually the log service here only ever changes when the framework is restarted so I could add such code but it would not make much of a difference I assume (beside making the implementation more complex).

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
tjwatson: > I first thought of that as well but it has additional check on the bundle and bundlestate :-\
> 
> Also we here explicitly reference the framework logger so I don't see much of an benefit using "standard" API.

Not sure why you want this slf4j stuff to work with a `null` bundle or one that is not resolved.

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
tjwatson: > Hm... I think it would not really matter because the slf4j do not really use services anywhere and as long as it is resolved client bundles can use the loggers. if it is really uninstalled this bundle will be stopped as well and all its services and classes released so I don't think any action is required here.

I don't mean this bundle, I mean the using bundle that you got from the stack walk.  If that bundle is uninstalled then you need to clear out this map.

tjwatson: > Actually the log service here only ever changes when the framework is restarted so I could add such code but it would not make much of a difference I assume (beside making the implementation more complex).

Clearing out the logservice on stop so that you re-get it on restart is just a best practice that I would like to have in new equinox code.  It doesn't seem like it would be difficult to do.

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
laeubi: Main point is I want it to make sure it is the Equinox Log service, so I can use that other method probably but will it help something?!?

laeubi: I now use `LogService` api as much as possible

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLoggerFactory.java
laeubi: This should be fixed now!

iloveeclipse:
1) Is this planned for 4.35?
2) Do I understand it right, this PR would allow bundle logging for bundles like jgit (which uses slf4j internally) be redirected to Eclipse workspace log?

laeubi:
> Is this planned for 4.35?

I could try to finish it if there is interest, I just not have tested it very much

> Do I understand it right, this PR would allow bundle logging for bundles like jgit (which uses slf4j internally) be redirected to Eclipse workspace log?

Yes.

laeubi:
I have now further tested this and its works as expected so far. So if there are no further concerns I want to merge this so it can be part of the release.

By intention this is currently not part of any features, so I assume it needs to be added to the `eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/category.xml` once it is available to be included in the repository/simrel contribution @merks ?

iloveeclipse:
Logging is a controversial terrain. How does it work with multiple slf4j "providers"? Which one will "win"? I believe without this patch one has to contribute something to see slf4j logs, so I assume users who did that will have troubles with this one? Or users would need to explicitly configure this provider?

Would be nice if you could add more information to the PR description about theory of operation, interoperability and configuration (if one required).

laeubi:
> How does it work with multiple slf4j "providers"? 

SLf4j by design always only picks one provider.

> I believe without this patch one has to contribute something to see slf4j logs

You can supply none (in wich case you get noop logging by default)

> so I assume users who did that will have troubles with this one?

One should always only include one logging provider in a certain product, so user (or usually product owners) has to choose wich one to use.

> Or users would need to explicitly configure this provider?

Yes you need to explicitly include it in your product (e.g. in an EPPs)

iloveeclipse:
> user (or usually product owners) has to choose wich one to use.

Could you please update PR  description with information how that should be done for proposed provider. Also please provide information how to verify that it works with some slf4j client in Exlipse. I would like to test it with jgit (which I believe uses slf4j). Not sure if Xtext uses slf4j too.
@cdietrich ?

merks:
I'm so sick of logging and logging problems.  It's just unreal how many different frameworks there are and how easily they run into conflicts:

https://github.com/eclipse-birt/birt/issues/2045

merks:
For it to end up in SimRel it probably needs to be listed:

image

laeubi:
> I would like to test it with jgit (which I believe uses slf4j)

Yes jgit uses slf4j, I updated the PR description.

iloveeclipse:
> I'm so sick of logging and logging problems

:) I feel with you.

One more question to understand deployment plan.

Do I understand it right, assuming the code here is merged to equinox, would it be automatically included into SDK, and SDK would by default also need to ship slf4j? 

Or is this supposed to be **optionally** available bundle on some update site?

laeubi:
SLF4J (API) is already part of SDK, but this (IMPL) should be not part of the sdk product (or at least not now) but part of the repository so people can include it in their product if they want to.

iloveeclipse:
> this (IMPL) should be not part of the sdk product (or at least not now) but part of the repository so people can include it in their product if they want to.

@merks : in that case it shouldn't increase your pain ?

merks:
No, it's just a pointless rant.   Maybe it can even be used in BIRT and make the world a better place...

laeubi:
I'm merging this now, further improvements can be made on top of this.

iloveeclipse:
@laeubi : I believe we should add a note to the N&N notes regarding this new bundle and the required product configuration to get it used with slf4j.

PR #744, opened 2025-01-17
Remove the Equinox p2 Examples working set from the setup

merks:
- Specify an ID for WorkingSetTask so that there are no positional EMF fragment path.


merks:
This has no build impact.  Merging immediately.

PR #765, opened 2025-01-24
EclipsePreferences: mt hardening

jukzi:
volatile booleans,
final descriptor


laeubi:
See 
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/604

PR #768, opened 2025-01-26
[Build] Fetch binaries in workflow for automated version increments

HannesWell:
Follow-up to
- https://github.com/eclipse-equinox/equinox/pull/678

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -19,4 +19,5 @@ jobs:
     with:
       botName: Eclipse Equinox Bot
       botMail: equinox-bot@eclipse.org
-      extra-maven-args: '-f bundles'
+      extra-setup-command: 'git clone --depth 1 https://github.com/eclipse-equinox/equinox.binaries.git binaries'
+      extra-maven-args: '-Dequinox.binaries.loc=$(pwd)/binaries'
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@ bin_test
 /bundles/archived/*/bin/
 .DS_Store
 
+# Ignore clone of equinox.binaries repo created in version-check/increment workflow
+/binaries/
+
 # maven
 target/
 # files generated (temporary) by the tycho maven builds


pom.xml
```diff
@@ -229,6 +230,45 @@
 		    <module>bundles/org.eclipse.osgi.tests</module>
         </modules>
     </profile>
+    <profile>
+      <id>fetch-natives</id>
```
HannesWell, line 234: Actually I would have liked to add this to the launcher-binary-parent, but because it's only the parent of all `org.eclipse.equinox.launcher.*` artifact the defined antrun plugin to clone the binaries-repo has a high chance to run concurrently which will probably lead to build failures du to concurrent clones of the repository.

Adding it to the root parent ensures that there is only exactly one execution of the clone.

HannesWell: Unfortunately I'm not aware of a convenient trick to just execute the mojo on the reactor root, which is why the execution runs on each project (if the profile is active) and just skips the cloning.

.github/workflows/pr-checks.yml
```diff
@@ -19,4 +19,4 @@ jobs:
     with:
       botName: Eclipse Equinox Bot
       botMail: equinox-bot@eclipse.org
-      extra-maven-args: '-f bundles'
+      extra-maven-args: '-Dequinox.binaries.loc=$(pwd)/binaries -Pfetch-natives'
```
HannesWell, line 22: An alternative to the new profile would be to use https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2783 and just add:
```suggestion
      extra-setup-command: 'git clone --depth 1 https://github.com/eclipse-equinox/equinox.binaries.git binaries'
      extra-maven-args: '-Dequinox.binaries.loc=$(pwd)/binaries -Pfetch-natives'
```
This would avoid all changes in the pom.xml files and would be much more compact.
On the other hand, having an 'internal' definition of the binaries-fetch would allow to re-use it in other workflows and even locally, which would simplify local builds as it avoids the need to clone the binaries repo in advance.

HannesWell:
But all of this cannot be verified since we seem to have a sever outage of download.eclipse.org...
https://www.eclipsestatus.io/

HannesWell:
Ok, for now lets keep it simple and  just use the approach from https://github.com/eclipse-equinox/equinox/pull/768#discussion_r1929801772.
The ability to fetch all binaries within the repo can still be applied later in a greater rework.

That being said, I think the current approach won't work well with changes in the launcher binaries since the version-check operates on the existing 'old' binaries and won't see changes. To consider even that the workflow would have to fetch the results of a native's build from Jenkins and then run again.
However I think that's an issue for a later rework as well.

HannesWell:
Testing the version-increment workflow was finally successful. Submitting.

PR #829, opened 2025-02-01
Move Equinox Launcher to Java-17

HannesWell:
Move Equinox Launcher to Java-17 as suggested in https://github.com/eclipse-equinox/equinox/pull/782#issuecomment-2615880254.
And leverage the new language features to modernize the code a bit.


vogella:
Thanks for this work @HannesWell 

What is the reason not to jump directly to Java 21? I think @tjwatson indicated that that would also be fine

laeubi:
I can see that `org.eclipse.equinox.launcher` is used in some places in Tycho to support running eclipse applications, so one should make sure that it is still possible to launch older eclipse installations.

> What is the reason not to jump directly to Java 21? I think @tjwatson indicated that that would also be fine

Large parts of Eclipse/Equinox can still run on Java 17, so what would be the immediate benefits here to break those?

bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -2609,6 +2592,7 @@ private void setMultiValueProperty(String property, String[] values) {
 	 * nobody will actually call them (unless they casted the policy to EclipsePolicy and
 	 * called our methods)
 	 */
+	@SuppressWarnings({"removal", "deprecation"})
```
laeubi, line 2595: If we have new warnings due to updated java version it should not be suppressed but fixed along side with the java update.

laeubi (dismissed):
I would wait until next release for this change, at least warnings/deprecation should be fixed and not suppressed.

bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
laeubi: 🙈

tjwatson: I do think there are scenarios (ant?) that still optionally set a security manager when using Java 21 so I don't think we can really remove this policy stuff just yet.

HannesWell:
> I would wait until next release for this change, at least warnings/deprecation should be fixed and not suppressed.

I'm fine postponing the entire change to the next dev-cycle, I don't think it's urgent. Maybe I'll add some follow-up changes in separate commits in the meantime, but they can later be moved to separate PRs as well.

bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
HannesWell: Ant isn't such scenario as it just sets the security-manager if the running JVM is blow 18:
- https://github.com/apache/ant/pull/216

But there could be other scenarios that require it. So I assume that part should stay as long as the security-manager code is removed.
Btw. is there a plan for how to deal with that? Or is the plan to keep it as long as it works/compiles on all released Java versions?

tjwatson: > Btw. is there a plan for how to deal with that? Or is the plan to keep it as long as it works/compiles on all released Java versions?

It depends on the component.  For the most part, there should be little references to the actual `SecurityManager` or `Policy` in our code itself.  The `Framework` itself has many places it is referenced, but they should be able to be contained into an internal utility that fails gracefully on a future JVM where the actual classes are removed.

The biggest issue is code all over the place that have references to `java.security.AccessController.doPrivileged(PrivilegedAction<String>, AccessControlContext)` and hundreds of implementations of `java.security.PrivilegedAction<T>` or  `java.security.PrivilegedExceptionAction<T>`

My suspicion is that these APIs will be around much longer than the `SecurityManager` and `Policy` classes, but will become no-op methods when calling `doPrivileged`.  I don't see them being able to truly remove the `doPrivileged` APIs until all the long term support of Java 8 finally end.

HannesWell:
I have now updated this PR to the latest master and also extended the (first) part of the modernization and clean-up.

So besides (mainly) just moving the launcher bundles to Java-17, which is done in the first commit, this PR contains a second commit with the mentioned modernization, that also uses new language features not usable before.
Besides that I added a third commit, that contains a unification of the different version processing ways, unless there are objections against that unification I plan to squash that into the second commit.

@tjwatson or @laeubi do you want to review this and have the second/third commit in another follow-up PR or do you prefer it in one to directly have the move leveraged?

HannesWell:
> @tjwatson or @laeubi do you want to review this and have the second/third commit in another follow-up PR or do you prefer it in one to directly have the move leveraged?

I have now made this simpler and moved all further enhancements to
- https://github.com/eclipse-equinox/equinox/pull/919

tjwatson:
> I have now made this simpler and moved all further enhancements to

Should this PR be closed then?

HannesWell:
> > I have now made this simpler and moved all further enhancements to
> 
> Should this PR be closed then?
> 

No, it's intended for submission and to contain the move to 17. The other one is a follow up. 

I just wanted to give some time for a final review, but I assume there are no objections.

laeubi:
> As I understand it: PDE's primary focus was to provide a first class development environment for Eclipse plugins for the Eclipse Platform. It only seems natural that when self-hosting the Eclipse Platform with PDE for development that it would use the same launcher that the Eclipse Platform uses.

But what is the difference (except splash screen!) to `EclipseStarter` then? A Junit-Plugin launch is also special in that regard, that one can of course run code that do not use Eclipse Plugins at all...

tjwatson:
> But what is the difference (except splash screen!) to EclipseStarter then? A Junit-Plugin launch is also special in that regard, that one can of course run code that do not use Eclipse Plugins at all...

I've lost track of the issue we are trying to solve here.  We already have a main class in `org.eclipse.osgi`.  I don't think we need to change that.  If `EclipseStarter` needs some enhancements then perhaps it is best to describe that and options to improve that for other cases.

Issue #831, opened 2025-02-01
Disable launcher making GTK3 calls (splash) when GTK4 is in use


jonahgraham:
For now we should disable all GTK3 calls in the launcher when GTK4 has been requested by the user. See https://github.com/eclipse-platform/eclipse.platform.swt/issues/1628#issuecomment-2613931878

In the future, when GTK4 becomes default, we can change it to the other way around.

Note that this is related to the dependency versions of GTK we have. xref https://github.com/eclipse-equinox/equinox/issues/830#issuecomment-2629042335

jonahgraham:
I will provide a patch for this.

Issue #190, opened 2023-01-20
org.eclipse.equinox.cm seem to fails some TCKs


laeubi:
Currently org.eclipse.equinox.cm seem to fail some TCK tests, a report XML that can be opened in Eclipse can be found here:
[TEST-org.osgi.test.cases.cm-8.1.0.202212101352.xml.zip](https://github.com/eclipse-equinox/equinox/files/10464027/TEST-org.osgi.test.cases.cm-8.1.0.202212101352.xml.zip)

laeubi:
@tjwatson can you take a look or have an idea what might causing this?

tjwatson:
> @tjwatson can you take a look or have an idea what might causing this?

It is because of the quick implementation I did to be able to update SCR in https://bugs.eclipse.org/bugs/show_bug.cgi?id=538717 didn't include support for the coordinator.

laeubi:
> > @tjwatson can you take a look or have an idea what might causing this?
> 
> It is because of the quick implementation I did to be able to update SCR in https://bugs.eclipse.org/bugs/show_bug.cgi?id=538717 didn't include support for the coordinator.

Do you plan to work on this or can explain what exactly is missing? Is it coordinator support at all or is it incomplete?

tjwatson:
> Do you plan to work on this or can explain what exactly is missing? Is it coordinator support at all or is it incomplete?

The Equinox coodinator implementation is complete.  The CM implementation has no integration with the coordinator as described by the specification: https://docs.osgi.org/specification/osgi.cmpn/8.1.0/service.cm.html#service.cm-coordinatorsupport

tjwatson:
> Do you plan to work on this?

No

laeubi:
I'm just wondering if felix.cm is compliant and do anything differently than equinox? Maybe we can then better use the felix impl im we have no missing features then?

tjwatson:
> I'm just wondering if felix.cm is compliant and do anything differently than equinox? Maybe we can then better use the felix impl im we have no missing features then?

Is there anything we need CM for?  I'm not sure I would pull in another implementation if we don't actually use CM for anything.  What would the point be?

laeubi:
> Is there anything we need CM for? 

I don't think the CM is part of any EPP yet (@jonahgraham ?) or required by anything (I didn't find it in platform directly but maybe @merks has a reference in simrel)

> I'm not sure I would pull in another implementation if we don't actually use CM for anything.

[...]

PR #833, opened 2025-02-02
Automatically turn off splash screen if SWT_GTK4=1

jonahgraham:
When using GTK4 we must not run any GTK3 initialization. Therefore automatically apply -nosplash when SWT_GTK4=1

Fixes https://github.com/eclipse-equinox/equinox/issues/831

### DIFF
--- a/features/org.eclipse.equinox.executable.feature/library/eclipse.c
+++ b/features/org.eclipse.equinox.executable.feature/library/eclipse.c
@@ -445,6 +445,20 @@ JNIEXPORT int run(int argc, _TCHAR* argv[], _TCHAR* vmArgs[])
 	setenv(firstThreadEnvVariable, "1", 1);
 #endif
 
+#ifdef LINUX
+	if (!noSplash) {
+		char *swtGtk4 = getenv("SWT_GTK4");
+		if (swtGtk4 != NULL && strcmp(swtGtk4, "1") == 0) {
+			// The eclipse launcher uses GTK3 exclusively, therefore when starting GTK4
+			// it must be started without a splash screen
+			// https://github.com/eclipse-equinox/equinox/issues/831
+			// If the user didn't already specify -nosplash print a warning and turn the splashscreen off
+			_ftprintf(stderr, "WARNING: SWT_GTK4 does not support splash screen yet. Therefore it has been disabled. To suppress this message launch with -nosplash\n");
+			noSplash = 1;
+		}
+	}
+#endif
+
 	return _run(argc, argv, vmArgs);
 }
 


merks:
There are errors like this suggesting a version bump is needed:
> 00:04:58  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.11-SNAPSHOT:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.equinox.launcher.win32.win32.aarch64: Only qualifier changed for (org.eclipse.equinox.launcher.win32.win32.aarch64/1.2.1100.v20250202-2240). Expected to have bigger x.y.z than what is available in baseline (1.2.1100.v20240722-2106) -> [Help 1]

jonahgraham:
The failing test (`testCoordinatedConfigurationOnBeforeRegisteredManagedService (org.osgi.test.cases.cm.junit.CMCoordinationTestCase) failed`) seems to be https://github.com/eclipse-equinox/equinox/issues/190 so I think that failure can be ignored and I don't think anything in this PR could affect this test anyway.

laeubi:
Yes the TCK test failure is expected see:
- https://github.com/osgi/osgi/issues/718

features/org.eclipse.equinox.executable.feature/library/eclipse.c
```diff
@@ -445,6 +445,20 @@ JNIEXPORT int run(int argc, _TCHAR* argv[], _TCHAR* vmArgs[])
...
 
+#ifdef LINUX
+	if (!noSplash) {
+		char *swtGtk4 = getenv("SWT_GTK4");
+		if (swtGtk4 != NULL && strcmp(swtGtk4, "1") == 0) {
+			// The eclipse launcher uses GTK3 exclusively, therefore when starting GTK4
+			// it must be started without a splash screen
+			// https://github.com/eclipse-equinox/equinox/issues/831
+			// If the user didn't already specify -nosplash print a warning and turn the splashscreen off
+			_ftprintf(stderr, "WARNING: SWT_GTK4 does not support splash screen yet. Therefore it has been disabled. To suppress this message launch with -nosplash\n");
```
HannesWell, line 456: Would it be possible to support splash-screens with GTK4 or is this too complicated?

jonahgraham: It will be - this is a temporary state to resolve #831. To do splash screens on GTK4 we need to use GTK4 to start the splash on the native side. That means the launcher will need GTK4 soon.

What I think we will need to do is similar to what we do in SWT. I created #836 for adding the real version of the splash screen for GTK4.

jonahgraham: PS I was looking for a simple patch to write for eclipse launcher to make sure I understood build flow. If this change seems unnecessary, then we can also just close it and work on the correct GTK4 splash screens at a suitable time in the future.

laeubi (approved):
It is better to have no splash screen until we have GTK4 support instead of failing completely.

laeubi:
@jonahgraham thanks for working on this, I think we have even one more case here:

If the application crash (e.g. exit with an exception in the main, JDK can not be loaded and so on) the the launcher shows a popup (see option `--launcher.suppressErrors`) I suspect this is using GTK too and would need a similar guard.

jonahgraham:
> @jonahgraham thanks for working on this, I think we have even one more case here:
> 
> If the application crash (e.g. exit with an exception in the main, JDK can not be loaded and so on) the the launcher shows a popup (see option `--launcher.suppressErrors`) I suspect this is using GTK too and would need a similar guard.

Please see https://github.com/eclipse-equinox/equinox/issues/836#issuecomment-2639875919 I specifically didn't put a guard on error messages because I thought it best that they still appear even is user does SWT_GTK4=1

PR #839, opened 2025-02-11
Enable region tests

laeubi:


laeubi:
@tjwatson the `RegionPerformanceTests` seem to fail mostly because of 

```
A bundle is already installed with the name "PackageProvider1" and version "1.0.0"
org.osgi.framework.BundleException: A bundle is already installed with the name "PackageProvider1" and version "1.0.0"
	at org.eclipse.osgi.container.ModuleContainer.install(ModuleContainer.java:480)
	at org.eclipse.osgi.storage.Storage.install(Storage.java:755)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.installBundle(BundleContextImpl.java:183)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.installBundle(BundleContextImpl.java:176)
	at org.eclipse.equinox.region.tests.BundleInstaller.installBundle(BundleInstaller.java:55)
	at org.eclipse.equinox.region.tests.BundleInstaller.installBundle(BundleInstaller.java:48)
	at org.eclipse.equinox.region.tests.system.RegionPerformanceTests.setUp(RegionPerformanceTests.java:42)
	at junit.framework.TestCase.runBare(TestCase.java:140)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at junit.framework.TestSuite.run(TestSuite.java:236)
```

so it seems it either misses cleanup or should check if a bundle is already there.

The aspect4j seems to be required for the embedded virgo tests, if I locally remove that a large part of the test are affected so it seems one would need a replacement for that mocking (maybe https://sling.apache.org/documentation/development/osgi-mock.html).

laeubi:
A maybe better way would be to finally add OSGi Test support to PDE...

- https://github.com/eclipse-pde/eclipse.pde/issues/877

then we don't need the mocking here... on the other hand I wonder why we need to mock OSGi at all? Can't the test just run with a connect framework or similar?

PR #842, opened 2025-02-13
Fix org.eclipse.equinox.slf4j bundle localization and build properties

iloveeclipse:
Small cosmetic change to show proper bundle name in installer dialog & to fix warning in build properties


laeubi:
Verification build will not test this

PR #844, opened 2025-02-13
Use import-package instead of require-bundle

laeubi:
So it can work with "older" eclipse release used in Tycho-Api tools


.mvn/maven.config
```diff
@@ -1,3 +1,3 @@
--Pbuild-individual-bundles
+-Pbuild-individual-bundles,!api-check
```
laeubi, line 1: Needs to be reverted after build verified.

PR #895, opened 2025-03-09
README: Add link to new documentation website and fixed homepage link

chrisrueger:
- added Documentation: https://equinox.eclipseprojects.io/
- fixed Homepage link, which gave a 404 before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Equinox implements the follwoing specification with the given level of complianc
 
 # More information
 
-- Homepage: https://www.eclipse.org/equinox/
+- Homepage: https://equinox.eclipseprojects.io/
 - Bug tracker: https://github.com/eclipse-equinox/equinox/issues
 - Asking questions and share ideas: https://github.com/eclipse-equinox/equinox/discussions
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,12 +1,10 @@
 # Equinox Documentation
 
-This is the documentation of the Eclipse Equinox Framework.
-[https://github.com/eclipse-equinox/equinox](https://github.com/eclipse-equinox/equinox)
+This is the documentation of the [Eclipse Equinox Framework](https://projects.eclipse.org/projects/eclipse.equinox).
 
 Equinox is an implementation of the OSGi core framework specification, a set of bundles that implement various optional OSGi services and other infrastructure for running OSGi-based systems.
 
-PRs to improve the documentation are always welcome.
-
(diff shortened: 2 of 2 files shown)


laeubi:
@chrisrueger many thanks, by the way we have a snapshot of the old site for example here:

http://web.archive.org/web/20240720181650/https://eclipse.dev/equinox/

Maybe there is some useful information we should preserve for the new site? @tjwatson ?

merks:
I think it's a bug that these don't work. 

- https://www.eclipse.org/equinox/
- https://eclipse.dev/equinox/

I assume it's because of the disabled PHP and that this is still being served and blocked:

https://github.com/eclipse-equinox/equinox-website/blob/master/index.php

But when I edit here:

https://projects.eclipse.org/projects/eclipse.equinox/edit

image

I imagine we'll want to change the project metadata to point to this as well:

https://equinox.eclipseprojects.io/

I just expect both the above 2 links to redirect to what's specified in the project metadata:

image

I've opened this issue:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/5722

chrisrueger:
> I think it's a bug that these don't work.

@laeubi @merks ah ok. So let me know if I should adjust some links in the PR  then.

chrisrueger:
@merks Maybe it is a general problem? also https://www.eclipse.org/eclipse gives a 404 which is linked from the Eclipse JDT main README https://github.com/eclipse-jdt

merks (commented):
Firstly thanks for taking action!!

Technically this is the project's home page, which does work now:

https://eclipse.dev/equinox/

But it redirects to 

https://equinox.eclipseprojects.io/

So probably best we link directly to that.  I do think that the above is the home page and I don't know that the following should be highlighted as the home page:

https://projects.eclipse.org/projects/eclipse.equinox

It doesn't really have much useful information...

That's just my personal opinion...

What do @tjwatson and @laeubi  think?

chrisrueger:
> https://equinox.eclipseprojects.io/
> 
> So probably best we link directly to that

Thanks @merks I adjusted the PR and just point the Homepage to https://equinox.eclipseprojects.io/

merks (approved):
It looks good to me.

waynebeaton:
> form org point of view we probably should have had release records there for each release

The EMO's current thinking is that release information should be where the community expects to find it.

https://gitlab.eclipse.org/eclipsefdn/it/websites/projects.eclipse.org/-/issues/146

merks:
I like the proposal the way it is.   I think if we want a link to the portal, we should provide it from the new home page...

PR #916, opened 2025-03-25
Fix branding images in Oomph setups

merks:


merks:
This has no build impact.  Merging immediately.

PR #921, opened 2025-03-31
Fix bugs in the file-search logic at Equinox launcher Main

HannesWell:
1. When searching the library-path with an explicit bootLocation available not always the fragment from the current directory was considered. Instead the first one found was reused for all other directories.
2. When searching for a target directory the final result was resolved against the given start directory instead of the resolved search root, where the target-directory was actually found.

### DIFF
--- a/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
+++ b/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
@@ -506,7 +506,6 @@ public class Main {
 
 	private String getLibraryPath(String fragmentName, URL[] defaultPath) {
 		String libPath = null;
-		String fragment = null;
 		if (inDevelopmentMode && devClassPathProps != null) {
 			String devPathList = devClassPathProps.getProperty(PLUGIN_ID);
 			String[] locations = getArrayFromList(devPathList);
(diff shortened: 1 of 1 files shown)


bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1089,7 +1086,7 @@ protected String searchFor(final String target, String start) {
 		if (result == -1) {
 			return null;
 		}
-		File candidate = new File(start, names[result]);
+		File candidate = new File(root, names[result]);
```
HannesWell, line 1089: @tjwatson is it actually intended here to return the unresolved `start` directory?

tjwatson: I don't think so, that does look wrong to me.  Particularly because we then try to get the absolute path to the unresolved start as the returned result.  How is this working now?

HannesWell: My guess is that the `start` file is actually never resolved because it's absolute.
But I haven't tried to reproduce it.

HannesWell:
@tjwatson are you then fine with this change in total?

PR #924, opened 2025-04-13
Clean-up Equinox Launcher Main and make all fields and methods private

HannesWell:
Also remove unused launcher constants.

@tjwatson is it fine to make all these fields, methods and inner classes private or are there any known external users of this?


HannesWell:
@tjwatson do you want to have a look at this?
If I don't hear anything from you till tomorrow evening (Friday Europen evening), I'll consider it as a silent consent.

HannesWell:
I have now kept the `StartupClassLoader` class public just in case some special users somehow have to access the startup class loader reflectively. I believe you have mentioned something in that direction in the past, but maybe that's not related at all.

Once the build completes I plan to submit this.

PR #926, opened 2025-04-19
Simplify argument processing in Equinox Launcher Main

HannesWell:


bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -138,8 +140,8 @@ public class Main {
 	private String exitData = null;
 
 	private String vm = null;
-	private String[] vmargs = null;
-	private String[] commands = null;
+	private List<String> vmargs = null;
+	private List<String> commands = null;
```
merks, line 144: Maybe during cleaning, pointless explicit initialization to the default value could be removed too…

HannesWell:
If there are no further remarks on this, I plan to submit this tomorrow (European) evening.

PR #936, opened 2025-05-03
Change NLS.bind(String, Object[]) to varargs

HannesWell:
This avoids the explicit array creation when invoking this method.

I wonder if change is binary compatible, but I didn't find anything in the JLS about _varargs_ or _ellipsis_:
https://docs.oracle.com/javase/specs/jls/se21/html/jls-13.html

Looking at the produced byte-code I only see
```
  // access flags 0x9
  public static bind(String, Object[]) : String
   L0
    LINENUMBER 123 L0
    ALOAD 0: message
...
```
becoming
```
  // access flags 0x89
  public static varargs bind(String, Object[]) : String
   L0
    LINENUMBER 123 L0
    ALOAD 0: message
...
```
with this change. So the only change is the additional `varargs` keyword.
@iloveeclipse can you tell if this would be binary compatible or where one could find this out?

From a source compatibility POV I only see errors in the `EquinoxContainer`, which is fixed along side (but was actually unnecessary) and in
https://github.com/eclipse-platform/eclipse.platform/blob/86aa0ac329edcb4246009fe19b4ac0d680d4c491/team/bundles/org.eclipse.jsch.ui/src/org/eclipse/jsch/internal/ui/KeyboardInteractiveDialog.java#L87

But that could be fixed easily and was also overly complex.

### DIFF
--- a/bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/util/NLS.java
+++ b/bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/util/NLS.java
@@ -119,7 +119,7 @@ public abstract class NLS {
 	 * @return the manipulated String
 	 * @throws IllegalArgumentException if the text appearing within curly braces in the given message does not map to an integer
 	 */
-	public static String bind(String message, Object[] bindings) {
+	public static String bind(String message, Object... bindings) {
 		return internalBind(message, bindings, null, null);
 	}
 
--- a/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/internal/app/EclipseAppHandle.java
+++ b/bundles/org.eclipse.equinox.app/src/org/eclipse/equinox/internal/app/EclipseAppHandle.java
@@ -7,7 +7,7 @@
  * https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
- * 
+ *
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Josh Arnold - Bug 180080 Equinox Application Admin spec violations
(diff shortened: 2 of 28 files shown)


bundles/org.eclipse.osgi/supplement/src/org/eclipse/osgi/util/NLS.java
```diff
@@ -119,7 +119,7 @@ public static String bind(String message, Object binding1, Object binding2) {
 	 * @return the manipulated String
 	 * @throws IllegalArgumentException if the text appearing within curly braces in the given message does not map to an integer
 	 */
-	public static String bind(String message, Object[] bindings) {
+	public static String bind(String message, Object... bindings) {
```
laeubi, line 122: Is this binary compatible change?

HannesWell: That's the big question described initially. :)

laeubi: > @laeubi Could we use the cleanup jobs to remove the then unnecessary Array creation?

If there is a cleanup action available it only needs to be enabled. 

For this I do the following:

1. create a new cleanup profile and export it to a file
2. then toggle the option of interest and export it to another file
3. now select both files and choose compare with each other
4. Then one knows the settings and only need to add them [here](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/5ea5bc8009b35479c891b367805563b08ff02a34/eclipse-platform-parent/pom.xml#L566)
5. Then waiting for the next build (happens once a day) and it will create PRs, then one needs to carefully review the changes and report any issues to JDT if any so they can be fixed, when the next i-build is there one can rerun the job or wait for the next day.

HannesWell: Will do that, thanks.

PR #938, opened 2025-05-04
Improve check for visiting same child in a status cycle

merks:
- We must prevent endless recursion if there is a cycle in status children, but we shouldn't prevent the same child from appearing more than once in the tree.


merks:
@iloveeclipse  @tjwatson 

Given the short/late runway for M2, the triviality of the change, and the fact that it's a recent regression, I plan to merge this at some point **today**.

merks:
> Thanks for the quick fix. I will update the Equinox test for this so it is caught in Equinox PR builds instead of only the platform builds

I'm here to help 😁  Yes, a test would be a good thing. 👼

PR #939, opened 2025-05-05
Make var-args overload of NLS.bind() generic

HannesWell:
This avoids a compiler warning if an explicit array of a sub-type of Object[] is passed as last argument:
`NLS.bind("message: {0} {1}", new String[] { "Hello", "World" })`

The byte-code generated for the `NLS.bind()` method and for all its callers is not changed by this adjustment. It just avoids what looks like a source-compatibility issues with the previous introduction of var-args in `NLS.bind()`. Although the example above works the same in all three states.

Follow-up on:
- https://github.com/eclipse-equinox/equinox/pull/936


HannesWell:
This should avoid the warnings fixed in:
- https://github.com/eclipse-platform/eclipse.platform/pull/1856

and I have also checked that the errors (which are actually just elevated compiler-warnings) from https://github.com/eclipse-platform/eclipse.platform/pull/1854, would not have occurred with this.

In general these changes should still be applied as they simplify code that is currently needlessly verbose, but I think it's better without warnings/errors to keep existing callers more comfortable.

tjwatson (approved):

HannesWell:
The failing quality-gate was indeed relevant as some calling projects that now get slight different byte-code generated because the var-args array generated by the compiler now can have a more specific type than `Object[]`, common for all passed arguments.

PR #953, opened 2025-05-12
Avoid registry lock when calling hooks

tjwatson:
There are currently 9 hook types the framework calls. Some hooks get called large amounts of time.  For example, Service event hooks during service operations, or weaving hooks during class loads.

The framework should cache the various hook registrations instead of doing a full registry lookup which requires the singleton registry lock.

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceRegistry.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceRegistry.java
@@ -44,6 +44,7 @@ import org.eclipse.osgi.internal.framework.BundleContextImpl;
 import org.eclipse.osgi.internal.framework.EquinoxContainer;
 import org.eclipse.osgi.internal.framework.FilterImpl;
 import org.eclipse.osgi.internal.messages.Msg;
+import org.eclipse.osgi.internal.serviceregistry.ServiceRegistrationImpl.FrameworkHookRegistration;
 import org.eclipse.osgi.internal.serviceregistry.ServiceUse.ServiceUseLock;
 import org.eclipse.osgi.storage.BundleInfo.Generation;
 import org.eclipse.osgi.util.NLS;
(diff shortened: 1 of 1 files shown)


laeubi (commented):
Maybe instead of individual fields it would be good to use a `ConcurrentHashMap<Class, List>`... if the values of that map are then of type `CopyOnWriteArrayList` you should be able to access all datastructures in a concurrent way without any synchronization. The cost for adding/removing items is a bit higher but I think we can neglect this here as there are usually not much items in a list and they not change very often.

tjwatson:
> Maybe instead of individual fields it would be good to use a `ConcurrentHashMap<Class, List>`... if the values of that map are then of type `CopyOnWriteArrayList` you should be able to access all datastructures in a concurrent way without any synchronization. The cost for adding/removing items is a bit higher but I think we can neglect this here as there are usually not much items in a list and they not change very often.

Maybe, but I don't see the need to have a `ConcurrentHashMap` with `CopyOnWriteArrayLists` since I would not change the key/value objects to new values.  Instead it would be primed the map with a constant set of `CopyOnWriteArrayList` and then just add/remove from the existing list.  But I don't want to do that because then I would still need to "snapshot" the list each use because the various hooks in the framework depend on that list to remain constant while they are in use.  There may be some hooks that require multiple iterations over the list of hooks and we don't want any chance of that iteration changing with a `CopyOnWriteArrayList`

If we don't use a `CopyOnWriteArrayList` then we do need something like `ConcurrentHashMap` that resets the lists to new a new List each time a hook is added/modified/removed.

laeubi:
You can of course always set the list, I just use `CopyOnWriteArrayList` when dynamically adding/removing services, but if the list has to remain constant it might be better to do it explicitly.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/serviceregistry/ServiceRegistry.java
```diff
@@ -310,6 +313,15 @@ private List<Class<?>> getHookClass(String className, List<Class<?>> hookTypes)
 		}
 	}
 
+	private void setHookRegistrations(String hookClass, List<ServiceRegistrationImpl<?>> hooks) {
+		hooks = hooks == null || hooks.isEmpty() ? Collections.emptyList() : new ArrayList<>(hooks);
+		frameworkHooks.put(hookClass, hooks);
```
laeubi, line 318:
```suggestion
		frameworkHooks.put(hookClass, hooks == null ? List.of() : List.copyOf(hooks));
```

laeubi: > I assume you are trolling me to use post Java 8+ methods :)

It seems its too long since I'm writing against Java 8 API :-\

ArrayList seems fine then if we assume its only used internally and we always using it "right".

PR #1002, opened 2025-05-22
Add option to ban weaving hooks

tjwatson:


laeubi:
@tjwatson what do you think about making it more generic currently we have `bannedWeavingBundles` hard coded but it could better be a way of property that defines either a whitelist or blacklist off hooks?

tjwatson:
> @tjwatson what do you think about making it more generic currently we have `bannedWeavingBundles` hard coded but it could better be a way of property that defines either a whitelist or blacklist off hooks?

I went ahead and polished off the PR.  We can decide to merge this after the release as a safeguard in the future.

laeubi:
@tjwatson as master is open now should we merge it?

tjwatson:
> @tjwatson as master is open now should we merge it?

Sure, I merged it.  Now we have an out if we want to configure a banned hook bundle, with the option of a version range for the banned bundle.

PR #1006, opened 2025-06-04
Add version bumps for 4.37

HannesWell:
Provoke version increment for 4.37.

Part of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3093

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -391,3 +391,5 @@ pipeline {
 		}
 	}
 }
+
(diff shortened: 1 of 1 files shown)


HannesWell:
In fact https://github.com/eclipse-equinox/equinox/pull/1005 has already included all necessary version-increments and this is not necessary. I just expected more, but when comparing it with previous releases, it's fine.

PR #1007, opened 2025-06-05
[Build] Use Eclipse Equinox Bot to commit built launcher binaries

HannesWell:
Use the `Eclipse Equinox Bot` to commit built launcher binaries with its correct email as recorded in the EF-API (https://api.eclipse.org/bots ) and that is also associated with it's Github account. Before the `Eclipse Releng Bot` was used for that, but with an incorrect email.

Not it's the same bot that is also used for version increments:
https://github.com/eclipse-equinox/equinox/blob/84219045d0efe0124a4bfc1a11ebf5d8625d28b8/.github/workflows/pr-checks.yml#L20-L21

I think that is more suitable than using the Releng Bot (even with a fixed mail).

Fixes
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6227

@tjwatson, @laeubi any objection?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -111,7 +111,7 @@ pipeline {
 					script {
 						def authorMail = sh(script: 'git log -1 --pretty=format:"%ce" HEAD', returnStdout: true)
 						echo 'HEAD commit author: ' + authorMail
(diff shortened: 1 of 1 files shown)


HannesWell:
I consider the absence of objections as consent. :)

PR #1011, opened 2025-06-06
Use Maven 3.9.10 in GHAs

akurtakov:


akurtakov:
Test failure unrelated.

PR #1038, opened 2025-06-18
[Build] Refine check for launcher binary rebuild commit

HannesWell:
Currently the build is aborted if the branch tip is at a commit, which was committed with the email of the 'Eclipse Equinox Bot'. This is done to prevent rebuilds after a commit for a launcher binary rebuild was pushed by the bot. But because that bot is also used for other kind of commits, this test leads to false positives.

This reworks the pipeline to check if the head commit has a launcher-binaries-build tag attached which is a more precise indicator.


HannesWell:
Another option I'm about to investigate is if the test for a launcher-binary build can just be skipped entirely. After analyzing the pipeline again, pushing the update commit after a launcher binary rebuild just should lead to one new build (as usual on a push to master). That second build is actually not necessary, but we shouldn't enter a infinite native re-build loop as initially feared.

HannesWell:
> Another option I'm about to investigate is if the test for a launcher-binary build can just be skipped entirely. After analyzing the pipeline again, pushing the update commit after a launcher binary rebuild just should lead to one new build (as usual on a push to master). That second build is actually not necessary, but we shouldn't enter a infinite native re-build loop as initially feared.

The testing demonstrated that this works as expected/hoped and the pipeline in fact doesn't not end up in an endless native rebuild loop in case that check is removed. Only one new build triggered, when a commit for a launcher binary rebuild is pushed (as usual for a push), which does not rebuild the native binaries again (because the latest launcher-binaries-build tag points to the then HEAD commit).
So as the binaries are only changed seldomly I'll continue with that option as this also simplifies the pipeline in general.

HannesWell:
> > Another option I'm about to investigate is if the test for a launcher-binary build can just be skipped entirely. After analyzing the pipeline again, pushing the update commit after a launcher binary rebuild just should lead to one new build (as usual on a push to master). That second build is actually not necessary, but we shouldn't enter a infinite native re-build loop as initially feared.
> 
> The testing demonstrated that this works as expected/hoped and the pipeline in fact doesn't not end up in an endless native rebuild loop in case that check is removed. Only one new build triggered, when a commit for a launcher binary rebuild is pushed (as usual for a push), which does not rebuild the native binaries again (because the latest launcher-binaries-build tag points to the then HEAD commit). So as the binaries are only changed seldomly I'll continue with that option as this also simplifies the pipeline in general.

Finally I got a full run that confirms this, so I consider this ready for submission.

PR #1061, opened 2025-06-28
Fix some generics warnings o.e.equinox.app

akurtakov:


akurtakov:
New warnings are actually moved ones. There is net reduction of 17.
image

Issue #843, opened 2025-02-13
EquinoxLogger extends deprecated class org.slf4j.helpers.MarkerIgnoringBase


iloveeclipse:
In the new org.eclipse.equinox.slf4j bundle I see deprecation warning that new class `EquinoxLogger` extends deprecated class `org.slf4j.helpers.MarkerIgnoringBase`.

I don't think it is OK to introduce new code that depends on deprecated stuff, however I have no idea which replacement is supposed to be used as there is no proper documentation why it is deprecated.

iloveeclipse:
@laeubi, FYI.

laeubi:
Yes it was deprecated in 1.x but replacement is only available in 2.x

iloveeclipse:
> Yes it was deprecated in 1.x but replacement is only available in 2.x

Now that the dependency switched from 1.x to 2.x, it can be fixed, right? What is the replacement btw?

laeubi:
> > Yes it was deprecated in 1.x but replacement is only available in 2.x
> 
> Now that the dependency switched from 1.x to 2.x, it can be fixed, right?

Yes I just wanted to first make the build / api-tools work again ;-)

> What is the replacement btw?

It is not really documented in the API docs but I think `org.slf4j.helpers.AbstractLogger` should be the one to use, that also requires to only override one method.

PR #1063, opened 2025-07-04
Update MarkerIgnoringBase with AbstractLogger

SougandhS:
This commit replaces org.slf4j.helpers.MarkerIgnoringBase with org.slf4j.helpers.AbstractLogger implementation

based on : https://github.com/eclipse-equinox/equinox/issues/843#issuecomment-2655675412

Fixes : https://github.com/eclipse-equinox/equinox/issues/843

### DIFF
--- a/bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
+++ b/bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
@@ -10,14 +10,17 @@
  *
  * Contributors:
  *     Christoph Läubrich - initial API and implementation
+ *     IBM Corporation -  Update deprecated logger 
  *******************************************************************************/
 package org.eclipse.equinox.slf4j;
 
 import org.osgi.service.log.Logger;
+import org.slf4j.Marker;
+import org.slf4j.event.Level;
 import org.slf4j.helpers.FormattingTuple;
 import org.slf4j.helpers.MessageFormatter;
 
-class EquinoxLogger extends org.slf4j.helpers.MarkerIgnoringBase {
+class EquinoxLogger extends org.slf4j.helpers.AbstractLogger {
 
 	private static final long serialVersionUID = 1L;
 	private final Logger logger;
@@ -240,6 +243,41 @@ class EquinoxLogger extends org.slf4j.helpers.MarkerIgnoringBase {
 			FormattingTuple tp = MessageFormatter.arrayFormat(format, argArray);
 			logger.error(tp.getMessage(), tp.getThrowable());
 		}
+	}
+
+	@Override
+	public boolean isTraceEnabled(Marker marker) {
+		return isTraceEnabled();
+	}
+
+	@Override
+	public boolean isDebugEnabled(Marker marker) {
+		return isDebugEnabled();
+	}
+
+	@Override
+	public boolean isInfoEnabled(Marker marker) {
+		return isInfoEnabled();
+	}
+
+	@Override
+	public boolean isWarnEnabled(Marker marker) {
+		return isWarnEnabled();
+	}
+
+	@Override
+	public boolean isErrorEnabled(Marker marker) {
+		return isErrorEnabled();
+	}
+
+	@Override
+	protected String getFullyQualifiedCallerName() {
+		return this.name;
+	}
+
+	@Override
+	protected void handleNormalizedLoggingCall(Level level, Marker marker, String messagePattern, Object[] arguments,
+			Throwable throwable) {		
 	}
 
 }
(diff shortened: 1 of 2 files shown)


SougandhS:
Hi @laeubi, could you please check this PR when you get some time ?

laeubi (changes requested):

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
```diff
@@ -242,4 +245,39 @@ public void error(String format, Object... argArray) {
...
+		return isErrorEnabled();
+	}
+
+	@Override
+	protected String getFullyQualifiedCallerName() {
+		return this.name;
+	}
+
+	@Override
+	protected void handleNormalizedLoggingCall(Level level, Marker marker, String messagePattern, Object[] arguments,
```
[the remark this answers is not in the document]
SougandhS: Thanks, I have updated it now

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
```diff
@@ -32,213 +34,77 @@ public boolean isTraceEnabled() {
...
+			return;
 		}
-	}
-
-	@Override
-	public void error(String format, Object arg1) {
-		if (isErrorEnabled()) {
-			FormattingTuple tp = MessageFormatter.format(format, arg1);
-			logger.error(tp.getMessage(), tp.getThrowable());
+		String formattedMessage = MessageFormatter.basicArrayFormat(messagePattern, arguments);
```
laeubi, line 92: One should only lazy create the formatted message, I would simply inline that call so it is not created when the log level is disabled.

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
```diff
@@ -32,213 +34,77 @@ public boolean isTraceEnabled() {
...
 	public boolean isWarnEnabled() {
 		return logger != null && logger.isWarnEnabled();
 	}
 
-	@Override
-	public void warn(String msg) {
-		if (isWarnEnabled()) {
-			logger.warn(msg);
-		}
+	public boolean isErrorEnabled() {
```
laeubi, line 52: Missing `@Override`?

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
```diff
@@ -32,213 +34,77 @@ public boolean isTraceEnabled() {
...
+		return logger != null && logger.isErrorEnabled();
 	}
 
-	@Override
-	public void warn(String msg, Throwable t) {
-		if (isWarnEnabled()) {
-			logger.warn(msg, t);
-		}
+	public boolean isTraceEnabled(Marker marker) {
```
laeubi, line 56: Missing `@Override`?

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
```diff
@@ -32,213 +34,77 @@ public boolean isTraceEnabled() {
...
+		return isErrorEnabled();
 	}
 
 	@Override
-	public void error(String msg) {
-		if (isErrorEnabled()) {
-			logger.error(msg);
-		}
+	protected String getFullyQualifiedCallerName() {
+		return this.name;
```
laeubi, line 82: The [SimpleLogger](https://github.com/qos-ch/slf4j/blob/183aaa507040ce6c61f70762c13e7d11aa4fd54e/slf4j-simple/src/main/java/org/slf4j/simple/SimpleLogger.java#L471-L473) only return `null` here, as there is no javadoc what this method is used for one might either check other Slf4J implementations or also just return null instead of the logger name.

laeubi (changes requested):

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
SougandhS: Accidental removal, updated 👍

bundles/org.eclipse.equinox.slf4j/src/org/eclipse/equinox/slf4j/EquinoxLogger.java
SougandhS: Updated 👍

SougandhS:
Thank you @laeubi

SougandhS:
> Thanks for taking care of this, please let us know if you see other improvements!

sure 👍

PR #1091, opened 2025-07-20
Remove unused imports and use dynamic import instead

laeubi:
Currently `com.sun.el` is detected as unused because we only use it implicitly in BundleProxyClassloader. On the other hand there we check if a class starts with the package name so the package itself might not be enough as this can match much more classes than we import.

This now aims to resolve the issue by using DynamicImport-Package

@merks according to the history you did added the import, can you remember if it was explicitly required to use an explicit import instead of a dynamic one? Do we maybe only require classes from `com.sun.el`  but nothing else or even a specific one?


laeubi:
I even tried to find out what classes are loaded here but it seems with a usual Eclipse HelpCenter I can't trigger any classlaods. Any hint what feature require this? I can see the package `javax.el` is requested (even though this would not load any class?) but nothing from `com.sun`.

laeubi:
New compiler issues are due to changed JDT
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4210

PR #1094, opened 2025-07-21
Throw exception when last candidate is removed and mandatory requirement

laeubi:
Currently it can happen that the resolver is removing a candidate but it is the last possible one to select. If the requirement is not optional this lead guaranteed to an unresolvable state and must be handled already on the call site.

This adds throwing a runtime exception for this case to make sure such illegal modifications do not goes unnoticed and we can fix such places.

Only a draft as it would be currently too dangerous to be merged but hopefully will fail some tests to further analyze the problem.


laeubi:
I was now able to even further reduce the testcase.

laeubi:
I think I now understand what is happening here, we start for the testcase with this rather simple setup

```
[ aQute.libg ]
  [?]aQute.libg 7.1.0.202411251545 (UNRESOLVED))
    [?]Import-Package: aQute.bnd.exceptions; version="[3.0.0,4.0.0)": 
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="aQute.libg"; bundle-version="7.1.0.202411251545"; version="3.0.0"
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="biz.aQute.bnd.util"; bundle-version="7.1.0.202411251545"; version="3.0.0"

[ biz.aQute.bnd.util ]
  [?]biz.aQute.bnd.util 7.1.0.202411251545 (UNRESOLVED))
    [?]Import-Package: aQute.bnd.exceptions; version="[3.0.0,4.0.0)": 
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="aQute.libg"; bundle-version="7.1.0.202411251545"; version="3.0.0"
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="biz.aQute.bnd.util"; bundle-version="7.1.0.202411251545"; version="3.0.0"
        
[ biz.aQute.bndlib ]        
  [?]biz.aQute.bndlib 7.1.0.202411251545 (UNRESOLVED))
    [?]Import-Package: aQute.bnd.exceptions; version="[3.0.0,4.0.0)": 
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="aQute.libg"; bundle-version="7.1.0.202411251545"; version="3.0.0"
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="biz.aQute.bnd.util"; bundle-version="7.1.0.202411251545"; version="3.0.0"
    [!]Import-Package: aQute.bnd.result; version="[2.0.0,3.0.0)": 
        Export-Package: aQute.bnd.result; bundle-symbolic-name="biz.aQute.bnd.util"; bundle-version="7.1.0.202411251545"; version="2.0.0"; uses:="aQute.bnd.exceptions"
```

Because of the uses constraint we can reduce `biz.aQute.bndlib` to the following:
```
[ Reduced biz.aQute.bndlib ]
  [!]biz.aQute.bndlib 7.1.0.202411251545 (UNRESOLVED))
    [!]Import-Package: aQute.bnd.exceptions; version="[3.0.0,4.0.0)": 
        Export-Package: aQute.bnd.exceptions; bundle-symbolic-name="biz.aQute.bnd.util"; bundle-version="7.1.0.202411251545"; version="3.0.0"
    [!]Import-Package: aQute.bnd.result; version="[2.0.0,3.0.0)": 
        Export-Package: aQute.bnd.result; bundle-symbolic-name="biz.aQute.bnd.util"; bundle-version="7.1.0.202411251545"; version="2.0.0"; uses:="aQute.bnd.exceptions"
```

this is because `bndlib` import package `aQute.bnd.result` that has only a single provider that has a uses constraint on `aQute.bnd.exceptions` and therefore we must throw away the `aQute.libg` provider.

Now the resolver begins and come to the conclusion that `aQute.libg` has a substitution package and decides to resolve it using the _INTERNAL_ rule that means it tries to discard from all dependents until it finds its own export, but when like in this case, the provider `aQute.libg` was already dropped, this means it will _effectively drop everything_ until nothing is there anymore and then the resource `biz.aQute.bndlib` will not have any providers left and is considered unresolved.

So I think this assumption taken here is only ever valid in special cases where the provider of the package was not previously ruled out, I also don't see why we should remove any *higher ranked providers* ... so this part must be changed to ignore/remove only the one provider regardless of where it is coming in the list! 
This is currently not really possible due to the data-model.

laeubi:
While working on this I came to the conclusion that there is another case where the Bundle `A` has package `X` and it uses this initial ordering:

- X
  - B
  - A (our own candidate)
  - C

So while we have selected `B` for `X` the export must be discarded. When we choose `A` for `X` we need to keep it and then for `C` we need to again discard the export. It therefore seems necessary to be able to determine the substitution state we not need only two permutations but three. But one could combine all cases where the export is dropped, its just not 100% clear to me if we then still retain priority order and it also mean we need to perform this once globally what would result in much candidate objects to be created.

laeubi:
This now has lead me back to `org.apache.felix.resolver.Candidates.canRemoveCandidate(Requirement)` what also previously have lead to some problems but I didn't wanted to touch it back then. So if we can solve the permutation problem this method is likely not needed anymore together with all the lockups associated with substitution packages.

My rough idea here would be as follows:

- Whenever we pop a permutation from the stack we call a method to resolve substitution-packages
- For this to work, we need a list of all `CandidateSelector`s that are substitution packages (that is similar to what `org.apache.felix.resolver.Candidates.populateSubstitutables()` currently does)
- We then need to iterate over all substitution packages and do the following
  - If there are more than one candidate we need to create a permutation to not loose any alternative and add it to the stack
  - purge all other providers from the list to make this unique and prevent it from being further permutated (as we already have put such permutation on the stack in case there are more than one).
  - If the current candidate resolves to the same resource as the requirement we have the _Internal_ case and are done
  - Otherwise we need to purge the Provider form all other resources list of alternatives to account for the _External_ case.

This then has the advantage that many things become unique and we can even rule out other alternatives. My plan would be to make this extracted into an own class, so we can have an emergency switch in case something goes really wrong after the release as is is a quite fundamental change.

laeubi:
Some improvements has already been merged and the testcase is added as part of:
- https://github.com/eclipse-equinox/equinox/pull/1119

PR #1101, opened 2025-07-27
Remove source features and source bundles from the sdk features

merks:
- The repositories and products that use these SDK features use includeAllSources to ensure that sources are available.

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3206


merks:
Admin TCK always fails.

PR #1105, opened 2025-07-29
Do not add the inital candidate to the session

laeubi:
Currently the inital candidate is added to the session as us USES permutation. This is not only confusing because it is not a permutation of the initial problem and not caused by a sue constraint violation, this also gives false statistics reported as always one uses permutation is assumed even if the problem can be solved without any permutation.

To prevent his and to allow further improvements that can act at the time a permutation is really removed from the stack, this now passes the initial permutation to the method that uses it.


laeubi:
Seems this needs smaller increments to see whats going wrong ...

PR #1110, opened 2025-07-29
Improve substitution package handling

laeubi:
Currently substitution packages are handles in a way that can leads to situation where a permutation explosion occurs and even unresolved states are produced. To prevent this we have a method that checks if a requirement can be removed to prevent such removal but this then leads to use-constraint violation and even more permutations computed that are guaranteed to fail in many cases.

This now is improved by the following changes:

- when a permutation is removed from the stack, every requirement is checked if it is a substitution package and if it is one the subtitution is resolved by either being external or internal
- if other candidates exits for this they are permutated once and then dropped to prevent any further permutation as the choice can not be reverted and therefore other permutations in this path will be invalid
- now if the requirement has only a single provider (what can happen independently or as part of the resolving to either external or internal) the use constraints are checked if they are in conflict with a currently selected provider, in which case the invalid choice is dropped
- this can lead to the situation that a requirement become effectively unresolved (that means it has no providers) and if we detect this case we first choose other permutations until we found the one with the most resolved resources requirements
- only after that we check the package space consistency (that is a far more expensive check) what probably will result in more permutations created and checked until we found the best result to use

## This is currently work in progress


laeubi:
Will be continued here:
- https://github.com/eclipse-equinox/equinox/pull/1124

PR #1113, opened 2025-07-31
Do not return null from ResolverImpl#findValidCandidates

laeubi:
Currently the `ResolverImpl#findValidCandidates` can return `null` when no more permutations are found. The calling code is not really prepared to handle that and its likely not a "valid" candidate anyways.

This now makes sure that `null` is never returned from the function by requiring a first non-null candidate to be present, then when encountering no more candidates simply return the last non null item.


laeubi:
Included here now:
- https://github.com/eclipse-equinox/equinox/pull/1114

PR #1115, opened 2025-07-31
Pull out package space computation into an own type PackageSpaces

laeubi:
Currently a big part of the complexity in `ResolverImpl` is the
calculation of the package spaces and this part is actually rather
isolated as it is effectively only calling one method at one place.

Because of this and to make further enhancements easier, this now
extracts this part into a new type `PackageSpaces`, this removes about
1000 LOC from the resolver and makes it much more manageable.


trancexpress:
@laeubi I'm checking a performance degradation after updating to a newer Eclipse version and I stumbled across this change.

This change is only moving code, right? There is no new code in it.

Why I ask:

I'm checking a performance degradation when updating our Eclipse. Call counting YourKit snapshots show a big difference in map operations:

| method | diff | old | new |
|--------------------------------|----------|----------|----------|
| org.apache.felix.resolver.util.OpenHashMap$MapIterator.nextEntry() OpenHashMap.java |  7825896 |  1143202  | 8969098 |
| org.apache.felix.resolver.util.OpenHashMap.get(Object) OpenHashMap.java | 7598143 |1085341 | 8683484 |
|org.eclipse.osgi.container.ModuleCapability.getResource() ModuleCapability.java | 5342862 |829160 | 6172022 |

In the CPU sampling snapshots, this itself doesn't make a big difference (though sampling is not always reliable). But the call stacks come from the two changed classes here.

laeubi:
> This change is only moving code, right? There is no new code in it.

Yes this should be the case.

> I'm checking a performance degradation when updating our Eclipse. Call counting YourKit snapshots show a big difference in map operations

The main problem currently is that even slight differences in the dependencies can lead to an explosion of permutations, it can even make a difference if you _upgrade_ or start _clean_.

Some things that can be tried out:
- Check if starting with `-clean` make a difference.
- If it takes really long (> 1 minute) you can try to dsable the batch resolving with `-Dequinox.resolver.revision.batch.size=1`

In any case please consider starting your application with `-console` and use the command `exportFrameworkState <path to store the data> false` this will create a file with all the manifest data of the application. This can then be used to be transtformed into a testcase to be furhter analyzed like here:
- https://github.com/eclipse-equinox/equinox/pull/1125

Sadly we do not have much of such real world examples yet, but this would then ensure that a particular setup do not result in a performance degration.

PR #1123, opened 2025-08-06
Add back support for parent url factories

tjwatson:


HeikoKlare:
Thank you so much for working on this so quickly, @tjwatson! Would be really great if we get this improvement into the upcoming 4.37 release.

I've tested this PR with our products in Tomcat environments and it appears to work fine. I see the Gogo shell being properly initialized again and our issues with downstream business logic seem to be fixed as well.
To confirm this, I can roll out this change through our CI pipeline, so that I also have some results based on the Jenkins executions where I found the original issue. I am very confident that the results will be just as positive as my local execution, but better safe than sorry. The rollout process takes some time, so I am not sure whether I can report on that before the weekend. I will try my best.

Would you say this change is already in a state that you would merge or do plan to further work on is before merging?

tjwatson:
> Would you say this change is already in a state that you would merge or do plan to further work on is before merging?

Now that we have some confirmation that it works for the real scenario I have some cleanup I need to do and javadoc updates.  I'll get to these today.  But I plan to make sure this is merged before the release.  I have my own testing I need to do for Open Liberty as well to make sure I didn't break anything there also.

HeikoKlare:
I have now tested again with the latest changes (https://github.com/eclipse-equinox/equinox/commit/fb806da7741f50e71e81aeb4360078a59c8a8606) and with our CI for the uses cases that broke before. They all work fine again.

merks:
Great teamwork guys!

HeikoKlare:
Great to see this going into the upcoming release. Thank you again for taking action so quickly. Highly appreciated!

PR #1124, opened 2025-08-10
Resolve Substitution packages

laeubi:


laeubi:
I now finally sorted out the problems with the test.

For my testcase it is

**Before:** Resolve takes 66 sec and needed to process 2663 permutations
**After:** Resolve takes 1 sec and needed to process 15 permutations

PR #1125, opened 2025-08-10
Add a new testcase to demonstrate a long running resolve operation

laeubi:
This now finally adds a testcase that shows the behavior of the current resolver implementation to sometimes creating very large sets of permutations until a solution is found. This currently takes > 1 minute to resolve and produces > 40.000 permutation of which > 2600 needs to be actually processed before a solution is found.


laeubi:
[testLargeSet](https://ci.eclipse.org/releng/job/equinox/job/PR-1125/1/testReport/AutomatedTests%20org.eclipse.osgi.tests.container.AllTests%20org.eclipse.osgi.tests.container/TestModuleContainer/testLargeSet) is executed on Jenkins (takes 2m 11sec)

PR #1126, opened 2025-08-11
Use empty candidates instead of null when removing from the map

laeubi:
Currently in certain situations we remove the requirement form the map completely when no providers are left, but this results in the FaultyResourcesReport to not working properly and assume a permutation is valid / better than it is.

This now replaces the entry in the map with an empty instance instead of completely removing it.


laeubi:
It seems 3 tests currently relies on the removal for some reason here, need to investigate that and will reduce the scope of the PR until then.

PR #1138, opened 2025-08-31
Do not depend on PlatformAdmin for computing the container timestamp

laeubi:
Currently equinox.registry uses the old compatibility service PlatformAdmin to gather a timestamp that can be used to detect if something changed. In platform we already migrated to an approach that uses the last modified of the modules instead.

This now applies the same here to get one less reference to the PlatformAdmin service.

@tjwatson any concerns?


laeubi:
If performance ever gets relevant one might want to improve the computation, but I also doubt its noticeable.

PR #1143, opened 2025-09-07
Bump version for split package signer problem  o.e.equinox.registry

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3300


merks:
TCK failure known and unrelated.

PR #1145, opened 2025-09-16
Replace org.eclipse.core.runtime split-package imports with bundle req

sratz:
org.eclipse.core.runtime is split across
  equinox/org.eclipse.equinox.common
and
  eclipse.platform/org.eclipse.core.runtime

which can lead to dependency resolution errors.

Replace Import-Packge with Require-Bundle to make sure eclipse.platform/org.eclipse.core.runtime
is actually included.

See also https://github.com/eclipse-tycho/tycho/pull/5362 which will make this problem visible as it reduces the implicit target platform content.

### DIFF
--- a/bundles/org.eclipse.equinox.cm.test/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.cm.test/META-INF/MANIFEST.MF
@@ -3,9 +3,8 @@ Bundle-ManifestVersion: 2
 Bundle-Name: Test Plug-in
 Bundle-Vendor: Eclipse.org - Equinox
 Bundle-SymbolicName: org.eclipse.equinox.cm.test
-Bundle-Version: 1.2.0.qualifier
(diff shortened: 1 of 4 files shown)


tjwatson:
I don't understand this commit.  Our split packages that only have "part" of the package should have a `mandatory` directive that prevent `Import-Package` like we had here to wire to them because they are missing the required matching attribute.  So the only `org.eclipse.core.runtime` package that should have matched was the "full" one from the core.runtime bundle.

merks:
Yes, that's true:

image

But how OSGi does the wiring versus how this metadata maps to p2 metadata that p2 uses to resolve requirements don't always coincide.  So I wasn't sure if the latter was the problem and didn't say anything...

tjwatson:
> As it is only test-bundles it seem easier to just use require bundle than to try get it work otherwise.

As long as we are only doing this for test bundles that is fine with me.  But I don't want to see this happen for the "real" bundles.

Issue #1136, opened 2025-08-29
License details for org.eclipse.equinox.executable are inconsistent across the platforms


merks:
While working on SBOM generation I started to wonder why only a subset of executable IUs have license information, i.e., specifically only the linux IUs

<img width="954" height="512" alt="Image" src="https://github.com/user-attachments/assets/2b46bc0d-bdd2-4b7a-96ad-d2459cdadd94" />

I traced that back to the `about.html` in the `gtk_root` folder which is used in the files we see in the search:

<img width="939" height="996" alt="Image" src="https://github.com/user-attachments/assets/4a757659-c42e-4909-acfe-59ad0d41bfd2" />

I imagine we should have just a `root` folder and all the places that specify ``gtk_root`  should use that folder while the analogous things for win32 and macos should also use that `root` folder.

This stuff is all quite special (with horrible ant scripts) so I would be quite concerned that its easily broken.

So for now I just record this issue.  

@HannesWell  @akurtakov 

Nothing urgent here, but probably you guys have some more experience here.

merks:
I generated an SBOM for the latest I-build and all the binaries have an associated license.:

<img width="529" height="282" alt="Image" src="https://github.com/user-attachments/assets/8d9aa37e-a164-4823-89fa-3f38db961f99" />

PR #1149, opened 2025-09-30
Include about.html for all executables not just for linux executables

merks:
https://github.com/eclipse-equinox/equinox/issues/1136


merks:
The build issues unrelated.

PR #1151, opened 2025-10-07
Mark for deletion the constants for AIX, Solaris, HP/UX, QNX, OS/400,

vogella:
OS/390, z/OS, and Photon windowing system

### DIFF
--- a/bundles/org.eclipse.equinox.launcher/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.launcher/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.equinox.launcher;singleton:=true
-Bundle-Version: 1.7.0.qualifier
+Bundle-Version: 1.7.100.qualifier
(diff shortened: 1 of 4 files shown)


merks:
This serves to create new warnings in org.eclipse.equinox.launcher.Main.   When marking something for deletion, removing all uses of it would seem to be a prerequisite...

tjwatson:
> This serves to create new warnings in org.eclipse.equinox.launcher.Main. When marking something for deletion, removing all uses of it would seem to be a prerequisite...

I agree.  Otherwise I don't have an objection with the additional deprecation, although it seems more trouble than it is worth.

HannesWell:
> > This serves to create new warnings in org.eclipse.equinox.launcher.Main. When marking something for deletion, removing all uses of it would seem to be a prerequisite...
> 
> I agree. Otherwise I don't have an objection with the additional deprecation, although it seems more trouble than it is worth.

The Constants is package private, therefore I don't see how any of the now deprecated constants can be accessed from anywhere else than this package. Furthermore the `org.eclipse.equinox.launcher` jar is usually not contained in the OSGi runtime and just used as startup. So I believe there cannot be any reference to it and I think we can safely remove these now unused constants. The potentially troubling change was this removal of their consideration. Or do I miss something?

PR #1170, opened 2025-10-21
Add debug/loader/package/list trace option to the framework

tjwatson:
The org.eclipse.osgi/debug/loader/packages trace option is String
that specifies a comma separate list of package names to trace for
class loading.

Like other framework trace options; this new trace option can be
configured with the OSGi LoggerAdmin. In order to map into LoggerAdmin's
LoggerContext the logger name includes the package name for example

The Map<String,LogLevel> logLevels set with a LoggerContext may use the
following values

org.eclipse.osgi/debug/loader/packages=DEBUG
org.eclipse.osgi/debug/loader/packages/+/pkgname1=DEBUG
org.eclipse.osgi/debug/loader/packages/+/pkgname2=DEBUG

This ends up setting into Equinox trace options a String like
org.eclipse.osgi/debug/loader/packages=packagename1,packagename2

Note that the root logger org.eclipse.osgi/debug/loader/packages
must be set to debug or trace to get the messages becuase that is
the logger name the framework uses for all packages that are traced.

### DIFF
--- a/bundles/org.eclipse.osgi.tests/src/org/eclipse/equinox/log/test/LogEquinoxTraceTest.java
+++ b/bundles/org.eclipse.osgi.tests/src/org/eclipse/equinox/log/test/LogEquinoxTraceTest.java
@@ -15,14 +15,18 @@ package org.eclipse.equinox.log.test;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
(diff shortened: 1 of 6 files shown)


jonahgraham:
@tjwatson I can't see how this change could have caused my problem, but I am grasping at straws trying to understand the regression reported here https://github.com/eclipse-platform/.github/issues/257 and it seems possible that a change in equinox could cause the behaviour.

PR #1184, opened 2025-11-04
Update MacOS to newer version

akurtakov:
As per https://github.com/actions/runner-images/issues/13046

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
         config: 
           - { name: Linux, runner-os: ubuntu-latest, ws: gtk, os: linux, native-extension: so }
           - { name: Windows, runner-os: windows-2022, ws: win32, os: win32, native-extension: dll }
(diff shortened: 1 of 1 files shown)


HannesWell:
> @HannesWell You might have something to share here too.

With respect to the native builds this should be fine as the results of the GH workflows is not used further. 

Regarding Java-8, could we just not test on java-8 on Mac?

akurtakov:
> > @HannesWell You might have something to share here too.
> 
> With respect to the native builds this should be fine as the results of the GH workflows is not used further.
> 
> Regarding Java-8, could we just not test on java-8 on Mac?

Not that easy as bundles that have BREE 1.8 will fail to resolve as we use -Pbree-libs so for that to happen bigger changes will be needed and someone would have to fight with that ( I have zero intend to be the one).

HannesWell:
> All GHAs doing the same is fine as this never gets published anywhere.

Yes. But if possible I think it's also doesn't hurt to have the others to still be strict.
I just pushed an update that should demonstrate how it could be done to drop the `bree-libs` profile only for macOS.

PR #1185, opened 2025-11-06
Fix NPE during shutdown when ServiceRegistry is cleared before ServiceTracker cleanup

laeubi:
@tjwatson WDYT?


tjwatson:
Let me try to reproduce this because this is only one place where a null check would be needed and I don't understand how the registry is null at this point.

PR #1186, opened 2025-11-06
Fix javax.servlet.descriptor appears unused

laeubi:
Fix https://github.com/eclipse-equinox/equinox/pull/1083 as the problem was detected there.

This pull request addresses a dependency issue in the `ServletContextAdaptor` class by ensuring that `JspConfigDescriptor` is explicitly referenced, which resolves problems with automatic dependency detection by bnd. The change improves compatibility and correctness when proxies are created for servlet contexts.

Dependency management improvements:

* Added the `@Referenced` annotation to the `ServletContextAdaptor` class to explicitly reference `javax.servlet.descriptor.JspConfigDescriptor`, ensuring the class is available for proxy creation and addressing a limitation in bnd's dependency analysis.


chrisrueger:
Improvement in bnd underway https://github.com/bndtools/bnd/pull/6921

PR #1203, opened 2025-11-11
Add org.eclipse.equinox.p2.sdk.feature.group to API baseline

merks:


merks:
Has no build impact.  Merging immediately.

PR #1205, opened 2025-11-13
Enable Control Flow Guard in Windows executables

### LINKED ISSUE #851: Enable Control Flow Guard and Stack Protection in compiled Eclipse executables
Security teams are starting to flag the eclipse executable as being insecure due to Control Flow Guard and Stack Protection being disabled. Those features should be enabled at compile time so that OS with related features enabled can enforce the protection that they provide.

Here is a screenshot taken from viewing the properties of the eclipse.exe process in Sysinternals Process Explorer, showing Control Flow Guard and Stack Protection being disabled.
Image

https://learn.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard?view=msvc-170
https://learn.microsoft.com/en-us/windows/win32/secbp/control-flow-guard
https://developers.redhat.com/articles/2022/06/02/use-compiler-flags-stack-protection-gcc-and-clang

laeubi:
Add /guard:cf flag to compilation and linking steps in make_win64.mak to enable Control Flow Guard protection. Stack protection with -GS flag was already enabled.

Fixes #851

@njakubiak can you check if this fixes the issue for you?

### DIFF
--- a/bundles/org.eclipse.equinox.launcher.win32.win32.aarch64/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.equinox.launcher.win32.win32.aarch64/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.equinox.launcher.win32.win32.aarch64;singleton:=true
-Bundle-Version: 1.2.1400.qualifier
+Bundle-Version: 1.2.1500.qualifier
 Fragment-Host: org.eclipse.equinox.launcher;bundle-version="[1.7.0,1.8.0)"
 Eclipse-PlatformFilter: (& (osgi.ws=win32) (osgi.os=win32) (osgi.arch=aarch64))
 Bundle-Localization: launcher.win32.win32.aarch64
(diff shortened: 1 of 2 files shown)


merks:
FYI, I tried copying the eclipse_*dll and the eclipse.exe into an installation and it did not appear to cause any immediate problems.

HannesWell:
> Something is broken with the native builds, we need to wait for @HannesWell here if no one else has an idea what might gets wrong.

At the moment one of the Windows build agents is offline. I asked the IT team to re-connect it:
- https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6984

The previous build failed after the native build succeeded. I haven't investigated yet if this is related.

arunjose696:
> At the moment one of the Windows build agents is offline. I asked the IT team to re-connect it:
> 
> * https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6984

From the issue, it looks like the agent is now back online. Could someone please rerun the workflows to verify?
It would also be great if you could rerun the workflows for the following PR, which is affected by the same issue:

- https://github.com/eclipse-equinox/equinox/pull/1213

laeubi:
@arunjose696 I already rerun the workflow, but now the x86 workflow fails (that previously succeeded).

So it seems for some reasons the both windows agents influence each other and only one of them can succeed at a time.

laeubi:
The Jenkins build succeed, to make progress here I merge this anyways GH actions currently seem to be broken.

PR #1211, opened 2025-11-25
Temporarily restore EquinoxBundle#getModuleClassLoader(boolean)

HeikoKlare:
The private method `EquinoxBundle#getModuleClassLoader(boolean)` was recently (https://github.com/eclipse-equinox/equinox/pull/1157) replaced by a method with a different signature. Some relevant consumers (at least Jetty OSGi's [DefaultBundleClassLoaderHelper](https://github.com/jetty/jetty.project/blob/c0c432b079d6d506d1e598a4d1a6cc7a3a381af4/jetty-core/jetty-osgi/src/main/java/org/eclipse/jetty/osgi/util/DefaultBundleClassLoaderHelper.java#L233-L241)) reflectively access the replaced method. To avoid that compatibility of such consumers breaks, this change temporarily reintroduces the replaced method. It prints an according warning about the upcoming removal to stderr. It is supposed to be removed once the known consumers are adapted.

This proposal is based on the discussion on the PR that changed the method start with this comment: https://github.com/eclipse-equinox/equinox/pull/1157#issuecomment-3574481401

I have already tested the change with one of our Jetty products. It (1) solves the issue of failing to properly start the Jetty product and (2) prints the error message as expected because of the readded method being called.

@laeubi I am happy to adapt this in any desired way. I chose to use `System.err` instead of `System.out` to make it a bit better visible within all the output you get especially when running a Jetty product. I will also raise an issue in the Jetty project today.

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/EquinoxBundle.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/EquinoxBundle.java
@@ -695,6 +695,17 @@ public class EquinoxBundle implements Bundle, BundleReference {
 		return new ResolvedModuleClassLoader(cl, report);
 	}
 
+	// This method is only preserved as a temporary workaround for consumers that
+	// reflectively access it until they are adapted, such as Jetty OSGi's
+	// DefaultBundleClassLoaderHelper
+	// (see https://github.com/jetty/jetty.project/pull/14086)
+	@Deprecated
+	protected ModuleClassLoader getModuleClassLoader(boolean unused) {
+		System.err.println(
+				"WARNING: Reflective call to EquinoxBundle#getModuleClassLoader(boolean) detected. This method will be removed soon."); //$NON-NLS-1$
+		return getModuleClassLoader().moduleClassLoader;
+	}
+
 	@Override
 	public Enumeration<URL> getResources(String name) throws IOException {
 		try {


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/EquinoxBundle.java
```diff
@@ -695,6 +695,16 @@ public ModuleClassLoader run() {
 		return new ResolvedModuleClassLoader(cl, report);
 	}
 
+	// This method is only preserved as a temporary workaround for consumers that
+	// reflectively access it until they are adapted, such as Jetty OSGi's
+	// DefaultBundleClassLoaderHelper
```
laeubi: Can we reference https://github.com/jetty/jetty.project/pull/14086 here maybe (or any other bug that tracks the problem at jetty)

HeikoKlare: Of course. I added a reference to the Jetty PR with https://github.com/eclipse-equinox/equinox/pull/1211/commits/6a7fb39a0ce003c97e09734c595d9b5ae9c7f8d6

Awesome that you already made a proposal for a fix to Jetty! ❤️

laeubi (approved):
+1 as PL, looks good to me and we give some more time for jetty project to adopt.

HeikoKlare:
I have just verified your [Jetty fix](https://github.com/jetty/jetty.project/pull/14086) with our product (using the current development state of jetty.osgi) and it works decently fine so far.

I propose to proceed as follows: I will monitor when a new Jetty version with that fix will become available at SimRel: https://download.eclipse.org/tools/orbit/simrel/maven-jetty/release/
Once it is available, I will propose a PR to Equinox that reverts the temporary restoration of `EquinoxBundle#getModuleClassLoader(boolean)`.

merks:
Jetty 12.1.5 is out:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3546

I'll build new Obit repos soon:

https://github.com/eclipse-orbit/orbit-simrel/commit/2cd4957c78ec3f02f503252d3e0d918ddf9e60c2

I am doing a nightly build for Jetty now, but will hold off producing the release repository until after the weekend.

tjwatson:
I suggest we wait a few releases before pulling out the internal method since we know at least one user of it.  It does us no real harm to keep it a few releases.

PR #1216, opened 2025-12-05
[Build] Adapt to move of macOS entitlement in platform.releng.aggregator

HannesWell:
Consequence of
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3480

This shouldn't be submitted before the linked PR is.


merks:
@jonahgraham 

FYI I compare the one being used here and the one being used by EPP and there is no difference:

| diff platform-sdk-4.38/git/eclipse.platform.releng.aggregator/products/eclipse.entitlement epp/git/packages/releng/org.eclipse.epp.config/macos/product.entitlements

PR #1217, opened 2025-12-09
Query String is not correctly parsed for bundle resource URLs

laeubi:


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/url/BundleResourceHandler.java
```diff
@@ -34,7 +34,7 @@
 /**
  * URLStreamHandler the bundleentry and bundleresource protocols.
  */
-
+//TODO needs fixing for parsing query correctly!
```
tjwatson: Did you intend to leave this TODO here?

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/url/BundleResourceHandler.java
```diff
@@ -94,16 +94,32 @@ protected void parseURL(URL url, String str, int start, int end) {
...
-			path = spec.substring(pathIdx, end);
-		else if (end > pathIdx) {
+		// Extract query and fragment from the spec
+		// Query comes after '?' and before '#' (fragment)
+		String query = null;
+		int queryIdx = spec.indexOf('?', pathIdx);
+		int fragmentIdx = spec.indexOf('#', pathIdx);
+		int pathEnd = end;
+		
+		// Determine where the path ends (before query or fragment)
```
tjwatson: Confirming the intention.  If the fragment `#` comes before the query `?` then you treat the `?` as part of the ref?

tjwatson (commented):
I think we need a test that confirms the URLs with queries can be opened and read correctly.

laeubi:
@tjwatson all test pass now!

PR #1223, opened 2025-12-11
Migrate appadmin ConfigurationSessionTests to JUnit 5

HeikoKlare:
The test classes `ApplicationAdminTest` and `ApplicationRelaunchTest` rely on the JUnit-3-specific ConfigurationSessionTest.
The SessionTestExtension has been introduced as a JUnit-5-based replacement. This change adapts the test classes to be executed with JUnit 5, using the SessionTestExtension. It also adapts the according test suite to be executed with JUnit 5.

Follow up to:
- #654


akurtakov:
Thanks for further modernizing the tests!

PR #1242, opened 2026-01-26
Bump minor version of Windows launcher fragments

HeikoKlare:
The recent changes of the Windows native launcher (DPI awareness, default autoscaling mode) significantly change the experience of the launcher behavior:
- https://github.com/eclipse-equinox/equinox/pull/1239
 
This change bumps the minor version of the fragments to reflect the relevance of the changes.

This is just a proposal as there is no clear definition when to bump the minor version upon changes of the native implementation parts. I consider a minor bump reasonable for the mentioned changes, but it's up to the committers to decide if the bump should be applied.


HeikoKlare:
> Yes sounds good. But shouldn't the Windows ARM fragment be bumped too?

Indeed. I updated the PR to bump the version of both Windows fragments.

PR #1265, opened 2026-04-04
[Build] Ensure availability guards in macOS native launcher and clean up

HannesWell:
Also build native launchers for macOS on agents of the targeted architecture.

This also aligns the Equinox pipeline with SWT:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3182


HannesWell:
> This should not be submitted before the old mac agents are uninstalled and rebuilds of the native launcher binaries will definitively use the new mac agents:

The new agents are active and the old agents are removed. Submitting this now.

PR #1271, opened 2026-04-09
Add lockAndOpen method to always keep lock

tjwatson:
Update usage to always unlock in finally regardless of the success of the bundle file open.


tjwatson:
This is released to #896

I don't believe this fixes the issue observed there.  This change updates the code pattern to a more "standard" lock-try-finally-release model by always leaving the lock acquired on return of `lockAndOpen` so that we can always release the lock in a finally block regardless of successfully opening the bundle file.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/storage/bundlefile/CloseableBundleFile.java
```diff
@@ -77,18 +78,39 @@ public CloseableBundleFile(File basefile, BundleInfo.Generation generation, MRUB
...
+				debug.trace(OPTION_DEBUG_BUNDLE_FILE_OPEN_LOCK, "lockOpen obtained lock - " + toString()); //$NON-NLS-1$
+			}
 			return true;
 		} catch (Throwable e) {
-			// always unlock on any throwable
-			openLock.unlock();
+			if (debug.DEBUG_BUNDLE_FILE_OPEN_LOCK) {
+				debug.trace(OPTION_DEBUG_BUNDLE_FILE_OPEN_LOCK, "lockOpen failed lock - " + toString()); //$NON-NLS-1$
+			}
+			if (unlockOnError) {
```
tjwatson, line 110: This is flawed because it is possible to throw an exception out of the method without ever releasing the lock.  That will cause the thread to retain the lock when it shouldn't

PR #1273, opened 2026-04-14
[Build] Build macOS native launcher binaries on macOS 15 agents

HeikoKlare:
With recent changes, the macOS native launcher binaries are built on the Jenkins agents with the fitting architecture. This includes the usage of different macOS SDKs. For the x86 build, a macOS 15 SDK is used, whereas the aarch64 build uses a macOS 26 SDK. After that change, the behavior of applications using the binaries built with macOS 26 SDK show significantly changed behavior and appearance (such as title bar text alignment, background colors, invisible lists and the like), rendering applications hardly usable or even partly unusable. See, e.g.:
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3884
- https://github.com/eclipse-platform/eclipse.platform.ui/issues/3885
- https://github.com/eclipse-platform/eclipse.platform.swt/issues/3230

With this change, the binaries are always built on the x86 agent, which runs macOS 15 and uses the according SDK, to restore existing behavior. This is supposed to be a temporary solution until the Eclipse Platform has been updated to be compatible with the macOS 26 SDK behavior. See https://github.com/eclipse-platform/eclipse.platform.ui/issues/3884#issuecomment-4241839229.

The aarch64 binary is now properly built with/against macOS 15 SDK now: https://ci.eclipse.org/releng/job/equinox/view/change-requests/job/PR-1273/5/execution/node/3/ws/equinox.binaries/org.eclipse.equinox.launcher.cocoa.macosx.aarch64/
```
vtool -show eclipse_11921.so     
eclipse_11921.so:
Load command 8
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 11.0
      sdk 15.5
   ntools 1
     tool LD
  version 1167.5
Load command 9
      cmd LC_SOURCE_VERSION
  cmdsize 16
  version 0.0
```

<s>
> [!WARNING]
>For now, this is for testing purposes. I am not even sure if it is possible to compile aarch64 binaries on an x86 system. We just did it vice versa before.
</s>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,6 +15,10 @@
 
 def runOnNativeBuildAgent(String platform, Closure body) {
 	def final nativeBuildStageName = 'Perform native launcher build'
+	// Use x86 agent for aarch64 build to ensure that the baseline macOS SDK (the oldest) is used consistently for all binary builds
+	if (platform == 'cocoa.macosx.aarch64') {
+		platform = 'cocoa.macosx.x86_64'
+	}
 	def agentLabel = 'native.builder-' + platform
 	if (platform == 'gtk.linux.x86_64') {
 		podTemplate(inheritFrom: 'basic' /* inherit general configuration */, containers: [


Phillipus:
You could download Command Line Tools for 16.4 and install that on the aarch64 machines (this includes SDK 15). https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_16.4/Command_Line_Tools_for_Xcode_16.4.dmg

HeikoKlare:
> You could download Command Line Tools for 16.4 and install that on the aarch64 machines (this includes SDK 15). https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_16.4/Command_Line_Tools_for_Xcode_16.4.dmg

Yes, that's basically what I did locally. But I am not sure if we want to add that hassle to EF infrastructure. In case it does not work to compile for aarch64 on x86 machine, that's probably the way to go.

sratz:
> > You could download Command Line Tools for 16.4 and install that on the aarch64 machines (this includes SDK 15). https://download.developer.apple.com/Developer_Tools/Command_Line_Tools_for_Xcode_16.4/Command_Line_Tools_for_Xcode_16.4.dmg
> 
> Yes, that's basically what I did locally. But I am not sure if we want to add that hassle to EF infrastructure. In case it does not work to compile for aarch64 on x86 machine, that's probably the way to go.

We are now in a situation where the build result highly depends on the infrastructure around it, and it is now very crucial that the x86 machine stays around and on 15.

If we can enforce the compiler by making sure the proper toolchain is used independently from the environment, that would make things more stable / reproducible / explicit.

So I think that is something we should still do.

I suppose there is no easy way to do this on-the-fly within the CI as the DMG download requires authentication?

But for now this change looks good.

HeikoKlare:
ftr: the cross-compilation seems to properly create an aarch64 binary:
```
objdump -f eclipse_11921.so

eclipse_11921.so:	file format mach-o arm64
```

HannesWell (approved):
> If we can enforce the compiler by making sure the proper toolchain is used independently from the environment, that would make things more stable / reproducible / explicit.
> 
In general I agree that we should define the properties of the binaries as much as possible in this repo's code (likewise for SWT).
But with what Apple provides that seems difficult to achieve.

> I suppose there is no easy way to do this on-the-fly within the CI as the DMG download requires authentication?

I think so. Maybe we can ask the IT for a corresponding token and authorize a download with that? If automated/tool based downloads are even possible and if the download isn't too large.

But for example for I-build tests we also download the corresponding JDKs on each run in multiple cases (only Linux has all pre-installed).


That being said, this change is fine for me (after slightly generalizing the comment).

basilevs:
Given different SDKs for SWT master, Equinox Launcher, which Java distribution is Eclipse RCP user supposed to use?

HeikoKlare:
> Given different SDKs for SWT master, Equinox Launcher, which Java distribution is Eclipse RCP user supposed to use?

In general, Eclipse is built and tested with Temurin. So it's usually best/safest to use a Temurin JDK/JRE. The versions that we target to support are documented in the project plan, e.g., for 4.40: https://eclipse.dev/eclipse/development/plans.html?file=plans/eclipse_project_plan_4_40.xml

I don't remember any issues of similar severity than the current macOS target SDK issues that would have required being more precise on which JDK build is supported. In particular, different kinds of OpenJDK builds just worked fine.

PR #1274, opened 2026-04-16
Add apache.commons.commons-fileupload to equinox.server.jetty feature

HannesWell:
The `org.eclipse.equinox.http.servlet` bundle requires Apache's `commons-fileupload` only optionally
https://github.com/eclipse-equinox/equinox/blob/36de39bde95dcbf3cd4364fe09dfc931264a1389/bundles/org.eclipse.equinox.http.servlet/META-INF/MANIFEST.MF#L23-L25

And the only hard requirement to it are currently just tests. In case they are removed from the Eclipse repository as proposed in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3772

... `commons-fileupload` would vanish from the Eclipse repository, without further changes.

As the `commons-fileupload` looks small and innocent I propose to at least require it in the `equinox.server.jetty` feature.
Alternatively `org.eclipse.equinox.http.servlet` could also make the dependency on `commons-fileupload` mandatory. I have to admit that I don't understand why it's optional at all.
In case we just make it mandatory, https://github.com/eclipse-equinox/equinox/pull/1275 could become even simpler and the clean-up should instead be included here.

### DIFF
--- a/features/org.eclipse.equinox.server.jetty/feature.xml
+++ b/features/org.eclipse.equinox.server.jetty/feature.xml
@@ -15,6 +15,10 @@
       %license
    </license>
 
+   <requires>
+      <import plugin="org.apache.commons.commons-fileupload"/>
+   </requires>
+
    <plugin
          id="org.eclipse.equinox.http.jetty"
          version="0.0.0"/>
--- a/features/org.eclipse.equinox.server.jetty/feature.xml
+++ b/features/org.eclipse.equinox.server.jetty/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.equinox.server.jetty"
       label="%featureName"
-      version="1.11.1000.qualifier"
+      version="1.11.1100.qualifier"
       provider-name="%providerName"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">


merks:
So you think specifying `x-installation:=greedy` would help?

HannesWell:
> So you think specifying `x-installation:=greedy` would help?

As far as I understand it, that should also pull it into the p2-repo, but also in installations, like with adding it to the feature.
The idea with that adding it to the feature was, that one is still able to install the bundle without fileupload, if desired (I still don't know why one should want that).
If fileupload is in the eclipse p2-repo, it will probably also be in a target if  `x-installation:=greedy` is used.
So one could have a really hard time to exclude it with that, if desired.

Since I don't understand why it's optional at all, I'd like seek advice from @tjwatson, @laeubi or @rotty3000 on how to proceed here.

Of course we could also just add fileupload only to the eclipse repo directly, but I fear that would lead to some disconnect to the need.

HannesWell:
@tjwatson and @laeubi, what's your opinion on this topic?

> Of course we could also just add fileupload only to the eclipse repo directly, but I fear that would lead to some disconnect to the need.

It turned out that `org.apache.commons.commons-fileupload` is in fact added explicitly to the p2-repository from which tests are build:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/3e7c7448db6928e3c1b58ded9ed8aa85e8f64c76/sites/eclipse-junit-tests/category.xml#L81

Therefore this is actually not required, but still it would allow to slightly simplify the set up of I-build tests.

Furthermore the general question from https://github.com/eclipse-equinox/equinox/pull/1275#discussion_r3102516593 still stands for me.

HannesWell:
Since there is no objection I'm continuing with this now.
Besides `org.apache.commons.commons-fileupload` this effectively only adds Apache `commons-io` and I assume both are not problematic.
Furthermore, now that mixed products are working well for some time, one can simply include the two Plugins contained in this feature directly in a product, in case `commons-fileupload` is actually harmful for a specific use-case.

PR #1275, opened 2026-04-16
Simplify optional registration of of MultipartSupportFactoryImpl

HannesWell:
It's sufficient to catch a potential `NoClassDefFoundError` when creating the `MultipartSupportFactoryImpl` instance. The indirection using a ServiceLoader is not necessary for that.

I did a test with a simple Java application and still `ServletRegistration` as initialized cleanly as expected when the `commons-fileupload` jar was absent.

### DIFF
--- a/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/registration/ServletRegistration.java
+++ b/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/registration/ServletRegistration.java
@@ -17,14 +17,13 @@
 package org.eclipse.equinox.http.servlet.internal.registration;
 
 import java.io.IOException;
-import java.util.*;
+import java.util.List;
 import javax.servlet.*;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.Part;
 import org.eclipse.equinox.http.servlet.internal.context.ContextController;
 import org.eclipse.equinox.http.servlet.internal.context.ServiceHolder;
-import org.eclipse.equinox.http.servlet.internal.multipart.MultipartSupport;
-import org.eclipse.equinox.http.servlet.internal.multipart.MultipartSupportFactory;
+import org.eclipse.equinox.http.servlet.internal.multipart.*;
 import org.osgi.framework.ServiceReference;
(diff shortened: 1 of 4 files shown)


bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/registration/ServletRegistration.java
```diff
@@ -35,19 +34,10 @@ public class ServletRegistration extends EndpointRegistration<ServletDTO> {
 	private static MultipartSupportFactory factory;
 
 	static {
-		ServiceLoader<MultipartSupportFactory> loader = ServiceLoader.load(MultipartSupportFactory.class);
-
```
tjwatson, line 39: I don't recall why ServiceLoader was used here.  Are we sure others are not using something like SPI-fly to substitute an implementation when ours cannot be loaded?

tjwatson: I guess that is not really possible because the interface is in a package that isn't exported.  But perhaps fragments could be used.  I'm pretty sure I remember @rotty3000 did it this way for a reason (although, I'm not sure he will see my `@` to him).

HannesWell: > I guess that is not really possible because the interface is in a package that isn't exported.

Yes, plus to Service Loader Mediator headers required for an OSGi runtime are also not provided for this service.
Though, through Plug-ins it should still be possible to supply other implementations.

I assumed that the Service-Loader was just used as a very save way to avoid failures in class-loading of the caller in case the optional dependency is missing.
Nevertheless it would be good if @rotty3000 could clarify.

But the more interesting question to me is actually why the dependency is optional at all, respectively if it would be fine to add it at least to the jetty feature:
- https://github.com/eclipse-equinox/equinox/pull/1274

HannesWell:
Since it's unclear how this part of the code is used and the gain is only moderate I'll close this for now as it's not worth the potential further effort for me.
If somebody is interested, I'm glad to revive this at any time.

PR #1276, opened 2026-04-17
Suppress stack trace on resolution-error BundleException

vogella:
## Summary

- The stack trace on a `RESOLVE_ERROR` `BundleException` is always the same OSGi event-dispatch / start-level chain and carries no diagnostic value — the message already identifies the bundle and the unresolved requirement.
- Suppress the trace by default so bundle-resolution failure logs show only the actionable information.
- Set the system property `equinox.resolver.errorStacktrace=true` to restore the full stack trace for debugging.
- Other `BundleException` types (`ACTIVATOR_ERROR`, `NATIVECODE_ERROR`, `STATECHANGE_ERROR`, etc.) are unaffected — their stack traces point at the actual culprit and remain useful.

### Before

```
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.pde.spy.css [326]
  Unresolved requirement: Require-Bundle: org.eclipse.jface; bundle-version="3.39.0"

    at org.eclipse.osgi.container.Module.start(Module.java:495)
    at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:2110)
    at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$1$1.execute(EquinoxContainerAdaptor.java:146)
    at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:2101)
    ... (several more frames of OSGi internals)
```

### After

```
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.pde.spy.css [326]
  Unresolved requirement: Require-Bundle: org.eclipse.jface; bundle-version="3.39.0"
```

### Affected throw sites

- `Module.java:495` — bundle start fails due to unresolved requirements
- `SystemModule.java:91` — system bundle init fails due to unresolved requirements
- `EquinoxBundle.java:653` — class load fails because the bundle cannot resolve

### Implementation

Added a shared helper `InternalUtils.newResolveError(String)` that constructs a `BundleException` whose `fillInStackTrace()` is a no-op by default. When `equinox.resolver.errorStacktrace=true` is set, a regular `BundleException` is constructed so the trace is available for debugging.

## Test plan

- [ ] Trigger a resolution failure (e.g. install a bundle requiring a missing package) and verify the log contains only the message, no stack trace
- [ ] Set `-Dequinox.resolver.errorStacktrace=true` and verify the full stack trace is restored
- [ ] Trigger a non-resolution `BundleException` (e.g. a failing BundleActivator) and verify its stack trace is still present
- [ ] Existing `org.eclipse.osgi.tests` pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/Module.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/Module.java
@@ -25,6 +25,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.eclipse.osgi.container.ModuleContainerAdaptor.ModuleEvent;
 import org.eclipse.osgi.framework.util.ThreadInfoReport;
 import org.eclipse.osgi.internal.container.EquinoxReentrantLock;
+import org.eclipse.osgi.internal.container.InternalUtils;
 import org.eclipse.osgi.internal.messages.Msg;
 import org.eclipse.osgi.report.resolution.ResolutionReport;
 import org.eclipse.osgi.util.NLS;
@@ -492,7 +493,7 @@ public abstract class Module implements BundleReference, BundleStartLevel, Compa
 					return;
 				if (getState().equals(State.INSTALLED)) {
 					String reportMessage = report.getResolutionReportMessage(getCurrentRevision());
-					throw new BundleException(Msg.Module_ResolveError + reportMessage, BundleException.RESOLVE_ERROR);
+					throw InternalUtils.newResolveError(Msg.Module_ResolveError + reportMessage);
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


tjwatson:
Are the before and after descriptions accurate?  From the summary they appear reversed.

vogella:
> Are the before and after descriptions accurate? From the summary they appear reversed.

I think they are. Before this patch you get always the org.eclipse.osgi.container.Module.start(Module.java:495) stack trace and after this patch you only see the dependency errors.

IIRC we discussed that trace ~10 years ago and at that point I think you said the trace can be useful that is why I add the property to show it. I personally don't think I will ever need this property.

vogella:
Before this patch:

image

tjwatson:
I have to admit that I don't understand the motivation for removing the stack trace from methods like `Bundle.start()` when there are resolution issues vs. any other kinds of error.  I believe the premise that these methods are only called by the start-level thread is wrong.  I know of lots of other examples where some code is installing and starting bundles themselves.

laeubi:
I agree that it does not sounds useful. These kind of things are not user errors in any way. If you really likes to improve things it would be much more valuable to make the logging configurable.

PR #1281, opened 2026-04-28
Cache build-in factories from ServiceLoader at Plurl construction 

### LINKED ISSUE #1280: Plurl creates ServiceLoader in Constructor which is not thread-safe
In PlurlImpl the method ServiceLoader.load() is called which returns a ServiceLoader. A ServiceLoader is defined by documation as not thread-safe.
This PlurlImpl is stored in URLConnection.setContentHandlerFactory().
When working with URLConnections from different threads this can lead to failures.

A detailed bug report is stored at: https://github.com/eclipse-osgi-technology/plurl/issues/30

I am not sure which version should be the master of the implementation. So I created a bug in both projects.

tjwatson:


merks:
Wow, that’s an impressively fast response. 🚀

PR #1292, opened 2026-05-19
Remove unnecessary single quotes for links

deepika-u:
This change removes the unnecessary quotes around `Secure Storage` in the properties strings.

The single quotes around the placeholder text in the preference page were likely added following older UI conventions. Quotes often differentiate literal text or placeholders from the rest of the message. Since the text already appears as a blue clickable link, the additional quotes are unnecessary. Removing them keeps the link still clickable while making the message cleaner and consistent with other preference links across the UI, where such quotes are not used.
The PR changes the link like other links provided in other areas in Preference page like `File Associations`.
As an example -> 
image

Before pr
image

After pr
image

### DIFF
--- a/bundles/org.eclipse.equinox.security.ui/src/org/eclipse/equinox/internal/security/ui/SecurityUIMsg.properties
+++ b/bundles/org.eclipse.equinox.security.ui/src/org/eclipse/equinox/internal/security/ui/SecurityUIMsg.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2007, 2008 IBM Corporation and others.
+# Copyright (c) 2007, 2026 IBM Corporation and others.
 #
 # This program and the accompanying materials
 # are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


deepika-u:
@merks : Can you take a look when you get some time please?

bundles/org.eclipse.equinox.security.ui/src/org/eclipse/equinox/internal/security/ui/SecurityUIMsg.properties
```diff
@@ -47,7 +47,7 @@ WIZARD_ERROR_NO_WRITE_ENGINE = There is no a writable trust engine!
 STR_CERT_VIEWER_FIELD=Field
 STR_CERT_VIEWER_VALUE=Value
 
-CATPAGE_LABEL_STORAGE = See <a>''{0}''</a> for settings related to the encrypted storage system.
+CATPAGE_LABEL_STORAGE = See <a>{0}</a> for settings related to the encrypted storage system.
```
merks, line 50: Are the following 3 going to come in as separate PRs?  Why not do all 4 here?

deepika-u: I dint actually touch the others because i dont get to see them in the UI like this. 
image

If you suggest we better do it, then i can do it right away.

merks: Yes, given it's the same pattern, wherever it appears will have the same "issue" so better just to fix them all.  😀

deepika-u: sure, will do it right away.

deepika-u:
Done, Thanks for your time.

merks:
In general @HannesWell  prefers that version increments are done as a separate commit (which is automatically generated) so that it's possible to revert a change without reverting the version increments (which would be bad and wrong).  In this case, I don't imagine we'd ever need a revert.

deepika-u:
Do you have any outstanding actions for me to handle in this pr? or are we good.

merks (approved):
Looks good.

PR #1297, opened 2026-05-28
disable external DTD loading in DOM parser factory

netliomax25-code:
The SAX branch in `XMLParsingServiceFactory.createService()` disables three
features but the DOM branch only disables two: `load-external-dtd` is missing.
Even with `external-general-entities`/`external-parameter-entities` off, a
`<!DOCTYPE root SYSTEM "http://attacker/x.dtd">` on input still causes the
DocumentBuilder returned from the system-published `DocumentBuilderFactory`
service to fetch that URL (SSRF / DTD-fetch via Xerces). Verified locally:
without the flag the parse throws `ConnectException` to the attacker host;
with it added the parse completes without any network attempt. Patch just
makes the DOM branch symmetric with the SAX branch right above it.

### DIFF
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
@@ -78,9 +78,10 @@ class XMLParsingServiceFactory implements ServiceFactory<Object> {
 		}
 		DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
 		try {
-			// completely disable external entities declarations:
+			// completely disable external entity declarations and external DTD loading:
 			factory.setFeature("http://xml.org/sax/features/external-general-entities", false); //$NON-NLS-1$
 			factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false); //$NON-NLS-1$
+			factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false); //$NON-NLS-1$
 		} catch (ParserConfigurationException e) {
 			throw new IllegalStateException(e.getMessage(), e);
 		}
--- a/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
+++ b/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.osgi.internal.framework;
 
+import javax.xml.XMLConstants;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.parsers.SAXParserFactory;
@@ -71,6 +72,8 @@ class XMLParsingServiceFactory implements ServiceFactory<Object> {
 				factory.setFeature("http://xml.org/sax/features/external-general-entities", false); //$NON-NLS-1$
 				factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false); //$NON-NLS-1$
 				factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false); //$NON-NLS-1$
+				// enable secure processing:
+				factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
 			} catch (Exception e) {
 				throw new IllegalStateException(e);
 			}
@@ -82,6 +85,8 @@ class XMLParsingServiceFactory implements ServiceFactory<Object> {
 			factory.setFeature("http://xml.org/sax/features/external-general-entities", false); //$NON-NLS-1$
 			factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false); //$NON-NLS-1$
 			factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false); //$NON-NLS-1$
+			// enable secure processing:
+			factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
 		} catch (ParserConfigurationException e) {
 			throw new IllegalStateException(e.getMessage(), e);
 		}


bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
```diff
@@ -81,6 +81,7 @@ private Object createService() {
 			// completely disable external entities declarations:
```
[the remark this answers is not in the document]
netliomax25-code: Good point. Updated the comment to mention external DTD loading so it covers all three features now.

netliomax25-code:
Rebased onto master to pick up the 4.41 version bump for org.eclipse.osgi. The branch was still at 3.24.200, which is what tripped the baseline comparator in the Maven check.

manandbytes:
@netliomax25-code 

> Before your contribution can be accepted by the project team contributors must electronically sign the Eclipse Contributor Agreement (ECA).

Check https://github.com/eclipse-equinox/equinox?tab=contributing-ov-file#eclipse-contributor-agreement for details.

> Last validated: 6 Jul 2026
> Errors: 2
>    Author Eclipse account cannot be found.
>        Email: netliomax25@gm*il DOT com
>        GitHub ID: netliomax25-code
>    Committer Eclipse account cannot be found.
>        Email: netliomax25@gm*il DOT com
>        GitHub ID: netliomax25-code

https://api.eclipse.org/git/eca/status/gh/eclipse-equinox/equinox/1297

netliomax25-code:
Yep, that's on me. Setting up the Eclipse account for netliomax25@gmail.com with the same GitHub ID and signing the ECA now; I'll retrigger the validation once it goes through.

tjwatson:
If we are trying to be secure by default should the following also be set for both SAX and DOM parsers here:

```java
    factory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
```

netliomax25-code:
Makes sense, done. Pushed a commit that sets FEATURE_SECURE_PROCESSING to true for both the SAX and DOM factories, after the existing feature flags.

Checked locally:
1. setFeature accepts it on both factory types without throwing. Support for this feature is mandatory per JAXP, so it shouldn't break alternative implementations resolved through the TCCL either.
2. A document with a SYSTEM DOCTYPE still parses cleanly with no network attempt, same as before.

The one behavior change beyond the flag itself is that the implementation's entity expansion and similar limits now apply to parsers handed out by this service, which seems consistent with the secure-by-default intent here.

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
```diff
@@ -71,16 +72,21 @@ private Object createService() {
 				factory.setFeature("http://xml.org/sax/features/external-general-entities", false); //$NON-NLS-1$
 				factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false); //$NON-NLS-1$
 				factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false); //$NON-NLS-1$
+				// enable secure processing:
```
tjwatson, line 75: Because we are starting to use XMLConstants here; could you update the above features to also use the constants instead of our own `http` strings?

I figured I would have you do this while we wait for the ECA to get straightened out for you.  Not sure what the hold up is there.

tjwatson (changes requested):

bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/internal/framework/XMLParsingServiceFactory.java
netliomax25-code: Took a look at this. Those three are SAX/Apache feature URIs, and javax.xml.XMLConstants doesn't define constants for them; the only feature it exposes is FEATURE_SECURE_PROCESSING, which is why that's the one that could move over to a constant. So there isn't anything in XMLConstants to swap the other three for.

Two options if you'd rather not keep the raw strings inline:
1. Pull them into private static final String constants in this class, named and shared between the SAX and DOM branches. Behavior stays identical, it just removes the literals.
2. Switch the DTD case to the standard XMLConstants.ACCESS_EXTERNAL_DTD property (set to ""), but that's a different mechanism (a property, not a feature) and only covers external DTDs, not the general/parameter entity features, so it wouldn't replace all three.

I'd lean toward option 1 to keep it minimal. Happy to push that if it works for you.

tjwatson: You are correct.  I missed that these values are not in the `XMLConstants` class.  For readability of the code I am fine leaving it as-is instead of a redirect to a local static.

Do you know why the ECA for your account isn't in sync?

netliomax25-code: Sounds good, leaving the feature strings as-is then.

On the ECA, that was lag on my end:
1. The Eclipse account for netliomax25@gmail.com only got created and the ECA signed on the 9th, after manandbytes flagged it, so the earlier validations had no account to match against.
2. The validation re-ran after that went through, and the eclipsefdn/eca check on the PR is green now.

One other thing I looked at: the red Configuration Admin TCK check is testCoordinatedConfigurationOnBeforeRegisteredManagedService, a coordination timing case. The full test results run on the same commit has 0 failures and nothing in that test touches XML parsing, so it doesn't look related to this change.

merks (approved):
This looks good to me. 

I don't know for sure, but I thought there were some tests that always fail....

netliomax25-code:
Right, the one red check is the Configuration Admin TCK (testCoordinatedConfigurationOnBeforeRegisteredManagedService), a coordination timing case that doesn't touch XML parsing. The full Test Results run on the same commit is green with 0 failures, and everything else including the ECA check passes now.

netliomax25-code:
Yes, that matches what I see here. The only red check is the Configuration Admin TCK, and the single failure is testCoordinatedConfigurationOnBeforeRegisteredManagedService, a coordination timing case that doesn't go near the XML parsing service. Everything else is green now, including the ECA check and the aggregated Test Results on this commit (0 failures).

merks:
FYI, I would merge this, but I prefer that @tjwatson does that.  It is summer vacation season, so that can lead to delays...  That being said, M2 starts this week (Thursday's build) so it would seem good to get this in for M2...

tjwatson (approved):
LGTM

PR #1299, opened 2026-05-29
reject inverted byte range in ResourceServlet Range.isValid

netliomax25-code:
Fuzzing the Range header against the resource servlet turned up `bytes=9999-1000`: Range.isValid() only checks each position against the resource length, never that first-byte-pos <= last-byte-pos. An inverted range is accepted, so contentLength() goes negative and the reply is a 206 with a malformed `Content-Range: bytes 9999-1000/<len>` instead of a 416. RFC 7233 treats such a range as unsatisfiable, so reject it.

### DIFF
--- a/bundles/org.eclipse.equinox.http.servlet.tests/src/org/eclipse/equinox/http/servlet/tests/ServletTest.java
+++ b/bundles/org.eclipse.equinox.http.servlet.tests/src/org/eclipse/equinox/http/servlet/tests/ServletTest.java
@@ -2005,6 +2005,34 @@ public class ServletTest extends BaseTest {
 	}
 
 	@Test
+	public void test_ResourceRangeRequest_InvalidRange() throws Exception {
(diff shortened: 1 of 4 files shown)


netliomax25-code:
Pushed a fix for the red CI:

1. Code Analysis was failing on the Tycho baseline compare ("only qualifier changed") because the code change to org.eclipse.equinox.http.servlet didn't come with a service-version bump. Bumped the bundle from 1.8.600 to 1.8.700.
2. Added my DCO sign-off so the ECA check is satisfied.

Folded both into the existing commit, no behavior change beyond the original isValid() fix and its test.

netliomax25-code:
Pushed a fix for the Tests / Build failure. The inverted-range test is the first to drive a bodyless 416 through ServletRequestAdvisor.request, whose getErrorStream() comes back null for an empty error body, so the helper threw an NPE on close before the assertion ran. Guarded the request helper against a null stream so the 416 is reported; the range logic itself is unchanged.

PR #1303, opened 2026-06-01
escape message in built-in error page

netliomax25-code:
The fallback error page in ResponseStateHandler writes the error message into the HTML body at ResponseStateHandler.java:131 without escaping, and that message carries the request URI (ResourceServlet/ProxyServlet pass "ProxyServlet: " + req.getRequestURI() to sendError). A request URI with markup is reflected unescaped, so escape it before writing.

### DIFF
--- a/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/servlet/ResponseStateHandler.java
+++ b/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/servlet/ResponseStateHandler.java
@@ -128,7 +128,7 @@ public class ResponseStateHandler {
 				writer.print(httpStatus.description());
 				writer.println("</small></h1>"); //$NON-NLS-1$
 				writer.print("<p>"); //$NON-NLS-1$
-				writer.print(message);
+				writer.print(escapeHTML(message));
 				writer.println("</p></div></body></html>"); //$NON-NLS-1$
 			}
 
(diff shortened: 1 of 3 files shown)


bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/servlet/ResponseStateHandler.java
```diff
@@ -169,6 +169,14 @@ public void setException(Exception exception) {
 		this.exception = exception;
 	}
 
+	private static String escapeHTML(String value) {
+		if (value == null) {
+			return null;
+		}
+		return value.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$ //$NON-NLS-6$
```
akurtakov, line 176: This will break in the case of string contains already escaped symbols e.g. "&lt;" will become "&amp;lt;:

netliomax25-code: Good point on the double-escape. In this spot the message is always plain text though, it's either the sendError() string or exception.getMessage(), not pre-escaped HTML. So a literal "&lt;" in that text is meant to show as "&lt;" on the page, and "&amp;lt;" is exactly what renders it that way. Escaping the whole string once at output is the standard handling for the reflected case. I can special-case existing entities if you'd rather, but detecting them reliably gets fragile and reopens the bypass, so I'd lean toward keeping the single pass.

netliomax25-code:
Pushed a fix for the Code Analysis build failure: the Tycho baseline check was failing because I'd changed the http.servlet source without bumping the bundle, so I moved Bundle-Version from 1.8.600 to 1.8.700. Also added my Signed-off-by. The ECA check still reports my Eclipse account can't be found, so I'll get that registered on my end.

netliomax25-code:
Rebased onto current master to clear the Code Analysis failure. The baseline errors were against org.eclipse.osgi and the launcher, which got version bumps on master after this branch was cut, so the stale base was the cause rather than the patch itself. No code changes in the rebase, and the http.servlet bundle still builds clean locally.

netliomax25-code:
Pushed a fix for the Code Analysis failure. The comparator flagged the org.eclipse.equinox.server.jetty feature this time: the http.servlet version bump changes the feature's content, so the feature version has to move as well. Bumped it from 1.11.1100 to 1.11.1200, same shape as the bot's version bump commits, and the feature builds clean locally.

netliomax25-code:
any update?

PR #1305, opened 2026-06-02
use constant-time comparison in User.hasCredential

netliomax25-code:
hasCredential compares a stored credential against the supplied value non-constant-time: the byte[] branch returns on the first differing byte (and earlier on a length mismatch), and the String branch goes through equals(). For a password credential that leaks the matching-prefix length through timing.

Switch both branches to MessageDigest.isEqual.


netliomax25-code:
Re-pushed with a DCO `Signed-off-by` added to the commit, since the ECA check needs it. No code change otherwise: both branches of `hasCredential` still route through `MessageDigest.isEqual` instead of the short-circuiting compares.

PR #1308, opened 2026-06-03
Remove deprecated PluginVersionIdentifier

HannesWell:
The class PluginVersionIdentifier was deprecated for 20 years and marked for removal for over two years.

Follow-up on
- https://github.com/eclipse-equinox/equinox/pull/427


HannesWell:
When this removal is submitted, the Plug-in `org.eclipse.core.runtime` reexporting this bundle, needs it's version updated in the same way:
https://github.com/eclipse-platform/eclipse.platform/blob/859044ca5cad98c5f86c44ef53f68e12b7c3d2df/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF#L13

akurtakov (approved):

features/org.eclipse.equinox.server.core/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.equinox.server.core"
       label="%featureName"
-      version="1.16.900.qualifier"
+      version="1.16.1000.qualifier"
```
tjwatson, line 5: I'm surprised this isn't a minor version bump.

merks: Given the rules are stated as follows, the second case applies so indeed we should reflect in the feature version the fact that there are API changes in one of the bundles.

-  Increment the feature's major number if any contained plug-in or feature increases their major number
- Otherwise, increment the feature's minor number if any contained plug-in or feature increases their minor number
- Otherwise, increment the feature's service number if any contained plug-in or feature increases their service number.

HannesWell: Indeed. I didn't pay attention to that. Applied that in
- https://github.com/eclipse-equinox/equinox/pull/1310

PR #1309, opened 2026-06-03
Remove deprecated and (almost) empty org.eclipse.osgi.services bundle

HannesWell:
The bundle `org.eclipse.osgi.services` is deprecated for over two years and I assume sufficient time has now passed to finally remove it (similarly to code deprecated for removal).

Last part of
- https://github.com/eclipse-equinox/equinox/issues/18

Follow-up on
- https://github.com/eclipse-equinox/equinox/pull/403

In SimRel there is no dependency on `org.eclipse.osgi.services` anymore:
grafik


laeubi (approved):
The current one is available in numerous old releases and on maven central and we do not have planned to evolve it anyways so I see no problem in removing it.

PR #1310, opened 2026-06-04
Minor version bump(s) for 4.41 stream

HannesWell:
A contained `org.eclipse.equinox.common` Plug-in has its minor version increased, so the features should do the same.
Furthermore the `org.eclipse.osgi.services` Plug-in was removed from the equinox.core Feature.

Follow-up on
- https://github.com/eclipse-equinox/equinox/pull/1308#discussion_r3356189833


tjwatson (approved):

merks:
Thanks @HannesWell

PR #1313, opened 2026-06-12
[Win32] Add format for VS 2026 path to binaries build script

HeikoKlare:
Visual Studio 2026 uses a new default path pattern containing the internal version number instead of the external version year. This change adapts the logic for searching the visual studio installation in the build script to match this Visual Studio 2026 path.

This prepares for the GHA runner update to using a new Windows Server image that contains Visual Studio 2026 by default, which already produced build failures for SWT:
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3380


HannesWell (approved):
Thank you.

PR #1314, opened 2026-06-12
Remove temporarily preserved EquinoxBundle#getModuleClassLoader(boolean)

HeikoKlare:
The private method `EquinoxBundle#getModuleClassLoader(boolean)` was replaced by a method with a different signature for the 2025-12 release. Some relevant consumers (at least Jetty OSGi's `DefaultBundleClassLoaderHelper`) reflectively accessed the replaced method. To avoid that compatibility of such consumers breaks, the replaced method was temporarily preserved for the 2025-12 release:
- https://github.com/eclipse-equinox/equinox/pull/1211

Jetty, as the known consumer of that method, was updated immediately after that:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3546

The idea was to keep the temporary revert for some releases and then remove it again (https://github.com/eclipse-equinox/equinox/pull/1211#issuecomment-3617372252). Three releases seem to be sufficient time, so that I propose to remove the temporary revert for the 2026-09 release.

This change finally removes the replaced method by reverting commit 74e38821ef7ea257da027a00117c8484603702fb.

### DIFF
--- a/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF
@@ -107,7 +107,7 @@ Bundle-Activator: org.eclipse.osgi.internal.framework.SystemBundleActivator
 Bundle-Description: %systemBundle
 Bundle-Copyright: %copyright
 Bundle-Vendor: %eclipse.org
-Bundle-Version: 3.24.200.qualifier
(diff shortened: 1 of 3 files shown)


tjwatson (approved):

tjwatson:
Thanks!

PR #1315, opened 2026-06-12
Remove unused values for unsupported OS and WS from launcher Constants

HannesWell:
All references to these constants were removed in
- https://github.com/eclipse-equinox/equinox/pull/1151

and they cannot be referenced outside the package.

@tjwatson I consider your thump-up on https://github.com/eclipse-equinox/equinox/pull/1151#issuecomment-4624920598, as agreement.

### DIFF
--- a/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Constants.java
+++ b/bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Constants.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2024 IBM Corporation and others.
+ * Copyright (c) 2006, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.equinox.launcher/src/org/eclipse/equinox/launcher/Constants.java
```diff
@@ -35,75 +35,12 @@ class Constants {
...
-	@Deprecated(forRemoval = true, since = "2025-12")
-	public static final String OS_AIX = "aix";//$NON-NLS-1$
-
-	/**
-	 * Constant string (value "solaris") indicating the platform is running on a
-	 * Solaris-based operating system.
-	 *
-	 * @deprecated Solaris support has been removed
-	 */
-	@Deprecated(forRemoval = true, since = "2025-12")
```
akurtakov, line 53: This can be removed in 2027-12, no?

HannesWell: As described in https://github.com/eclipse-equinox/equinox/pull/1151#issuecomment-4624920598, I believe this constant is unusable outside of the package.
Generally everything in that class shouldn't be accessible since the class is package private and de-facto isn't API at all. So we don't have to wait for the removal.

Or do I miss something?

akurtakov: Ah, I missed that.

vogella:
Using the GitHub mobile app? ;-) This app creates merge commits for me even if I select rebase.

HannesWell:
> Using the GitHub mobile app? ;-) This app creates merge commits for me even if I select rebase.

Indeed I was using the Mobil app. To bad it has that bug, but for a long time it didn't even have that button.

Issue #1317, opened 2026-06-14
Unable to resolve bundle due to package-import substitution


ptziegler:
This is an issue I originally [reported](https://github.com/eclipse-pde/eclipse.pde/issues/2367) in the PDE project, but I'm now fairly certain that this is an edge-case that isn't handled properly in the Equinox implementation.

In short: Under certain situations it may happen that a package-import can't be resolved, even though a compatible package is being exported by a different bundle. The reason for this is that the imported package has to be within a given version range. But the exported package that would fit into this range is being substituted by a package with a higher version.

ptziegler:
I've created https://github.com/eclipse-equinox/equinox/pull/1318, which contains a small reproducer.

laeubi:
> I'm now fairly certain that this is an edge-case that isn't handled properly in the Equinox implementation.

The behavior is correct. Per OSGi specification the resolver must choose to either

1. Discard the export and use the import
2. Discard the import and use the export

Technically in almost all cases the import/export is actually wrong, because the spec requires for a substitution package to use a proper version range, what often is violated (e.g. in your case using no version range at all).

In general from my point of view substitution packages are almost always used wrong, they only solve a very specific use case that does not apply to 99% of bundles.

ptziegler:
Thanks for the clarification. So in short, the solution has to be to fix up the problematic bundle so that the packages are properly resolved again. Meaning there is nothing else that should be done here.

PR #1318, opened 2026-06-14
Create reproducer test case for incomplete bundle resolution #1317

ptziegler:


ptziegler:
This is the relevant code section. The package being exported in version 1.0.0 is being substituted by the package in version 2.0.0 and is therefore skipped when trying to find a match for the [1.0, 2) version range.

https://github.com/eclipse-equinox/equinox/blob/fde548a454aef85af997a0d8694122b0fc178165/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/module/ResolverConstraint.java#L51-L56

If I remove the `Import-Package` header from the exporting bundles, the test succeeds.

ptziegler:
Some additional logging when running the test. The package import from the first bundle is resolved against the second bundle. Which I suppose makes sense, as it doesn't have an upper bound. If I change it to `[1.0.0, 2.0.0)`, the test succeeds.

```
*** BEGIN RESOLUTION ***
** RESOLVING [A_0.0.0] **
Trying to resolve: [A_0.0.0], test
CHECKING: B_0.0.0, test
Trying to resolve: [B_0.0.0], test
CHECKING: B_0.0.0, test
Found match: [B_0.0.0]. Wiring [B_0.0.0]:test
[B_0.0.0] RESOLVED
Found match: [B_0.0.0]. Wiring [A_0.0.0]:test
CHECKING: A_0.0.0, test
[A_0.0.0] RESOLVED
** RESOLVING [B_0.0.0] **
  - [B_0.0.0] already resolved
** RESOLVING [C_0.0.0] **
Trying to resolve: [C_0.0.0], test
** IMPORT test[C_0.0.0] failed to resolve
[C_0.0.0] NOT RESOLVED
****** Result Wirings ******
    * WIRING for [A_0.0.0]
        (r) no requires
        (w) [A_0.0.0]:test -> [B_0.0.0]:test
    * WIRING for [B_0.0.0]
        (r) no requires
        (w) [B_0.0.0]:test -> [B_0.0.0]:test
    * WIRING for [C_0.0.0]
        (r) no requires
        (w) [C_0.0.0]:test -> NULL!!!
*** END RESOLUTION ***
```

The problem is that the manifest files where this originally showed up were generated by the felix-maven-plugin. So the manifest for a large chunk of OSGi bundles will look like this. :(

That said: If a package is exported by the bundle itself, shouldn't that take priority?

ptziegler:
> That said: If a package is exported by the bundle itself, shouldn't that take priority?

As an idea: Would it make sense  to sort the list of candidates, so that the bundle that is being resolved (if present) always shows up first?

bundles/org.eclipse.osgi.tests/src/org/eclipse/osgi/tests/resolver/TestSubstitution.java
```diff
@@ -0,0 +1,80 @@
...
+		assertTrue("unexpected bundle resolution state", bundle_1.isResolved());
+		assertTrue("unexpected bundle resolution state", bundle_2.isResolved());
+		assertTrue("unexpected bundle resolution state", bundle_3.isResolved());
+	}
+
+	public BundleDescription create_bundle_1(StateObjectFactory sof) throws BundleException {
+		Dictionary dictionary_1 = new Properties();
+		dictionary_1.put("Bundle-ManifestVersion", "2");
+		dictionary_1.put("Bundle-SymbolicName", "A");
+		dictionary_1.put("Import-Package", "test;version=\"1.0.0\"");
```
merks, line 59: One might argue that if you are going to import a package that you are also going export with a specific version that your import should be really careful about the version range of the import.  For example, doesn't this import allow to import package `test` with version `2.0.0` and then export it as if it had version `1.0.0`?  It certainly appears to allow the substitution, given the problem description...

The import probably should be **much** more careful to specify `[1.0.0,2.0.0)` or even `[1.0.0,1.0.1)`.   I know these kind of "misguided" results of "it's good to import your exports" have caused problem with Orbit as well, where do no import the exports when I synthesize the MANIFEST.MF.

Probably the real world example is a bundle for which you have no control...

ptziegler: > Probably the real world example is a bundle for which you have no control...

Yes, as mentioned in https://github.com/eclipse-equinox/equinox/pull/1318#issuecomment-4701596460, the problem goes away when setting an upper bound.

The "real world" bundle that's causing all of this is the [osgi-resource-locator](https://mvnrepository.com/artifact/org.glassfish.hk2/osgi-resource-locator/1.0.4), with the following manifest:

```
Manifest-Version: 1.0
Created-By: Apache Maven Bundle Plugin 6.0.0
Build-Jdk-Spec: 17
Bundle-ActivationPolicy: lazy
Bundle-Activator: org.glassfish.hk2.osgiresourcelocator.Activator
Bundle-Description: Eclipse Enterprise for Java (EE4J) is an open source
  initiative to create standard APIs,        implementations of those AP
 Is, and technology compatibility kits for Java runtimes        that ena
 ble development, deployment, and management of server-side and cloud-na
 tive applications.
Bundle-DocURL: https://www.eclipse.org
Bundle-License: http://www.eclipse.org/legal/epl-2.0, https://www.gnu.or
 g/software/classpath/license.html
Bundle-ManifestVersion: 2
Bundle-Name: HK2 OSGi resource locator
Bundle-SymbolicName: org.glassfish.hk2.osgi-resource-locator
Bundle-Vendor: Eclipse Foundation
Bundle-Version: 1.0.4
Export-Package: org.glassfish.hk2.osgiresourcelocator;uses:="org.osgi.fr
 amework";version="1.0.4"
Import-Package: java.io,java.lang,java.lang.invoke,java.net,java.securit
 y,java.util,java.util.concurrent.locks,org.osgi.framework;version="[1.1
 0,2)",org.glassfish.hk2.osgiresourcelocator;version="1.0.4"
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=11))"
Tool: Bnd-7.0.0.202310060912
```

So not something I have any control over and which has exactly the same problem.

merks:
I see you kind of reach the same conclusion about the cause.  I mostly this for Obit when generating the MANIFEST.MF

```
Export-Package:         *;version="${version}";-noimport:=true
```

When I don't do that, it generated more restricted versions for the imports::

image

So I'm not sure from where the bounded import comes from, but it seems bad and is not even the BND default behavior...

ptziegler:
fwiw, I've also created a [PR](https://github.com/eclipse-ee4j/glassfish-hk2-extra/pull/186) for the OSGi resource locator bundle.

I'm not sure if this is something that Equinox can/should handle. And even if the bundle is now using proper bounds, it doesn't really solve the issue we're having with our application. But I think that's something I can work around.

merks:
Compliments on creating a PR!

laeubi:
> This is the relevant code section. The package being exported in version 1.0.0 is being substituted by the package in version 2.0.0 and is therefore skipped when trying to find a match for the [1.0, 2) version range.
> 
> https://github.com/eclipse-equinox/equinox/blob/fde548a454aef85af997a0d8694122b0fc178165/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/module/ResolverConstraint.java#L51-L56
> 
> If I remove the `Import-Package` header from the exporting bundles, the test succeeds.

As explained here https://github.com/eclipse-equinox/equinox/issues/1317#issuecomment-4702451958 this is actually correct behavior per OSGi specification. A package that was substituted must discard the export package.

Technically the resolver has a bit of a choice, but this is currently not implemented and not trivial to implement in the current resolver code. 

I have been working on a new more deterministic resolver to solve this, but sadly this is the kind of things people are very upset when they are hit, but using a workaround is easier. So even if I would love to work on that topic there is currently no budget or sponsoring about this. A usual band-aid is to define a fragment to the problematic bundle that restricts the version range (even though this only helps at runtime not at development time).

If someone has interest in more details I can link the corresponding issues/PR and discussions at bnd+osgi but in general this is not a bug per-se.

laeubi:
> `Import-Package: ...,org.glassfish.hk2.osgiresourcelocator;version="1.0.4"`

This is almost always bogus and usless and even harmful (as you noticed), so one should really ever use `-noimport:=true` as mentioned by @merks

laeubi:
> That said: If a package is exported by the bundle itself, shouldn't that take priority?

OSGi rules are quite clear here, the resolver must always resolve the highest version if no upper bound is given.

> As an idea: Would it make sense to sort the list of candidates, so that the bundle that is being resolved (if present) always shows up first?

This would be violating the OSGi spec.

laeubi:
And one last comment: Even if that would be changed and would solve the issue in PDE, it would still fail at runtime probably (what uses the new Felix Resolver).

ptziegler:
> this is the kind of things people are very upset when they are hit

It's not anger, but rather the lack of transparency that bothers me. As mentioned in the PDE bug, I've set the version to 2.48.0 in the feature.xml. So when I launch the application, I expect that version to be picked up, as opposed to a different version that just happens to be in the target definition. Or at the very least I would've expected some notification that the specified version can't be used for reason X.

I probably also could've enabled the debug flags when launching the product. But for 700+ plugins, this likely would've become completely unreadable and I'm not aware of any other way to "easily" debug these kinds of problems.

> Even if that would be changed and would solve the issue in PDE, it would still fail at runtime probably

We only have one version of the bundle in the actual product. So at runtime, this wouldn't be an issue because there is only a single candidate.

That said, thanks again for the clarification. Given that there is nothing to do for Equinox, I'm closing this PR as well.

laeubi:
> It's not anger, but rather the lack of transparency that bothers me. As mentioned in the PDE bug, I've set the version to 2.48.0 in the feature.xml. So when I launch the application, I expect that version to be picked up, as opposed to a different version that just happens to be in the target definition.

That should actually be the case. I though that *both* versions are included isn't it?

> I'm not aware of any other way to "easily" debug these kinds of problems.

Its not really easier but often revealing to use [the `wires` command](https://equinox.eclipseprojects.io/commands/wires.html).

> We only have one version of the bundle in the actual product. So at runtime, this wouldn't be an issue because there is only a single candidate.

I see - sadly PDE only supports exclusions in form of a while list (selected bundles) and not a black list... Adding support for a black list is also on my list bu I also haven't found yet time to implement it :-\

ptziegler:
> That should actually be the case. I though that both versions are included isn't it?

No. We have two products, one that is using Jersey 2.48.0 and another one that is using Jersey 4.0.2, which is why both versions need to be specified in the target definition.

But because of the improper version ranges, the 2.48.0 version can't be resolved because it requires the osgi-resource-locator package in version 1.x, which is substituted by the 3.x version.

When launching a product from the IDE, PDE first filters out all unresolved bundles. From the remaining ones, it then selects the one where the versions match perfectly (if possible), otherwise it selects the latest one. But because the "perfect match" is among those that are being filtered out, it silently falls back to picking the latest version and the launch fails.

PR #1321, opened 2026-06-20
Replace ant-contrib with ifunless standard ANT attributes

HannesWell:
Ant 1.9.1 added https://ant.apache.org/manual/ifunless.html, which allows to replace all uses of `ant-contrib`.

Similar to
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/3395


akurtakov:
Thanks for cleaning this one too!

PR #1326, opened 2026-06-25
[Build] Fix executions of OSGi TCK test cases

HannesWell:
- Remove enabling the baseline comparison for all builds by default
It failed the build performed in preparation of the TCK execution because of missing version increments. For the TCK the repository is only cloned shallowly, which breaks the jgit timestamp provider and has the effect that the build-time is used as qualifier for all artifacts.
- Restrict test-runtime to include only JUnit-5 artifacts.
The biz.aQute.tester does not support JUnit-6 yet.
- Replace use of archived andymckay/append-gist-action for badge upload
It uses Node 20 which is deprecated
- Update badge gists only for TCK executions completed on the master
Previously the badges where also updated with PR results because due to
the setup, the uploading workflow has the master as direct context.
- Fail TCK build in case of errors, but ignore test-failures because they are handled by the specialized result collection.
- Remove some generally unused `tycho-surefire` parameters and apply general clean-ups

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,8 +114,8 @@ jobs:
         -Pbree-libs
         -Ptck
         -Dskip-default-modules=true
-        -Dtycho.resolver.classic=false
-        -fn
+        -Dmaven.test.failure.ignore
+        -fae
         clean verify
     - name: Upload TCK Results
       uses: actions/upload-artifact@v7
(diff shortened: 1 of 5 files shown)


laeubi (approved):
Looks good thanks for looking into this!

HannesWell:
Great, thanks.

I just updated this change to not fail the TCK build because of test failures, like the still occurring
- https://github.com/eclipse-equinox/equinox/issues/190

Generally test failures are still displayed by the separate TCK specific test result publication.
Other kind of failures should still fail the build.

PR #1336, opened 2026-08-20
Add missing binary inclusion of OSGI-INF folder for weaving bundles

HannesWell:
Without this the localization files are missing in the built bundles.

This was forgotten in
- https://github.com/eclipse-equinox/equinox/pull/521

I believe this change is save but on the other hand not urgent since nobody complained for the missing files for over two years. 
We are effectively post RC1 so this would require project lead approval.
@tjwatson, @laeubi should this be submitted now or postponed to the next cycle?


laeubi (approved):
I think it is safe to submit, so just in case we want to have it I approve this PR as a PL

merks:
It looks safe to me as well.  I assume we'd generate better maven artifacts if/when this is fixed?

image

vogella:
> It looks safe to me as well. I assume we'd generate better maven artifacts if/when this is fixed?

I assume you know but since day or weeks the Maven publishing job seems to fail:

````
Please go to https://ci.eclipse.org/releng/job/releng/job/deploy-maven/441/console and check the build failure._______________________________________________

````

merks:
> > It looks safe to me as well. I assume we'd generate better maven artifacts if/when this is fixed?
> 
> I assume you know but since day or weeks the Maven publishing job seems to fail:
> 
> ```
> Please go to https://ci.eclipse.org/releng/job/releng/job/deploy-maven/441/console and check the build failure._______________________________________________
> ```

Yes, we've opened two different help desk tickets for the two different problems we've had:

image

The first one took a week to resolve, and the one from today (after the failures started again) was resolved very quickly today and I reran the build to verify the fix.

HannesWell:
> I think it is safe to submit, so just in case we want to have it I approve this PR as a PL

Thanks, then lets have this now.


> It looks safe to me as well. I assume we'd generate better maven artifacts if/when this is fixed?

Yes, these are exactly the values that should be correctly resolved after this is submitted.

PR #1347, opened 2026-09-02
Add State.getExportedPackages(String) to query exports by package name

vogella:
State.getExportedPackages() copies every selected export of every resolved bundle into a new array on each call. PDE's manifest editor calls it once per Import-Package row on the UI thread only to check whether a single package is exported, which on an SDK sized target with 731 bundles copies about 3200 exports per row (see eclipse-pde/eclipse.pde#2444).

This adds a name-filtered getExportedPackages(String packageName). It is a default method on State that filters the full array, so external implementors keep compiling. StateImpl overrides it to walk the resolved bundles and removal pendings once under the state monitor and collect only matching exports, and ReadOnlyState delegates to the system state. Package org.eclipse.osgi.service.resolver goes to 1.8.0 and org.eclipse.osgi to 3.25.0.

### DIFF
--- a/bundles/org.eclipse.osgi.compatibility.state/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi.compatibility.state/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.eclipse.osgi.compatibility.state
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 1.3.100.qualifier
 ExtensionBundle-Activator: org.eclipse.osgi.compatibility.state.Activator
 Fragment-Host: org.eclipse.osgi;bundle-version="3.12.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
--- a/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/compatibility/state/ReadOnlyState.java
+++ b/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/compatibility/state/ReadOnlyState.java
@@ -75,6 +75,11 @@ public final class ReadOnlyState implements State {
 	}
 
 	@Override
+	public ExportPackageDescription[] getExportedPackages(String packageName) {
+		return platformAdmin.getSystemState().getExportedPackages(packageName);
+	}
+
+	@Override
 	public StateObjectFactory getFactory() {
 		return platformAdmin.getSystemState().getFactory();
 	}
--- a/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/resolver/StateImpl.java
+++ b/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/internal/resolver/StateImpl.java
@@ -837,6 +837,31 @@ public abstract class StateImpl implements State {
 		}
 	}
 
+	@Override
+	public ExportPackageDescription[] getExportedPackages(String packageName) {
+		fullyLoad();
+		synchronized (this.monitor) {
+			List<ExportPackageDescription> result = new ArrayList<>();
+			addSelectedExports(resolvedBundles.values(), packageName, result);
+			addSelectedExports(removalPendings, packageName, result);
+			return result.toArray(new ExportPackageDescription[result.size()]);
+		}
+	}
+
+	private static void addSelectedExports(Collection<BundleDescription> bundles, String packageName, List<ExportPackageDescription> result) {
+		for (BundleDescription bundle : bundles) {
+			ExportPackageDescription[] bundlePackages = bundle.getSelectedExports();
+			if (bundlePackages == null) {
+				continue;
+			}
+			for (ExportPackageDescription export : bundlePackages) {
+				if (export.getName().equals(packageName)) {
+					result.add(export);
+				}
+			}
+		}
+	}
+
 	private static final Comparator<BundleDescription> BY_BUNDLE_ID = Comparator.comparingLong(BundleDescription::getBundleId);
 
 	BundleDescription[] getFragments(final BundleDescription host) {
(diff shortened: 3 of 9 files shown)


tjwatson:
Do you have performance metrics for this change?
From https://github.com/eclipse-pde/eclipse.pde/pull/2444

>  replaces roughly 9600 reference stores per call with 731.

While I understand that is less, it does not always result in measurable performance improvement.

I think it is time to move the deprecated `org.eclipse.osgi.service.resolver` API out of `org.eclipse.osgi` and into the `org.eclipse.osgi.compatibility.state` fragment so it can be independently versioned outside of the framework.

I think we should do this without a disruptive major version change to `org.eclipse.osgi`.  Anyone using `Import-Package` to get the  `org.eclipse.osgi.service.resolver` package will find the package from the fragment for installation.

HannesWell:
> I think it is time to move the deprecated `org.eclipse.osgi.service.resolver` API out of `org.eclipse.osgi` and into the `org.eclipse.osgi.compatibility.state` fragment so it can be independently versioned outside of the framework.

Having the API and implementation in one bundle would also avoid the difficulties we currently have to ensure the implementation fragment is available.
 
> I think we should do this without a disruptive major version change to `org.eclipse.osgi`. Anyone using `Import-Package` to get the `org.eclipse.osgi.service.resolver` package will find the package from the fragment for installation.

In general there are probably only few consumers of that package. And adding Import-Package now would definitively also work with previous versions.

vogella:
> Do you have performance metrics for this change? From [eclipse-pde/eclipse.pde#2444](https://github.com/eclipse-pde/eclipse.pde/pull/2444)
> 
> > replaces roughly 9600 reference stores per call with 731.
> 
> While I understand that is less, it does not always result in measurable performance improvement.

Here is the measurement, took me a while to re-create it, my previous trace I just noticed that it takes unnecessary work.


### Setup: 

Eclipse SDK 4.42 I-build 20260902-2327, stock org.eclipse.osgi 3.24, headless under Xvfb, target platform = running platform (850 bundles, about 3200 exported packages).
Before: stock org.eclipse.pde.core 3.21.500.v20260902-1925. 
After: org.eclipse.pde.core built from PR 2444 (a7014cfb), everything else identical, IDE started with -clean.
Scenario, ten iterations per run: close the manifest editor, open META-INF/MANIFEST.MF, switch to the Dependencies page, wait until the UI thread is idle. The UI thread was sampled at 10 ms, then the same ten iterations ran again under JFR with allocation sampling.
Two manifests: org.eclipse.equinox.p2.tests (48 Import-Package rows) and a copy of org.apache.ant's manifest (105 rows). The table refreshes twice per open (creation and page activation), so each open is 96 or 210 isResolved() calls.

### Results

| | 48 rows before | 48 rows after | 105 rows before | 105 rows after |
|---|---|---|---|---|
| UI thread samples in getExportedPackages / all | 5 / 533 | 0 / 529 | 12 / 543 | 0 / 462 |
| UI thread samples in isResolved / all | 5 / 533 | 0 / 529 | 12 / 543 | 0 / 462 |
| Estimated UI thread time in isResolved per page open | about 5 ms | below 1 ms | about 12 ms | below 1 ms |
| JFR allocation weight under isResolved, ten opens | 175 MB | 4 MB | 470 MB | none sampled |
| Median switch to the Dependencies page | 68 ms | 66 ms | 57 ms | 55 ms |
| GC events during the run | 0 | 0 | 0 | 0 |


Before, every sampled stack is one path: DependenciesPage -> ImportPackageSection -> TableViewer refresh -> PDELabelProvider.getObjectImage -> ImportPackageObject.isResolved -> StateImpl.getExportedPackages, and the allocation under it is the fresh ExportPackageDescription[] plus the ArrayList growth that builds it, about 220 to 360 KB per call. 

After, the method no longer appears in 10 ms samples at all, and what JFR still attributes to isResolved is the BundleDescription[] copy that State.getResolvedBundles() returns, about 7 KB per call.

What this does and does not show: the work and the allocation scale with the row count as the PR says, and the change removes them. It does not make the page measurably faster at 48 or 105 rows, because 5 to 12 ms of UI thread per open was already below what the page switch costs, and no GC pause was triggered on either side. The benefit is fewer allocations and less UI thread work in a place that runs on every refresh, not a visible latency change.

before-flame-filtered-48rows

### My conclusion

I would argue that the memory safe is the most important part here providing this code can avoid garbage collector runs. And it is slightly faster.


> I think it is time to move the deprecated `org.eclipse.osgi.service.resolver` API out of `org.eclipse.osgi` and into the `org.eclipse.osgi.compatibility.state` fragment so it can be independently versioned outside of the framework.
> 
> I think we should do this without a disruptive major version change to `org.eclipse.osgi`. Anyone using `Import-Package` to get the `org.eclipse.osgi.service.resolver` package will find the package from the fragment for installation.

I assume that is a general discussion not related to this specific change?

tjwatson:
> I assume that is a general discussion not related to this specific change?

Related because once the package was deprecated there was an assumption that we would no longer evolve the deprecated API and therefore not require minor updates to `org.eclipse.osgi`.  But since we know PDE is going to use it long term, that is no longer the case as PDE needs to evolve the API.

Moving the API to the impl artifact makes that much more clean as we need to evolve the API more in the future.

vogella:
> > I assume that is a general discussion not related to this specific change?
> 
> Related because once the package was deprecated there was an assumption that we would no longer evolve the deprecated API and therefore not require minor updates to `org.eclipse.osgi`. But since we know PDE is going to use it long term, that is no longer the case as PDE needs to evolve the API.
> 
> Moving the API to the impl artifact makes that much more clean as we need to evolve the API more in the future.

Thanks for the explanation. Are you going to do that or do you want me to do as part of this PR?

tjwatson:
> Thanks for the explanation. Are you going to do that or do you want me to do as part of this PR?

No, that can be done later.

vogella:
> > Thanks for the explanation. Are you going to do that or do you want me to do as part of this PR?
> 
> No, that can be done later.

I would also think that I'm the wrong person for doing that and that you are the best for doing it.

vogella:
I think the TCK failure is known. @tjwatson is the change ok for you? It saves primary memory.

laeubi (approved):
It looks fine so far and even if we don't want to really promote/evolve the API it does not harm.

laeubi:
> Moving the API to the impl artifact makes that much more clean as we need to evolve the API more in the future.

As far as I know the impl is actually a fragment what makes it quite brittle to use. It also contains some other "legacy" implementation. At least for P2 I ([almost](https://github.com/eclipse-equinox/p2/issues/1091)) got rid of it, so one approach could be to finally move that stuff entirely to PDE.

vogella:
Thanks for the review @laeubi

PR #1348, opened 2026-09-02
Replace JNA with java.lang.foreign in equinox.security.linux

akurtakov:
Use FFM downcalls to libsecret/GIO instead of JNA mappings and drop the com.sun.jna requirement.

Needs BREE JavaSE-25, so the fragment no longer resolves on older JREs.

Assisted-by: Anthropic Claude Code (claude-opus-5[1m])


akurtakov:
@jjohnstn I would appreciate your review and testing

bundles/org.eclipse.equinox.security.linux/src/org/eclipse/equinox/internal/security/linux/LinuxPasswordProvider.java
```diff
@@ -40,104 +39,96 @@ public class LinuxPasswordProvider extends PasswordProvider implements IValidati
...
+				} finally {
+					// the out list holds a reference to each object it names
+					fLibGio.listFreeFullUnref(unlocked.get(ValueLayout.ADDRESS, 0));
+					fLibGio.listFree(list);
+				}
+
+				requireNoError(gerror, "Unable to unlock: "); //$NON-NLS-1$
+			}
+		} finally {
+			owned.forEach(fLibGio::objectUnref);
```
iloveeclipse, line 90: Is it OK to call `fLibGio::objectUnref` on `owned` elements created by `fLibSecret`? I believe this is what confused Copilot too?

akurtakov: It works just fine for me. Verified using :
```java
package org.eclipse.equinox.internal.security.linux;

import java.lang.foreign.*;
import java.util.*;

public class Leak {
      public static void main(String[] args) {
              System.out.println("with owned.forEach(fLibGio::objectUnref): " + cycle(true));
              System.out.println("without it (pre-fix behaviour):           " + cycle(false));
      }

      static String cycle(boolean unref) {
              LibGio fLibGio = LibGio.getInstance();
              LibSecret fLibSecret = LibSecret.getInstance();
              int first = 0, last = 0;
              for (int i = 0; i < 100; i++) {
                      List<MemorySegment> owned = new ArrayList<>(3);
                      try (Arena arena = Arena.ofConfined()) {
                              MemorySegment gerror = arena.allocate(ValueLayout.ADDRESS);
                              owned.add(fLibGio.busGetSync(GBusType.G_BUS_TYPE_SESSION, MemorySegment.NULL, gerror));
                              MemorySegment secretService = fLibSecret.serviceGetSync(
                                              SecretServiceFlags.SECRET_SERVICE_LOAD_COLLECTIONS, MemorySegment.NULL, gerror);
                              owned.add(secretService);
                              last = secretService.reinterpret(16).get(ValueLayout.JAVA_INT, 8); // GObject.ref_count
                              if (i == 0) first = last;
                      } finally {
                              if (unref) {
                                      owned.forEach(fLibGio::objectUnref);
                              }
                      }
              }
              return "ref_count " + first + " -> " + last;
      }
}
```

bundles/org.eclipse.equinox.security.linux/src/org/eclipse/equinox/internal/security/linux/SecretSchema.java
```diff
@@ -10,41 +10,60 @@
...
+final class SecretSchema {
+
+	/** Fixed size of the inline {@code attributes} array in the C struct. */
+	private static final int MAX_ATTRIBUTES = 32;
+
+	/** {@code SecretSchemaAttribute { const gchar *name; SecretSchemaAttributeType type; }} */
+	private static final GroupLayout ATTRIBUTE_LAYOUT = MemoryLayout.structLayout( //
+			ValueLayout.ADDRESS.withName("name"), //$NON-NLS-1$
+			ValueLayout.JAVA_INT.withName("type"), //$NON-NLS-1$
+			MemoryLayout.paddingLayout(4));
```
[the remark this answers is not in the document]
akurtakov: Eclipse doesn't support 32 bit at all so there is no need.

HannesWell:
Thanks for working on this.
Have you considered to generate the FFM bindings with [`jextract`](https://github.com/openjdk/jextract), based on existing C header files? Iike done in
- https://github.com/eclipse-platform/eclipse.platform/pull/2908

akurtakov:
Honestly no, I haven't looked into jextract at all. I left it for exercise for later as this one is noisy already to get Java 25 ready.

akurtakov:
It is in good shape according to my testing so unless there are concerns I plan to push this one early next week.

merks:
I think everyone has been supporting moving in this direction so it looks like a good plan and now is the best time to make such changes.

PR #1349, opened 2026-09-02
Update GHAs to use newer maven and setup-maven new version

akurtakov:


akurtakov:
TCK test is unstable. Merging.

PR #1352, opened 2026-09-08
Use upstream org.eclipse.osgitech.plurl package

tjwatson:


laeubi:
@tjwatson is the intend to copy the sources here or should we check if/how to consume the jar with the code directly e.g. from maven central.

Technically it should work to add it as a `extra.classpath` for the IDE and embedd the class-files during the maven build.

tjwatson:
> @tjwatson is the intend to copy the sources here or should we check if/how to consume the jar with the code directly e.g. from maven central.
> 
> Technically it should work to add it as a `extra.classpath` for the IDE and embedd the class-files during the maven build.

I had intended on consuming from maven central, but I really only want to do that if the sources.jar we publish to maven central actually contains the source for plurl that we consume [org.eclipse.osgi-3.24.300-sources.jar](https://repo1.maven.org/maven2/org/eclipse/platform/org.eclipse.osgi/3.24.300/org.eclipse.osgi-3.24.300-sources.jar)

On the other hand, everything else the core framework depends on from other projects (osgi spec, felix) we have copies of in the repo, so it would not be an exception to do that for plurl also.

tjwatson:
I meant to mention that plurl has not had a release yet.  This PR is a draft because I wanted to test out the current state of the plurl source in a PR build.

PR #1353, opened 2026-09-09
Use the StateObjectFactory.defaultFactory in platform admin

laeubi:
Currently it creates an own instance but the `StateObjectFactory` itself is state-less and matches what the default factory actually is. To avoid having lingering around different instances where one is accessible only through the `PlatformAdmin` (what is only available through the compatibility fragment) this now also uses the default.

See for example 
- https://github.com/eclipse-equinox/p2/pull/1124

that can avoid referencing `PlatformAdmin` all together then as it only wants to access a limited set of the (legacy) state API.

### DIFF
--- a/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/compatibility/state/PlatformAdminImpl.java
+++ b/bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/compatibility/state/PlatformAdminImpl.java
@@ -25,7 +25,6 @@ import org.eclipse.osgi.internal.framework.BundleContextImpl;
 import org.eclipse.osgi.internal.framework.EquinoxContainer;
 import org.eclipse.osgi.internal.module.ResolverImpl;
 import org.eclipse.osgi.internal.resolver.StateHelperImpl;
-import org.eclipse.osgi.internal.resolver.StateObjectFactoryImpl;
 import org.eclipse.osgi.service.resolver.BundleDescription;
 import org.eclipse.osgi.service.resolver.DisabledInfo;
 import org.eclipse.osgi.service.resolver.PlatformAdmin;
@@ -37,7 +36,7 @@ import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceRegistration;
 
 public class PlatformAdminImpl implements PlatformAdmin {
-	private final StateObjectFactory factory = new StateObjectFactoryImpl();
+	private final StateObjectFactory factory = StateObjectFactory.defaultFactory;
 	private final Object monitor = new Object();
 	private EquinoxContainer equinoxContainer;
 	private BundleContext bc;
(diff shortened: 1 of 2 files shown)


bundles/org.eclipse.osgi.compatibility.state/src/org/eclipse/osgi/compatibility/state/PlatformAdminImpl.java
```diff
@@ -37,7 +36,7 @@
 import org.osgi.framework.ServiceRegistration;
 
 public class PlatformAdminImpl implements PlatformAdmin {
-	private final StateObjectFactory factory = new StateObjectFactoryImpl();
+	private final StateObjectFactory factory = StateObjectFactory.defaultFactory;
```
tjwatson, line 39: What is the motivation to using this?

`public static final StateObjectFactory defaultFactory = new StateObjectFactoryProxy();`

> that can avoid referencing PlatformAdmin all together then as it only wants to access a limited set of the (legacy) state API.

But this `StateObjectFactoryProxy()` will throw `UnsupportedOperationException` for every method call unless the implementation is installed (compatibility fragment):

```java
		private StateObjectFactory getImplementation() {
			synchronized (this.monitor) {
				if (implementation == null) {
					try {
						Class<?> implClass = Class.forName(IMPL_NAME);
						implementation = (StateObjectFactory) implClass.getConstructor().newInstance();
					} catch (Throwable t) {
						throw new UnsupportedOperationException(
								"Not able to create StateObjectFactory implementation: " + IMPL_NAME, t); //$NON-NLS-1$
					}
				}
				return implementation;
			}
		}
```

laeubi: The motivation is mainly that when the compatibility fragment **is present** there will be two "singleton" instances:

1. `factory` from `PlatformAdminImpl`
2. The `defaultFactory` from `StateObjectFactoryProxy`

if the compatibility fragment is not there there are trivially zero instances (as none can be instantiated).

With this change there is one indirection but we ensure there will always ever be one. I have just noticed that there are some rare cases where it is checked if an object was created by the given factory - so it seem to be failing if I once create something through `defaultFactory` versus `PlatformAdmin`

tjwatson: Can we instead just move the API to the compatibility fragment so the implementation itself doesn't need this strange proxy to get back to itself?

laeubi: In principle yes, but then we would need to

1. Mark the OSGi bundle as extensible API for PDE to use it properly
2. We actually would break the API contract of the osgi bundle

so I'm not sure its worth the hassle here...

tjwatson: > In principle yes, but then we would need to
> 
> 1. Mark the OSGi bundle as extensible API for PDE to use it properly

That is just harmless meta-data only PDE cares about so no big deal to add it.  Just checked and it is already [there](https://github.com/eclipse-equinox/equinox/blob/f3ca48a38ed31d591831caf7ae3cb105eb95f4c4/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF#L113).

> 2. We actually would break the API contract of the osgi bundle
> 
> so I'm not sure its worth the hassle here...

The API has been deprecated for well over a decade.  The goal has always been to eventually delete it.  This is just one step closer to being able to delete it altogether once PDE and p2 get off it.

PR #1354, opened 2026-09-11
Disable jdk=BREE for org.eclipse.osgi

### LINKED ISSUE #4038: org.eclipse.osgi I-builds (since switch to JDK 25 build host) contain a class referencing java.util.SequencedMap, breaking on Java 8-20
The official `org.eclipse.osgi` I-build artifact currently references `java.util.SequencedMap`, a type that doesn't exist before JDK 21, in a class file nominally targeting Java 8 (major version 52).
This breaks `org.eclipse.osgi` at runtime on any JDK 8-20 with `ClassNotFoundException: java.util.SequencedMap` the moment the affected method is invoked.

### Affected artifact

```
https://download.eclipse.org/eclipse/updates/4.42-I-builds/I20260909-2300/plugins/org.eclipse.osgi_3.25.0.v20260909-1123.jar
SHA-256: ad4e25f8ce6d83eeec2c5ae660aac1257fbbd331c56c2d422b596873425e82ba
```

### Evidence

```
$ unzip -p org.eclipse.osgi_3.25.0.v20260909-1123.jar org/apache/felix/resolver/util/OpenHashMap.class > OpenHashMap.class
$ javap -v OpenHashMap.class | grep "major version"
  major version: 52
$ javap -p -c OpenHashMap.class | grep -A3 "reversed()"
  public java.util.SequencedMap reversed();
    Code:
       0: aload_0
       1: invokeinterface #491,  1  // InterfaceMethod java/util/SortedMap.reversed:()Ljava/util/SortedMap;
       6: areturn
```

`OpenHashMap` (vendored Felix resolver code inside `org.eclipse.osgi`) implements `java.util.SortedMap` without overriding `reversed()`.
Since JDK 21, `SortedMap` extends `SequencedMap`, adding `reversed()`.
ECJ, when hosted on a JDK 21+ runtime and compiling at `-source/-target 8` without a `--release`/bootclasspath restriction that correctly limits API visibility, synthesizes a bridge method for this new abstract method — even though it otherwise emits correct Java-8-format bytecode.
**`javac` given the same flags does not synthesize this bridge.**
Root cause + standalone (Maven-free) reproducer: eclipse-jdt/eclipse.jdt.core#5389, fix proposal: eclipse-jdt/eclipse.jdt.core#5390.

### Confirmed via before/after comparison

| Build | Version | Built | Host JDK | `SequencedMap` leak |
|---|---|---|---|---|
| 2026-09 Release | `org.eclipse.osgi_3.24.300.v20260721-1251` | 2026-07-21 | JDK 21 (pre-switch) | Clean |
| I-build I20260909-2300 | `org.eclipse.osgi_3.25.0.v20260909-1123` | 2026-09-09 | JDK 25 (post-switch) | Broken |

```
Release jar:  https://download.eclipse.org/releases/2026-09/202609091000/plugins/org.eclipse.osgi_3.24.300.v20260721-1251.jar
              SHA-256: 4a770aa4a78df653af871073c8b4ddd56884895f5fc71d10c40b4289ac7c8b9f
              -> javap: no reversed() bridge, no SequencedMap reference

I-build jar:  https://download.eclipse.org/eclipse/updates/4.42-I-builds/I20260909-2300/plugins/org.eclipse.osgi_3.25.0.v20260909-1123.jar
              SHA-256: ad4e25f8ce6d83eeec2c5ae660aac1257fbbd331c56c2d422b596873425e82ba
              -> javap: public java.util.SequencedMap reversed(); (leak present)
```

This means the currently released 2026-09 SDK/target platforms are safe to use; only nightly I-builds since the JDK 25 build-host switch are currently affected.

### Possible trigger

The build host JDK was switched from 21 to 25 shortly before the affected I-build:

```
commit d3ed722cb8, Hannes Wellmann, 2026-09-03
"[Build|RelEng] Require Java-25 for the Eclipse SDK"
-		jdk 'temurin-jdk21-latest'
+		jdk 'temurin-jdk25-latest'
```

Note JDK 21 already contains `SequencedMap` too, so this may only be a coincidental/correlating change rather than the sole cause of the leak appearing now — flagging for awareness, not certainty. The before/after comparison above is consistent with this switch being the trigger, but I have not been able to compare against an I-build from between 2026-07-21 and 2026-09-03 to narrow it down further.

### Impact

Any downstream consumer (e.g. Eclipse Equinox's own OSGi TCK suite, see eclipse-equinox/equinox#295) that resolves `org.eclipse.osgi` from this repository and runs on Java 8-20 will hit this at runtime.

### Ask

Please confirm/track whether the platform build needs an explicit `--release`/bootclasspath fix in its Tycho compiler configuration (independent of upstream ECJ fixing the underlying bridge-synthesis bug tracked at eclipse-jdt/eclipse.jdt.core#5389).

laeubi:
Currently `org.eclipse.osgi` has no (deprecated) EE header and therefore `jdk=BREE` has silently fall back to the target platform EE. Since this was recently changed to Java 25 this suddenly pulled in unwanted API due to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5389 and failed to start on JDKs < 21.

Due to a bug in Tycho https://github.com/eclipse-tycho/tycho/pull/6314 it was not discovered that effectively `jdk=BREE` was not effective at all for a long time. so this now explicitly uses `jdk=SYSTEM` to ensure the -release flag is used instead consistently.

Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4038

### DIFF
--- a/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF
+++ b/bundles/org.eclipse.osgi/META-INF/MANIFEST.MF
@@ -107,7 +107,7 @@ Bundle-Activator: org.eclipse.osgi.internal.framework.SystemBundleActivator
 Bundle-Description: %systemBundle
 Bundle-Copyright: %copyright
 Bundle-Vendor: %eclipse.org
-Bundle-Version: 3.24.400.qualifier
+Bundle-Version: 3.24.500.qualifier
 Bundle-Localization: systembundle
 Bundle-DocUrl: http://www.eclipse.org
 Eclipse-ExtensibleAPI: true
--- a/bundles/org.eclipse.osgi/pom.xml
+++ b/bundles/org.eclipse.osgi/pom.xml
@@ -19,12 +19,14 @@
 </parent>
   <groupId>org.eclipse.platform</groupId>
   <artifactId>org.eclipse.osgi</artifactId>
-  <version>3.24.400-SNAPSHOT</version>
+  <version>3.24.500-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
   <properties>
 	  <!-- The actual TCKs are executed in the org.eclipse.osgi.tck module because of reference to other service implementations -->
 	  <tck.skip>true</tck.skip>
 	  <failOnJavadocErrors>false</failOnJavadocErrors>
+	  <!-- This ensures we use the -release 8 for compilation to ensure only java 8 API is used -->
+	  <tycho.useJDK>SYSTEM</tycho.useJDK>
   </properties>
   
   <build>


tjwatson:
Are there plans to make the non-deprecated `osgi.ee` requirement work for this?  If not, or if that will be a very long time coming then should we consider moving back to using BREE for `org.eclipse.osgi`?  There is no risk because this is the framework, we will not remove support for BREE if the framework itself uses it.

Living in the limitations of the build system seems more reliable (by using BREE).

laeubi:
Actually equinox tried to use a property but it was borked https://github.com/eclipse-tycho/tycho/pull/6314.
I think the ee capability could be supported these days, but it has not been a high priority yet as e.g. PDE does not yet support it so its likely very seldom used.

For the matter of fact, it becomes more and more hard to find a "real" java 1.8 JVM for some platforms (and `-target` works good enough) so I'm not sure we should care too much here when it comes to compilation. Instead, if we are really concerned about java 1.8 support it would be much more useful to setup a github action that _runs_ Equinox on a Java 1.8 jvm to make sure it is actually functional but I don't know if someone want to volunteer to setup/maintain this.

laeubi:
As expected now we get new warnings about changed classfiles - I'll push a qualifier touch here as well.

merks:
Has the service been version changed?

